PR #23, opened 2026-04-25
fix(mcp): start via npx/bin symlink (entrypoint detection broken on 0.0.3)

kamiazya:
## Summary

Verified that `@kamiazya/whiteboard-mcp@0.0.3` cannot start when invoked
through the npm `bin` symlink (which is how `npx` and `claude mcp add ... -- npx`
and Codex's `command = "npx"` invoke it).

### Root cause

`packages/mcp-server/src/server/mcp/index.ts:1691`:

```ts
const isEntryPoint = process.argv[1] === fileURLToPath(import.meta.url)
```

When invoked via `node_modules/.bin/whiteboard-mcp`, `process.argv[1]`
is the symlink path while `fileURLToPath(import.meta.url)` is the real
file path. The strict `===` returns false, `main()` is never called,
the process exits 0 silently, and `claude mcp list` reports
"Failed to connect".

### Fix

Resolve both sides through `realpathSync` before comparing.

### Why the bug shipped

`smoke:e2e` and `smoke:packaged` both spawn `node <entry.js>` directly,
so neither passes through the symlink that real installs use. Added a
new `smoke:bin` that:

1. `npm pack` the workspace
2. `npm install` the tarball into a tmp dir
3. Spawn `node_modules/.bin/whiteboard-mcp` (the symlink path)
4. Verify `initialize` and `tools/list` round-trip

Wired into `ci.yml` and `release.yml` so this regression cannot recur.
Also added `smoke:template` to ci.yml (was only running in release.yml,
which is how a similar regression slipped in earlier).

## Test plan

- [x] `pnpm smoke:bin` → `[bin-smoke] ALL OK` locally
- [x] `pnpm test`, `pnpm typecheck` green
- [ ] CI green (smoke:bin runs and passes)
- [ ] After merge + release-please cycle, `claude mcp add whiteboard -- npx -y @kamiazya/whiteboard-mcp@latest` connects

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,5 +57,11 @@ jobs:
       - name: Packaged stdio smoke
         run: pnpm smoke:packaged
 
+      - name: Bin symlink smoke (npx / claude mcp / codex install path)
+        run: pnpm smoke:bin
+
+      - name: Template smoke
+        run: pnpm smoke:template
+
       - name: Codex config smoke
         run: pnpm smoke:codex-config
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -111,6 +111,10 @@ jobs:
         working-directory: .
         run: pnpm smoke:packaged
 
+      - name: Bin symlink smoke (npx / claude mcp / codex install path)
+        working-directory: .
+        run: pnpm smoke:bin
+
       - name: Codex config smoke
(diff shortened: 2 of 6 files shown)


kamiazya:
Superseded — same fix already on main via PR #22 (entrypoint helper with realpathSync). Closing as redundant.

PR #32, opened 2026-04-25
feat(mcp): support outputPath and overwrite on export tools

kamiazya:
## Summary
- Add `outputPath` (absolute path) and `overwrite` options to `export_png` and `canvas_export_json`. When set, the export is written there directly; otherwise it falls back to the current `{DATA_DIR}/{sessionId}/exports/` location.
- Lets a local `.excalidraw` round-trip back into its source file and lets a PNG land next to a source asset in one step instead of a copy.
- Validation runs in both the daemon route and the MCP tool surface: a relative path returns 400 `invalid_output_path`, an existing file with `overwrite!=true` returns 409 `output_exists`. Parent directories are created via `mkdir -p`.
- The arbitrary-path write surface stays scoped to the export tools only.

## Test plan
- [x] `pnpm test --project mcp-node` (1082 passed, +18 new across server fn / route / tool / shared util layers)
- [x] `pnpm typecheck`
- [x] `pnpm smoke:e2e` (`[e2e] ALL OK`, including `canvas_export_json`)

## Notes
- Shared `OutputPathError` (`code: 'invalid_output_path' | 'output_exists'`) lives in `src/server/output-path.ts` and is reused by `export-json.ts` and `routes/export.ts`. Route handlers translate it into 400 / 409 responses.
- The MCP tool re-throws the daemon's `error` / `message` verbatim so the LLM can tell whether the path was relative or already existed.
- `output-path.ts` validates the `stat()` result with an `ENOENT`-only check, so permission errors (e.g. EACCES) propagate instead of being silently treated as "file does not exist".
- `SKILL.md` lists `outputPath` / `overwrite` under the `export_png` options block to keep the SKILL ↔ schema consistency test green.

### DIFF
--- a/packages/mcp-server/skills/whiteboard/SKILL.md
+++ b/packages/mcp-server/skills/whiteboard/SKILL.md
@@ -207,6 +207,8 @@ Options:
 - `scale`: export DPI multiplier, default 1
 - `minFontPx`: boosts font size only in the export clone
 - `frameId`: exports only the frame plus its child elements
+- `outputPath`: absolute path to write the PNG to. When omitted, write to the workspace exports dir
+- `overwrite`: replace an existing file at `outputPath`. Default false; without it an existing file rejects with `output_exists`
 
 Inspect the exported PNG visually:
 
--- a/packages/mcp-server/src/server/export-json.test.ts
+++ b/packages/mcp-server/src/server/export-json.test.ts
@@ -1,0 +1,135 @@
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
+import { mkdtemp, readFile, rm, writeFile, mkdir } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { LoroDoc, LoroMap } from 'loro-crdt'
+import { exportCanvasJsonDoc } from './export-json.js'
+
+function buildDocWithRect(): LoroDoc {
+  const doc = new LoroDoc()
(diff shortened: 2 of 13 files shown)


kamiazya:
@gemini-code-assist Both review comments addressed in b3e9b46.

**1. `fileExists` and `outputPath` validation duplication:**
Extracted into a new shared module `packages/mcp-server/src/server/output-path.ts`.
- `OutputPathError` is defined there and re-exported from `export-json.ts` so existing imports (e.g. `routes/canvas.ts`) keep compiling without churn.
- A single `validateOutputPath(outputPath, overwrite)` runs the absolute-path check and the existence check. Both `export-json.ts` and `routes/export.ts` now call this one helper.

**2. Overly broad `stat` catch:**
The new `fileExistsOrThrow` only treats `(error as NodeJS.ErrnoException).code === 'ENOENT'` as "missing" and rethrows everything else (EACCES, EPERM, ELOOP, …). That way `writeFile` does not silently paper over a permission problem; the real cause reaches the caller.

**Regression test:**
Added `output-path.test.ts`. Beyond the relative-path / `output_exists` / `overwrite=true` happy paths, it probes a `chmod 0o000` directory and asserts that the thrown error is **not** an `OutputPathError` — i.e. the EACCES from the underlying `stat` propagates rather than getting normalized into a friendly "doesn't exist" answer (skipped on win32 since posix permission semantics are not portable).

`pnpm test --project mcp-node` (1082 passed, +5) and `typecheck` are both green.

PR #33, opened 2026-04-25
fix(app): move onSceneChange debounce out of the render path

kamiazya:
## Summary
- Move the `onSceneChange` debounce out of the render path. The hook used to mutate refs and call `cancel()` inline during render, which is unsafe under StrictMode (double-invoke) and Concurrent rendering (rendering can be discarded mid-way).
- Derive the debounced callback via `useMemo([canvasKey])` and cancel the previous one in a `useEffect` cleanup. Swapping canvases still swaps to a fresh 300ms window for the new doc and the old pending timer is canceled deterministically.
- This also fixes a leak that the previous comment acknowledged: pending debounce timers were never canceled on unmount.

## Test plan
- [x] `pnpm test --project mcp-jsdom` (115 passed, +5 new)
- [x] `pnpm test --project mcp-node` (1082 passed)
- [x] `pnpm typecheck`

## Notes
- Removed `onSceneChangeFnRef` and `prevCanvasKeyRef` since `useMemo` now keys identity off `canvasKey` directly.
- Regression test (`useWhiteboardSync.scene-change.test.tsx`) locks: stable identity on same-key re-render (also under `<StrictMode>`), fresh identity on key switch, `vi.getTimerCount()` drops by 1 on key switch and on unmount. The unmount assertion is the one that fails on `main` and passes after the fix.
- The closure still captures `sessionId` / `slug`; if React ever discards the memo, the worst case is a fresh 300ms window — never a write into the wrong doc.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test suite for whiteboard synchronization lifecycle, validating behavior during scene changes, canvas key updates, debounce timer management, and component unmounting scenarios.

* **Refactor**
  * Enhanced whiteboard synchronization implementation with improved lifecycle management for debounce operations, ensuring proper cleanup when switching canvases or unmounting components.

### DIFF
--- a/packages/mcp-server/src/app/App.tsx
+++ b/packages/mcp-server/src/app/App.tsx
@@ -10,7 +10,7 @@ export default function App() {
       <Routes>
         <Route path="/" element={<IndexPage />} />
         <Route
-          path="/canvas/:sessionId/*"
+          path="/canvas/:workspaceId/*"
           element={
             <ErrorBoundary>
               <CanvasPage />
--- a/packages/mcp-server/src/app/components/HeaderBranchBanner.test.tsx
+++ b/packages/mcp-server/src/app/components/HeaderBranchBanner.test.tsx
@@ -51,7 +51,7 @@ afterEach(() => {
 describe('HeaderBranchBanner', () => {
   it('does not render the banner when HEAD is main', async () => {
     state.current.state.head = 'main'
-    render(<HeaderBranchBanner sessionId="s1" slug="c1" />)
+    render(<HeaderBranchBanner workspaceId="s1" slug="c1" />)
     // Wait for the initial fetch path.
     await act(async () => {
       await Promise.resolve()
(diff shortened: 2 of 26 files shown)


kamiazya:
@gemini-code-assist Addressed in b0bc04a, plus a related rename pass.

**1. Exhaustive deps fix**
You're right — the comment claimed `sessionId` / `slug` were listed but the array only contained `canvasKey`. Dropped the `eslint-disable` and listed both explicitly so the lint rule and the closure's captured values stay honest.

**2. Related rename: `sessionId` → `workspaceId` across `src/app/`**
The repo already migrated the public concept (`resolveWorkspaceId`, `/api/workspaces/...`), but the frontend code still used `sessionId` variable names and a `SessionNames` type, which read as inconsistent next to the workspace-prefixed URLs the same files call. Word-boundary rename of:

- `sessionId` → `workspaceId` (function params, props, hook args, event detail keys, `useParams` generic, the React-Router placeholder `/canvas/:sessionId` → `/canvas/:workspaceId` in `App.tsx`)
- `SessionNames` → `WorkspaceNames`

Server route paths (`/api/sessions/:sessionId/...`) and the legacy compat field on the workspace list response are intentionally left as-is — those are wire format kept by `registerWorkspaceAlias` for backward compatibility, not internal naming.

**Verification**
- `pnpm test` → 1202 passed (mcp-node + mcp-jsdom + mcp-browser)
- `pnpm typecheck` → OK
- Manually dogfooded the canvas-switch flow earlier in this PR; the route param rename was double-checked via Playwright by navigating between two canvases under the new `/canvas/:workspaceId/*` pattern.

PR #116, opened 2026-07-05
feat(shared): MigrationBundle contract + package export

kamiazya:
Wave 0 / S-K' slice of the apps/web completion plan (plan gate must-fix applied: defines a NEW contract, not a move — no migrationBundleSchema previously existed).

- New \`packages/mcp-server/src/shared/migration-bundle.ts\`: \`migrationBundleSchema\` (Zod, \`.strict()\` at every level), \`MigrationBundle = z.infer\`. Shape: format/version/sourceProvider literals, \`createdAt\` (ISO datetime), \`canvases[]\` with per-canvas scene. Array form kept; length===1 enforcement is S-L's job.
- Exported as \`@kamiazya/whiteboard-mcp/migration-bundle\` — both the workspace \`exports\` and \`publishConfig.exports\` blocks updated, with \`publish-contract.test.ts\` expectations updated red-first.
- Browser-safe: zod-only imports; added to the \`web-app-boundary\` shared allowlist.
- apps/web: source-resolving alias in vite/vitest/vitest.browser configs + \`buildMigrationBundle()\` pure function (injectable clock) — UI wiring is S-M.

TDD red-first at each layer. mcp-node 69 tests + apps/web 176 tests + typecheck + build all pass. Review: 0 CRITICAL/HIGH; 2 LOW (createdAt now ISO-validated, elements comment) fixed in follow-up.

### DIFF
--- a/apps/web/src/lib/migration-export.test.ts
+++ b/apps/web/src/lib/migration-export.test.ts
@@ -1,0 +1,57 @@
+import { describe, expect, it } from 'vitest'
+import { migrationBundleSchema } from '@kamiazya/whiteboard-mcp/migration-bundle'
+
+import type { CanvasSnapshot } from './whiteboard-client.js'
+import { buildMigrationBundle } from './migration-export.js'
+
+function makeSnapshot(overrides: Partial<CanvasSnapshot> = {}): CanvasSnapshot {
+  return {
+    id: 'canvas-1',
+    name: 'My Canvas',
+    scene: { elements: [{ id: 'el-1', type: 'rectangle' }] },
+    updatedAt: '2026-07-01T00:00:00.000Z',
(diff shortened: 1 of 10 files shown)


kamiazya:
AI review triage (Gemini):
- **[MEDIUM] defensive optional chaining on snapshot.scene.elements** — **skipped (with reason)**: `CanvasSnapshot.scene: { elements: unknown[] }` is a required-typed shape and `buildMigrationBundle` is a pure function over already-validated snapshots. The validation boundary is the store read (caller's responsibility), and the import side (S-L) revalidates the whole bundle via `migrationBundleSchema.parse()`. Adding fallbacks here would mask the real malformed-data source rather than fix it, against the repo's Zod-single-source discipline.

PR #117, opened 2026-07-05
fix(web): reconnect useCanvasSync on backend swap so reload keeps elements

kamiazya:
Wave 0 top-priority slice — fixes the HIGH reload-loses-elements bug confirmed in production (kamiazya-whiteboard.pages.dev): draw a rectangle, reload, canvas is empty.

## Root cause

`useCanvasSync`'s connect effect had an empty dependency array, so it bound once to the placeholder backend (`BrowserLocalBackend('__placeholder__')` passed while the snapshot was still loading) and never reconnected to the real backend. Every edit was written to the placeholder Loro doc; the real canvas row stayed `elements: []`, so reload restored nothing.

## Fix (human decision: reconnect approach — reusable for future in-place browser-local→daemon migration)

- `useCanvasSync` now accepts `CanvasBackend | null`, and its connect/disconnect effect depends on `[backend]`: null → idle (no writes); backend identity change → disconnect old + full per-connection state reset (doc/undo/files caches, applyGeneration bump, pending debounce cancelled) before connecting the new one.
- Writes (local-update push + putFile) are bound to the per-connection backend and gated by a connection-generation guard, so late callbacks from a swapped-away backend are dropped (no cross-connection leak).
- Switch-time connect failure is scoped to the current generation: `syncStatus='error'`, the previous backend is not resurrected, no rejection escapes the effect.
- `BrowserLocalCanvasPage` passes the nullable memoized backend directly — the `__placeholder__` backend is gone.

## Tests

TDD red-first. jsdom (useCanvasSync): null→no-connect, null→connect, A→B swap (writes reach B only), A→B swap where B.connect fails. web-browser (real IndexedDB): draw → remount → element persists AND no `__placeholder__` row. Full apps/web jsdom 192 + browser 39 + typecheck green. Mutation-checked: reverting the deps fix / reintroducing the placeholder fails the reload test.

## Visual repro (built dist served with production headers)

Before reload (element drawn) → after reload (element restored):

### DIFF
--- a/apps/web/src/hooks/useCanvasSync.test.ts
+++ b/apps/web/src/hooks/useCanvasSync.test.ts
@@ -37,6 +37,7 @@ type FakeBackendControl = {
   handlers: CanvasBackendHandlers | null
   disconnectCalled: boolean
   pushLocalUpdateCalls: Uint8Array[]
+  putFileCalls: [string, unknown][][]
 }
 
 function makeFakeBackend(): CanvasBackend & { _ctrl: FakeBackendControl } {
@@ -44,6 +45,7 @@ function makeFakeBackend(): CanvasBackend & { _ctrl: FakeBackendControl } {
     handlers: null,
     disconnectCalled: false,
     pushLocalUpdateCalls: [],
+    putFileCalls: [],
   }
   const backend: CanvasBackend & { _ctrl: FakeBackendControl } = {
     _ctrl: ctrl,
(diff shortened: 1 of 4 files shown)


kamiazya:
AI review triage (Gemini):
- **[MEDIUM] void putFile → unhandled rejection** (useCanvasSync.ts:182) — **fixed**: a file-upload failure is non-fatal (elements persist via the separate Loro-commit path), but void-ing the promise let a rejection surface unhandled. Now `.catch()`-ed and logged.
- **[MEDIUM] import useRef in BrowserLocalCanvasPage** — **skipped**: the page already keeps a stable backend instance via `useMemo([canvasId])`; it uses no ref and needs none, so importing useRef would be dead code. (useCanvasSync itself already imports/uses useRef.)

PR #120, opened 2026-07-05
feat(web): editable canvas title with browser-local persistence

kamiazya:
Wave 1 slice S-B of the apps/web completion plan. The canvas title was a static `<h1>`; now it is editable and the rename persists through the browser-local store, surviving reload.

## Change

- New `CanvasTitle` component: a real `<input aria-label="Canvas title">` for editing plus a visually-hidden `<h1>` mirror so the page keeps a heading landmark. Enter commits, Escape reverts, blur commits. Keyboard events are `stopPropagation`-guarded so Enter/Escape/Backspace/Delete typed into the title never reach Excalidraw's document-level shortcuts (e.g. Delete won't nuke elements).
- `renameCanvas(name)` on the browser-local controller writes through the **same** pendingSnapshot/flushSave state machine as scene edits (no parallel storage path), merging with any pending scene edit, and flushes immediately so a rename never lingers as "Unsaved changes".
- Empty / whitespace-only input normalizes to `untitled` before it reaches the store — a blank name is never persisted.
- Existing header layout (flex column, beta banner, BackendConfigChip, Delete button, save-status) intact; the title's accessible name (`/canvas title/i`) is kept distinct from Delete (`/delete/i`) so existing role queries don't collide.

## Tests (TDD red-first)

jsdom 209 + web-browser 43 + typecheck all pass. Coverage: controller rename (persist, whitespace→`untitled`, no empty write, concurrent scene+rename merge, flush-races-unmount), CanvasTitle interaction (Enter/Escape/blur/fallback/heading-present), and a real-IndexedDB `.rename.browser.test.tsx` proving reload persistence + keyboard-isolation-doesn't-delete-elements. Review: 0 CRITICAL/HIGH/QA-fail.

## Visual repro (built dist served with production headers)

Edited title (Saved) → after reload (retained):

### DIFF
--- a/apps/web/src/components/canvas-title/CanvasTitle.test.tsx
+++ b/apps/web/src/components/canvas-title/CanvasTitle.test.tsx
@@ -1,0 +1,68 @@
+import { describe, expect, it, vi, afterEach } from 'vitest'
+import { render, screen, cleanup, fireEvent } from '@testing-library/react'
+import { CanvasTitle } from './CanvasTitle.js'
+
+describe('CanvasTitle', () => {
+  afterEach(() => cleanup())
+
+  it('renders the current name in a labeled textbox', () => {
+    render(<CanvasTitle value="My canvas" onRename={vi.fn()} />)
+    const input = screen.getByRole('textbox', { name: /canvas title/i }) as HTMLInputElement
+    expect(input.value).toBe('My canvas')
+  })
+
+  it('typing then pressing Enter commits the rename with the typed value', () => {
+    const onRename = vi.fn()
+    render(<CanvasTitle value="My canvas" onRename={onRename} />)
+    const input = screen.getByRole('textbox', { name: /canvas title/i })
+    fireEvent.change(input, { target: { value: 'Renamed via Enter' } })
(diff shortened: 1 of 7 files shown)


kamiazya:
AI review triage (Gemini):
- **[HIGH] input desync when value is normalized (empty→untitled)** — **fixed, but not via the suggested useEffect**: a `draft←value` useEffect regressed real tests — the snapshot loads asynchronously and pushes a new `value` while the user is mid-edit, so the effect clobbered in-progress typing. Instead, `commit()` now reflects the normalized value back into the field directly (empty→'untitled' shows in the input) without touching the async-load path. Added regression tests for both the normalization display AND the no-clobber-while-typing case.
- **[HIGH] renameCanvas should advance snapshotRef synchronously** — **applied**: snapshotRef is the fallback merge base (read when pendingSnapshotRef is null after a flush), so it now advances synchronously alongside pendingSnapshotRef to avoid a follow-up edit merging onto a pre-rename base.
- **[MEDIUM] import useEffect / value-sync test** — superseded by the commit-based fix above (no useEffect needed); the equivalent coverage is the two new CanvasTitle tests.

PR #121, opened 2026-07-05
refactor(web): make Loro the single source of truth for canvas elements

kamiazya:
Wave 1 slice S-Data of the apps/web completion plan. Removes the element double-write between the JSON `canvases` row and the Loro doc: elements are now canonical **only** in `loroCanvases`, and the JSON row is demoted to metadata.

## Change

- `canvasSnapshotSchema` / `CanvasSnapshot` narrowed to metadata-only `{ id, name, updatedAt }` — `scene` dropped. `CanvasSnapshot` is now `z.infer<typeof canvasSnapshotSchema>` so the previously hand-written parallel interface can't drift.
- Extracted the duplicated IDB opener + `DB_VERSION` + `onupgradeneeded` into one shared `apps/web/src/lib/browser-idb.ts`, consumed by both `browser-local-store` and `loro-store`. This makes "bump both together" **structural** (one `DB_VERSION`) instead of a hand-synced comment — a single opener can't drift into a VersionError.
- `DB_VERSION` 2→3 with a v2→v3 migration that strips the legacy `scene` field from existing `canvases` rows; `loroCanvases` (the element source of truth) is untouched.
- Removed the dead `controller.updateScene` JSON-element write path (the page never called it; elements flow through Loro via useCanvasSync). The pending→saving→saved persistence machine + metadata save path are intact.
- Reconciled `buildMigrationBundle` (S-K'): since the snapshot no longer carries `scene`, it now takes `{ snapshot, elements }` pairs where the caller supplies elements from Loro — the bundle contract shape (`scene:{elements}`) is unchanged.

## Verification

- apps/web jsdom 206 + web-browser 46 + typecheck all pass. The web-browser suite includes a **real-IndexedDB v2→v3 migration test** (seed a v2 fixture with `scene` on the JSON row + a populated loro envelope → open at v3 → no VersionError, scene stripped, loroCanvases canonical, no orphan rows) and the existing draw→reload element-persistence + loro-store-is-source-of-truth guards.
- **mutation-checked**: reverting `DB_VERSION` 3→2 fails the migration test.
- Manual (wrangler pages dev): confirmed the persisted `canvases` row is metadata-only (`id,name,updatedAt`, no `scene`) and the DB opens at version 3. (Element round-trip is proven by the deterministic browser regression tests rather than a headless pointer draw.)

Review: 0 CRITICAL/HIGH/QA-fail.

## Summary by CodeRabbit

* **New Features**
  * Canvas data now uses a simpler snapshot format, with canvas contents handled separately for more reliable saves and exports.
  * Editing changes are saved immediately for actions like renaming, reducing delays before updates persist.

* **Bug Fixes**
  * Improved browser database upgrades so older saved data is migrated safely without showing legacy content.
  * Fixed several IndexedDB edge cases to better detect missing, corrupted, or outdated local data.
  * Prevented cleanup flows from accidentally restoring removed canvas data.

### DIFF
--- a/apps/web/src/lib/browser-idb-migration.browser.test.tsx
+++ b/apps/web/src/lib/browser-idb-migration.browser.test.tsx
@@ -135,6 +135,39 @@ describe('whiteboard IndexedDB v2 -> v3 upgrade', () => {
     expect(await metaStore.getDefaultCanvasId()).toBe(canvasId)
   })
 
+  it('upgrades without aborting when a legacy canvases row is a non-object (corrupt data)', async () => {
+    // A null / non-object row must not throw a TypeError from `'scene' in value`
+    // inside the upgrade cursor — that would abort the transaction and brick the
+    // DB open for the user.
+    await new Promise<void>((resolve, reject) => {
+      const req = indexedDB.open('whiteboard', 2)
+      req.onupgradeneeded = (event) => {
+        const db = (event.target as IDBOpenDBRequest).result
+        if (!db.objectStoreNames.contains('meta')) db.createObjectStore('meta')
+        if (!db.objectStoreNames.contains('canvases')) db.createObjectStore('canvases')
+        if (!db.objectStoreNames.contains('loroCanvases')) db.createObjectStore('loroCanvases')
+      }
+      req.onsuccess = () => {
+        const db = req.result
+        const tx = db.transaction('canvases', 'readwrite')
+        // A corrupt non-object value stored under a key.
+        tx.objectStore('canvases').put(null, 'corrupt-row')
+        tx.oncomplete = () => {
+          db.close()
+          resolve()
+        }
+        tx.onerror = () => reject(tx.error)
+      }
+      req.onerror = () => reject(req.error)
+    })
+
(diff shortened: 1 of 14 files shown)


kamiazya:
AI review triage (Gemini) — all applied:
- **[HIGH] `'scene' in value` TypeError on non-object row** (browser-idb.ts) — **fixed**: guarded with `typeof value === 'object' && value !== null` so a corrupt/null legacy row can't abort the upgrade and brick the DB open. Added a regression test (seed a non-object canvases row → open at v3 → upgrade still completes) and mutation-checked it (removing the guard fails the test).
- **[MEDIUM] skip migration when oldVersion === 0** — **applied**: fresh install has an empty store, so `oldVersion > 0` guards the no-op scan.
- **[MEDIUM ×4] unhandled onabort → promise never settles** (setDefaultCanvasId / IndexedDBStore.save / removeCanvas / LoroStore.save) — **all applied**: each transaction now rejects + closes the db on abort, so a mid-write abort (e.g. QuotaExceeded) surfaces instead of hanging the caller's await.

apps/web jsdom 206 + web-browser 47 + typecheck green.

PR #122, opened 2026-07-05
feat(web): id-addressed multi-canvas foundation for browser-local storage

kamiazya:
Wave 1 slice S-C1 of the apps/web completion plan. Extends the browser-local persistence layer from a single-default-pointer model to an id-addressed multi-canvas foundation. **API-layer only — no UI** (canvas list / switcher is S-C2).

## Change

- `listCanvases()` / `createCanvas(name?)` / `switchCanvas(id)` added to `BrowserLocalStore` (both MemoryStore + IndexedDBStore) and the controller.
  - `listCanvases` openCursors the `canvases` store and hydrates each row through `canvasSnapshotSchema.safeParse`, skipping corrupt rows (one parse boundary, one bad row can't blank the whole list).
  - `createCanvas` issues a fresh id, writes the metadata row AND seeds an empty Loro doc for that id; on Loro-write failure it rolls back the orphan metadata row (mirrors `startFresh`).
  - `switchCanvas` flushes the current canvas first, then loads the target by id; unknown id is a safe no-op.
- `defaultCanvasId` demoted to a "last-opened" hint; all element/metadata access is by explicit id. **No workspace/slug identity** (YAGNI, per plan).
- **No DB_VERSION bump**: `canvases` / `loroCanvases` are already id-keyed maps, so holding many canvases needs no schema change. S-Data's v3 schema + migration are untouched.

## Verification

- apps/web jsdom 223 + web-browser 52 + typecheck all pass. The web-browser suite adds a real-IndexedDB multi-canvas test proving: createCanvas×2 → listCanvases returns both by real id, and two canvases hold **independent** elements in `loroCanvases` (writing to A never leaks into B). S-Data's v2→v3 migration test still passes unchanged.
- Review: 0 CRITICAL/HIGH/QA-fail; the dev-loop's triage/fix loop already hardened switchCanvas failure/degraded handling and a flush-waiter race.

This is invisible to end users (no UI yet), so test output is the evidence; the switcher UI + live switch flow land in S-C2 (Wave 1 final slice).

## Summary by CodeRabbit

* **New Features**
  * Added support for listing existing canvases, creating new canvases, and switching between canvases.
  * New canvases now start as empty documents and can be saved and reopened reliably.

* **Bug Fixes**
  * Improved handling of damaged or outdated saved canvas data so it’s skipped instead of breaking the canvas list.
  * Made canvas switching more reliable when saves are in progress or storage temporarily fails.

### DIFF
--- a/apps/web/src/lib/browser-local-store.test.ts
+++ b/apps/web/src/lib/browser-local-store.test.ts
@@ -1,6 +1,6 @@
-import { describe, expect, it, beforeEach } from 'vitest'
 import { IDBFactory } from 'fake-indexeddb'
-import { MemoryStore, IndexedDBStore } from './browser-local-store.js'
+import { beforeEach, describe, expect, it } from 'vitest'
+import { IndexedDBStore, MemoryStore } from './browser-local-store.js'
 import type { CanvasSnapshot } from './whiteboard-client.js'
 
 const snap: CanvasSnapshot = {
@@ -87,6 +87,22 @@ describe('MemoryStore', () => {
     expect(a.length).toBeGreaterThan(0)
     expect(a).not.toBe(b)
   })
+
+  it('listCanvases returns empty array when store is empty', async () => {
+    const store = new MemoryStore()
+    expect(await store.listCanvases()).toEqual([])
+  })
+
+  it('listCanvases returns all saved snapshots by id', async () => {
+    const store = new MemoryStore()
+    const a: CanvasSnapshot = { ...snap, id: 'c1', name: 'Canvas A' }
+    const b: CanvasSnapshot = { ...snap, id: 'c2', name: 'Canvas B' }
+    await store.save(a)
+    await store.save(b)
+    const list = await store.listCanvases()
+    expect(list).toHaveLength(2)
+    expect(list).toEqual(expect.arrayContaining([a, b]))
+  })
 })
 
 describe('IndexedDBStore', () => {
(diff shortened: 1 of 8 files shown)


kamiazya:
AI review triage (Gemini):
- **[MEDIUM] listCanvases unhandled onabort** — **applied**: the readonly transaction now rejects + closes the db on abort, consistent with the other store transactions (a mid-read abort no longer hangs the awaiting caller).
- **[HIGH ×2] switchCanvas rapid-switch race guard** — **not applied in this slice, deferred with reasoning**. I prototyped the generation guard but verified it does not manifest here and could not be given a meaningful regression test, so shipping it would be unverified concurrency code:
  1. **Doesn't manifest under real IndexedDB**: same-store ops queue FIFO, so the earlier-started switch resolves first and the last-requested switch applies last (wins). A real-IDB A→B rapid-switch browser test passes *without* any guard.
  2. **No concurrent caller in S-C1** — it is API-layer only; the switch UI (real concurrent caller) is S-C2.
  3. A jsdom mutation-check for the guard hung under act + concurrent-async and could not deterministically distinguish guard-present vs absent, so per YAGNI + "test the code you keep" I reverted it rather than ship untested concurrency machinery.
  The guard becomes genuinely load-bearing for the **Wave 2 daemon backend** (network responses are truly out-of-order, unlike FIFO IDB) and should land there — or in S-C2 with the switch UI — with a proper test. Tracked locally for that work.

PR #123, opened 2026-07-05
feat(web): multi-canvas UI — canvas switcher and New-canvas control

kamiazya:
Wave 1 final slice S-C2 of the apps/web completion plan. Wires the S-C1 store/controller API (listCanvases / createCanvas / switchCanvas) into the UI, so browser-local apps/web can hold and switch between multiple canvases. **This completes Wave 1 (First Step).**

## Change

- A canvas **switcher** (`<select aria-label="Canvases">`) and a **New canvas** button in the page header. Selecting an entry calls `switchCanvas(id)`; New calls `createCanvas()` then switches to it (name `untitled`, immediately renameable via the S-B title editor).
- The switcher marks the current canvas selected and synthesizes a fallback option so its controlled value always matches an option (no flash during the async list refresh). A stale `listCanvases()` resolution can't clobber a newer one (generation guard).
- New-canvas failure surfaces an inline error instead of an unhandled rejection.
- **Title-on-switch fix (caught in manual verification):** `CanvasTitle` is now keyed by canvas id, so switching canvases remounts it and reseeds the editable title from the new canvas's name. Without it the field kept showing the previous canvas's name (the heading was correct; the editor's internal draft was stale — CanvasTitle deliberately doesn't resync `value` to avoid clobbering in-progress typing).

## Verification

- apps/web jsdom 229 + web-browser 53 + typecheck all pass. Coverage: switcher lists + marks current, switch calls switchCanvas once, New creates+switches to untitled (title field asserted, **mutation-checked** — removing the key fails it), select-value-always-has-an-option, generation-guard drop, create-failure error surface, and a real-IndexedDB test proving draw-on-A → New→empty-B → switch-back-to-A restores A's element end to end (incl. the useCanvasSync reconnect path).

## Visual repro

Canvas A → New canvas shows `untitled` (title correctly updated) → switch back to A:

## Known limitation (tracked, follow-up)

Drawing then switching/creating a canvas **within ~300ms** can lose that last edit — a pre-existing `useCanvasSync` debounce behavior (zero diff before this slice) that the switch UI is the first to expose. A flush-on-teardown fix was attempted but could not be verified end-to-end in a real-IndexedDB browser test, so it was reverted rather than shipped half-working. Narrow window; deferred as the next fix after Wave 1.

## Summary by CodeRabbit

* **New Features**
  * Added clearer in-app feedback when creating a new canvas fails.
* **Bug Fixes**
  * Improved canvas switching so async canvas list refreshes can’t overwrite newer selections.
  * Kept the canvas selector stable while canvases load/refresh and ensured the title/editor remounts for the active canvas.
  * Improved persistence during backend teardown by flushing pending debounced edits/uploads so changes aren’t lost or routed to the wrong backend.
* **Tests**
  * Expanded browser and hook test coverage for multi-canvas persistence and flush-on-teardown behavior.

### DIFF
--- a/apps/web/src/hooks/useCanvasSync.test.ts
+++ b/apps/web/src/hooks/useCanvasSync.test.ts
@@ -189,7 +189,11 @@ describe('useCanvasSync', () => {
     expect(backend._ctrl.disconnectCalled).toBe(true)
   })
 
-  it('does not push after unmount when debounce timer fires', async () => {
+  // Deliberate behavior change (flush-on-teardown): a pending debounced scene
+  // edit is now flushed synchronously during teardown cleanup instead of
+  // being cancelled and lost. React runs the same cleanup path for a real
+  // unmount and a backend switch, so unmount also flushes.
+  it('flushes a pending debounced write to the backend on unmount instead of dropping it', async () => {
     const backend = makeFakeBackend()
     const api = makeApiStub()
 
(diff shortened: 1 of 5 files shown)


kamiazya:
Update: the known scene-loss limitation is now **fixed** in this PR (commits `flush pending scene edit on canvas backend teardown` + dedupe). `useCanvasSync`'s connect-effect cleanup now flushes the pending 300ms-debounced scene write into the outgoing connection's doc before disconnecting (bound to the effect-scoped backend, so it persists to the origin canvas and never leaks to the next one). Added a real-IndexedDB browser regression test (draw within the debounce window → New canvas → switch back → the edit is present), **mutation-checked**: reverting the cleanup flush to the old cancel() makes it fail (`late-edit` element lost). jsdom 230 + web-browser 54 + typecheck green.

kamiazya:
CodeRabbit review triage — all 3 applied:
- **[Major] No guard against rapid "New canvas" clicks** — **fixed**: the button is now disabled while a create/switch is in flight (`isCreatingCanvas` + `.finally`), so concurrent `createCanvas()` calls can't mint orphaned rows. Regression test added and mutation-checked (removing `disabled` fails it).
- **[Major, outside diff] clearDb missing onblocked** — **fixed**: `clearDb()` in the multi-canvas browser test settles on `onblocked` too, so a not-yet-closed prior connection fails clearly instead of hanging the suite on timeout.
- **[Minor] teardown assertions race the local-update microtask** — **fixed**: both flush-on-teardown unit tests now await the microtask (`await act(async () => await Promise.resolve())`) before asserting `pushLocalUpdate` landed.

jsdom 231 + web-browser 54 + typecheck green.

kamiazya:
Gemini review triage — all applied (some overlapped with the CodeRabbit round already merged):
- **:63 / :221 add isCreating + disable New button** — already done in the prior commit (`isCreatingCanvas` + `.finally`, regression-tested).
- **:204 disable the switcher during create** — **applied**: the `<select>` is now also disabled while a create/switch is in flight, so a concurrent switch can't race the create.
- **:85 listCanvases in effect deps / drop eslint-disable / clear stale createError** — **applied**: listCanvases is a stable `useCallback([])`, so adding it is behaviour-preserving and lint-honest (eslint passes with no disable); createError is now cleared when the active canvas changes.
- **:9 / :192 import fireEvent at top** — **applied**: removed the dynamic import.

jsdom 231 + web-browser 54 + typecheck + eslint green.

PR #129, opened 2026-07-06
feat(web): capability-gated daemon-feature teasers in apps/web

kamiazya:
First near-term slice of the **UI unification** initiative (ADR 0001 capability-gating step). Wires `WhiteboardCapabilities` into the apps/web UI and surfaces daemon-only features as **disabled teasers** with an upsell tooltip, so browser-local users discover what a local daemon unlocks — without porting the features yet (that's Stage 2).

## Change

- **Capability model**: `WhiteboardCapabilities` extended with additive `branches` / `merge` booleans (existing fields untouched). `BROWSER_LOCAL_CAPABILITIES` = false, `LOCAL_DAEMON_CAPABILITIES` = true. Kept a plain type + static per-kind const map (no Zod — it's an in-process derived constant, not a cross-process/persisted contract; noted in a comment so the Zod-discipline gate doesn't flag it).
- **Teasers**: new `CapabilityTeaser` component renders four daemon-only controls — **Version history / Workspaces / Branches / Merge** — in the page header. Under browser-local caps each is `aria-disabled` (NOT native `disabled`, so it stays keyboard-focusable) with an always-present `aria-describedby` description and a Radix tooltip reading **"Connect a local daemon (MCP) to enable <feature>"**. When caps flip to daemon, the disabled/teaser treatment drops.
- **Mode indicator stays read-only**: `BackendConfigChip` remains the fixed overlay showing "Browser only"; no runtime mode-switch control was added (per the decision that mode is config/deploy-time, auto-detected by the Wave 2 LNA probe).
- Header wraps / uses an overflow container so it never scrolls horizontally at 320–375px (asserted in a browser test).

## Verification

- apps/web jsdom 243 + web-browser 57 + typecheck all pass. Coverage: provider caps (branches/merge false for local, true for daemon), teasers aria-disabled + accessibly-described under local caps and enabled under daemon caps (mutation-checked across the flip), App passes `state.capabilities` down, a negative test asserts no mode-switch control, a narrow-viewport no-overflow browser test, and real-browser Radix tooltip coverage.
- Manual (wrangler pages dev): all four teasers render `aria-disabled=true`; the tooltip appears on **both hover and keyboard focus**.

## Visual repro

Header with disabled teasers, and the upsell tooltip:

## Scope / follow-ups

Out of scope (Stage 2): actual versions/branches/merge/workspaces features, DaemonBackend relocation + wiring, sync-hook consolidation, CanvasPage/routing port, daemon UI deletion (see `tmp/notes/ui-unification-stage2-plan.md`, 14-slice plan). A pre-existing "setState while rendering Excalidraw" React warning surfaced in browser tests (not from this slice) — tracked separately.

## Summary by CodeRabbit

* **New Features**
  * Added visible capability teasers on the canvas for version history, workspaces, branches, and merge.
  * Teasers now explain when a local daemon is needed and become active when those capabilities are available.

* **Bug Fixes**
  * Capability availability is now passed through the app correctly, so the canvas reflects the right feature state.
  * Improved accessibility and interaction behavior for disabled teaser controls.

### DIFF
--- a/apps/web/src/App.test.tsx
+++ b/apps/web/src/App.test.tsx
@@ -1,15 +1,20 @@
 import { cleanup, render, screen } from '@testing-library/react'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { App } from './App.js'
-import type { ProviderState } from './lib/provider.js'
+import type { ProviderState, WhiteboardCapabilities } from './lib/provider.js'
 
 afterEach(cleanup)
 
-// BrowserLocalCanvasPage pulls in Excalidraw/loro-crdt which need a real
-// browser (roughjs native bindings, WASM). Mock it here since this suite
-// only exercises the backend-configuration chip, not the canvas itself.
+// Records the props BrowserLocalCanvasPage receives so tests can assert
+// capabilities actually flow from App down to the page, not just that the
+// page mounts. BrowserLocalCanvasPage pulls in Excalidraw/loro-crdt which
+// need a real browser (roughjs native bindings, WASM), so it stays mocked.
+let receivedCapabilities: WhiteboardCapabilities | undefined
 vi.mock('./pages/BrowserLocalCanvasPage.js', () => ({
-  BrowserLocalCanvasPage: () => <div data-testid="browser-local-canvas-page" />,
+  BrowserLocalCanvasPage: ({ capabilities }: { capabilities?: WhiteboardCapabilities }) => {
+    receivedCapabilities = capabilities
+    return <div data-testid="browser-local-canvas-page" />
+  },
 }))
 
 const BROWSER_LOCAL_STATE: ProviderState = {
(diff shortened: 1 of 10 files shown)


kamiazya:
AI review triage (Gemini): **applied** — the disabled teaser's onClick already called preventDefault, but aria-disabled still *bubbles* the click, so a teaser inside a clickable ancestor could trigger it. Added event.stopPropagation() in the disabled branch + a regression test (mutation-checked: removing stopPropagation fails it). jsdom 244 + typecheck green.

kamiazya:
CodeRabbit triage (Radix aria-describedby merge): **verified, not reproduced — locked with a test.** The concern is real for Radix in general (issue #2598), but I verified empirically in a real-browser test that in the **closed** state the teaser button's aria-describedby still resolves to the sr-only guidance text (Radix only composes its own describedby on open, and does not clobber the manual one while closed). So the always-available description is not lost: closed → the sr-only description via aria-describedby; open → the Radix tooltip content (same text). Added a browser regression test asserting the closed-state linkage so a future Radix upgrade that changed this would fail CI. browser 58 + typecheck green.

PR #133, opened 2026-07-06
fix(web): allow browser-local mode on Cloudflare Pages preview origins

kamiazya:
## Bug

`https://latest.kamiazya-whiteboard.pages.dev/` (and every per-PR preview) rendered **"Runtime configuration is invalid."** instead of the app.

`resolveHostedProviderStateFromRaw` hard-rejected any browser origin classified as a Pages **preview** (`*.kamiazya-whiteboard.pages.dev` other than the exact production origin). That guard predates the preview environments — now that `latest` + per-PR previews are a supported surface, it blocked exactly what they exist for.

## Fix

- Preview browser origins now resolve to **browser-local** mode (offline, origin-agnostic — safe on any origin).
- The original security intent is **kept**: a `daemonBaseUrl` config on a preview origin is still refused (`invalid-config`), so a preview deploy can never reach a daemon. The error message still leaks no origin/URL fragments.
- `publicOrigin` (config-declared) validation is unchanged — declaring a preview `publicOrigin` in config remains a config error.

## Verification

- Red-first: new tests (preview / `latest.` alias origin → `browser-local`; preview + daemon config → `invalid-config`, no origin leak) failed against the old impl, pass now. provider 27 / jsdom 246 / browser 58 / typecheck green.
- **Live proof**: this PR's own Cloudflare preview (comment below) should render the whiteboard editor instead of the error — and after merge, `latest.kamiazya-whiteboard.pages.dev` will too.

### DIFF
--- a/apps/web/src/lib/provider.test.ts
+++ b/apps/web/src/lib/provider.test.ts
@@ -159,11 +159,28 @@ describe('resolveHostedProviderStateFromRaw', () => {
     }
   })
 
-  // browserOrigin guard: preview browser origin is rejected even with no runtime config,
-  // so a Cloudflare Pages preview deploy cannot silently enter browser-local mode.
-  it('preview browserOrigin returns invalid-config even with empty runtime config', () => {
+  // browserOrigin policy: preview deploys (latest.<project>.pages.dev, per-PR
+  // branch aliases, hash previews) run in browser-local mode — offline and
+  // origin-agnostic — but must never connect to a daemon from a preview origin.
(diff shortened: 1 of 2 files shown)


kamiazya:
**Live proof on this PR's own preview** (`fix-preview-origin.kamiazya-whiteboard.pages.dev`): the editor renders — no "Runtime configuration is invalid" — with the canvas title editor and teasers present. Verified via Playwright (invalidConfigError: false, editorRendered: true, titleEditor: true).

PR #134, opened 2026-07-06
ci: align GitHub Environments with deployed URLs and release previews on PR close

kamiazya:
Follow-up to the preview environments: makes the **Environments UI match reality** and **releases preview deployment records when a PR closes**.

## Changes

**1. Real URLs on deployment records** (Environments sidebar "View deployment" now works):
- `production-web` (release.yml deploy-web) → `https://kamiazya-whiteboard.pages.dev`
- `preview-web` latest job → `https://latest.kamiazya-whiteboard.pages.dev`
- `preview-web` per-PR deploy → that PR's branch-alias URL (`environment.url` resolves at job end, so it uses the deploy step's output)

**2. `preview-cleanup.yml`** — on PR close (merged or not): finds this PR's preview-web deployment records by their `environment_url` (branch-alias hostname, recomputed with the same sanitize rules as the deploy path), marks them **inactive**, then **deletes** them. So closed PRs no longer leave stale "Active" previews in the Environments view.
- Deliberately **no environment binding and no secrets**: the cleanup run itself never mints a new deployment record, and it only touches GitHub records via the workflow token (`deployments: write`).
- Cloudflare-side preview files are inert statics (overwritten per push to the same alias) and are left to Cloudflare's retention.

**3. Guard update**: `web-app-boundary` now accepts the `environment: {name, url}` mapping form for deploy-web (mutation-checked — dropping the environment declaration still fails the test). 42 mcp-node tests pass.

## Notes
- Records created *before* this change have no `environment_url`, so cleanup can't match them; they can be deleted once by hand (or just age out as new deploys supersede them).
- Untrusted branch names are read via `env:` and sanitized to `[a-z0-9-]`; no `pull_request_target`.

### DIFF
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -26,7 +26,10 @@ permissions:
 jobs:
   deploy-latest:
     runs-on: ubuntu-latest
-    environment: preview-web
+    environment:
+      name: preview-web
+      # Surfaces the real alias in the Environments UI ("View deployment").
+      url: https://latest.kamiazya-whiteboard.pages.dev
     env:
       CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
       CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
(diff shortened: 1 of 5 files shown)


kamiazya:
Gemini triage — **applied**: the environment-guard regex now tolerates optional quotes and any key order in the `environment:` mapping (both flagged variations), while still failing on a wrong or missing environment. Validated against 5 positive + 3 negative fixtures; suite 42/42 green. (Kept the string-matching style consistent with this file's other workflow guards rather than adding a YAML-parser dependency for one assertion.)

PR #136, opened 2026-07-06
feat(web): make useCanvasSync capability-complete for daemon wiring

kamiazya:
UI-unification Stage 2, slice 3 (ADR 0001). Makes apps/web's `useCanvasSync` the canonical, capability-complete sync hook: the six formerly-no-op daemon handlers are functionally wired, so the upcoming CanvasPage port (slices 9–10) and daemon wiring (slice 11) consume one hook instead of the frozen `useWhiteboardSync` fork.

## Change

- **Daemon receptors wired** (all guarded by the existing per-connection generation check): `onVersionCreated` / `onHeadChanged` → optional consumer callbacks (try/catch-guarded); `onRestoreStarted/Complete` → `restoreInProgress`/`restoreLabel` state + undo clear; `onViewportRequest` → fit/move handling; `onExportRequest` → queue-before-api-ready + flush + `sendExportResponse` (Zod-validated); `onAuthError` → error status. `sendClientReady` on connect and api-ready. All are no-ops under `BrowserLocalBackend` — browser-local behavior unchanged.
- **New return surface**: `restoreInProgress` / `restoreLabel` / `clearLocalUndo` — the direct observable product of the restore wiring; consumed by the CanvasPage-port slice. `BrowserLocalCanvasPage` is source-compatible unmodified.
- **putFile ordering unified**: scene commits are serialized on a commit chain; a firing with uploads awaits settlement before committing (upload failure stays non-fatal). Fork-diff audit in the dev-loop record; `useWhiteboardSync` (src/app) is **frozen** — `packages/mcp-server` has zero diff.
- **Chain robustness (review follow-ups, red-first)**: a throwing commit fails only its own firing (unguarded it rejected the chain and silently skipped every later commit for the session); putFile is raced against a 15s deadline so a hung upload can't block all later edits; consumer callbacks can't throw through backend handlers; the `as never` version-payload test was masking real contract drift and is now typed against `z.infer`.

## Verification

apps/web jsdom 277 ✓ / web-browser 58 ✓ (reload-elements + multi-canvas unchanged) / tsc ✓ / `pnpm --filter @kamiazya/whiteboard-mcp typecheck` ✓ with **zero mcp-server diff** (freeze respected). New coverage: receptor delivery + stale-generation drops for restore/viewport/export, export routing across a backend switch, commit-chain serialization, chain-poison and hung-upload regressions (both red-first).

## Noted follow-ups (not blocking)

Three MEDIUM coverage gaps from review remain on daemon-only branches that nothing consumes until slice 11 (malformed export_response payload branch, flushPendingExportRequests failure path, pending-exports-drop on mid-flight switch); tracked for the slice 11 test pass where a real DaemonBackend exercises them.

## Summary by CodeRabbit

* **New Features**
  * Added support for exporting canvases through sync, including queued handling when the editor isn’t ready yet.
  * Improved canvas sync controls with restore progress/status indicators and a local undo reset action.

* **Bug Fixes**
  * Exported content now respects the selected frame, theme overrides, and minimum text sizing.
  * Syncing is more reliable across backend switches, with better handling for stale updates, upload failures, and pending operations.
  * Export and upload callbacks now behave more consistently during reconnections and timeouts.


kamiazya:
Gemini triage:
- **Timer leak in the putFile deadline race** — **applied**: the timeout id is captured and cleared in a `finally`, so a fast upload no longer leaves a live 15s timer per firing.
- **Viewport 'move' partial-appState simplification** — **verified and not applied**: the suggestion does not typecheck against this Excalidraw version — `updateScene`'s `appState` is typed as the full picked shape, not `Partial` (`Type 'Partial<Pick<AppState, ...>>' is not assignable...`). The full-clone + mutable mapped type exists precisely to satisfy that typing; I documented the why in a comment so the next reader doesn't retry the same simplification.

tsc ✓ / jsdom 277 ✓ / browser 58 ✓.

PR #137, opened 2026-07-06
ci(web): add bundle-size gate to verify

kamiazya:
Implements the Q10 bundle-budget enforcement ahead of the Stage 2 daemon-UI port.

## What

- New `apps/web/scripts/smoke-bundle-size.mjs` (zero-dep, follows the existing `smoke-*.mjs` pattern): gzips built assets and asserts budgets — **entry JS ≤ 560 KB**, **CSS ≤ 30 KB**, and **`daemon-*.js` lazy chunks ≤ 40 KB** (skips until such chunks exist; once slices 5-8 ship daemon features behind `React.lazy`, this line keeps them out of first paint).
- Wired as `smoke:bundle-size` and a `verify` step right after the artifact smoke (dist already built there).

## Why these numbers

Measured today: entry 552.1 KB gzip / CSS 27.6 KB. The 560 KB ceiling is a **regression stop, not an endorsement** — the entry already exceeds the global <300 KB app budget (loro WASM glue + React + Excalidraw bootstrap), and shrinking it needs first-paint splitting tracked as a separate initiative. This gate prevents silent growth while that work is pending.

## Verification

Local run passes (entry 552.1/560, CSS 27.6/30, daemon chunk skip). Mutation-checked: lowering the entry budget to 500 KB fails the gate (exit 1); restored passes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -183,6 +183,9 @@ jobs:
       - name: Web app artifact smoke
         run: pnpm --filter @kamiazya/whiteboard-web smoke:artifact
 
+      - name: Web app bundle-size gate
+        run: pnpm --filter @kamiazya/whiteboard-web smoke:bundle-size
+
       - name: Web app preview-origin behavioral smoke
         run: pnpm --filter @kamiazya/whiteboard-web smoke:preview-origin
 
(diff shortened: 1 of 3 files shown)


kamiazya:
Gemini triage — both applied (same root issue): `existsSync` guard on `dist/assets` — a missing build now fails with an actionable one-liner (exit 1) instead of an ENOENT stack trace. Verified both paths locally: gate passes with dist present; exits 1 with the clean message when dist is absent.

PR #138, opened 2026-07-06
docs(adr): resolve ADR-0002 token story and add ADR-0004 unified CanvasPage

kamiazya:
Records two accepted Stage 2 design decisions as durable ADRs (docs only, no code changes).

- **ADR-0002 addendum** — closes the deferred token-delivery question: `runtimeConfigSchema` stays permanently token-free; token arrives via a dedicated `window.__WHITEBOARD_DAEMON_TOKEN__` global read once into an in-memory TokenStore (explicitly documented as serialization-surface reduction, not a security boundary); static Pages pairs via bootstrapToken → sessionToken (Stage 4); wire transport = WS subprotocol + HTTP Bearer only; local-daemon canvas GETs stay tokenless under the loopback trust model while `/api/runtime/*` keeps Bearer; ping's `pid` is replaced by a startup `instanceId` (preserving `server stop`'s mis-kill protection). The stale 'deferred to Stage 4 planning' bullet now points at the addendum.
- **ADR-0004** (0003 was taken) — the unified capability-gated CanvasPage: single page + `CanvasBackend` injection, capability-selected controllers, daemon overlays as `editing` sub-state, React Router with a shared `/canvas/:id` URL scheme, mode as read-only load-time status. Consequences honestly mark what is accepted-but-unshipped.

Both follow the existing MADR-lite template; index updated.

### DIFF
--- a/docs/contributing/adr/0002-browser-to-daemon-transport.md
+++ b/docs/contributing/adr/0002-browser-to-daemon-transport.md
@@ -47,4 +47,66 @@ Availability detection in `apps/web`:
 
 **Portless (`whiteboard.localhost`)** — Would avoid mixed-content for the hosted case. Rejected: `whiteboard.localhost` is not in `isLoopbackHostname`, so the daemon returns 403; deferred pending security review.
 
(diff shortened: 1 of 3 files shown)


kamiazya:
Gemini triage — **applied**: `/canvas/:id/*slug` corrected to `/canvas/:id/*` with the slug read via `params['*']` (React Router v6 has no named splat params; this also matches the daemon app's existing `/canvas/:workspaceId/*` route).

PR #139, opened 2026-07-06
feat(web): port pure-logic hooks (theme/dirty-state/fullscreen) to apps/web

kamiazya:
UI-unification Stage 2, slice 4 (ADR 0001/0004). Ports the four API-free pure-logic modules the unified CanvasPage needs, with tests, red-first per module:

- `useThemeMode` (+9 tests) — self-contained theme hook (localStorage `whiteboard:theme`, `prefers-color-scheme` subscription, `.dark` on `<html>`); flagged by architecture review as a hidden dependency of slices 9-10
- `useDirtyState` (+10) — window CustomEvent-based dirty tracking; event coupling kept as-is (the apps/web dispatcher is a slice 9-10 decision per ADR-0004's event-bus policy)
- `canvas-page-fullscreen` (+7) and `canvas-fullscreen-hash` (+5) — fullscreen detection + hash mirroring, relocated to `apps/web/src/lib/`

**Canonical/frozen policy**: the apps/web copies are canonical; the `src/app` originals stay in place until Stage 5 deletes them. A new **`port-parity.test.ts` drift guard** asserts the copies stay identical (modulo import-extension/pragma adaptations) so a fix landing on only one side fails CI during the dual-copy window (mutation-checked).

**One deliberate freeze deviation**: a real bug fix — `localStorage` access can throw (SecurityError in privacy-restricted/embedded contexts) — was applied **symmetrically to both copies** (with tests on both sides), because the parity guard by design requires fixes to land on both. Behavior only degrades to in-memory "system" default instead of crashing render. `useWhiteboardSync` and all other frozen files are untouched.

No consumers wired yet (BrowserLocalCanvasPage untouched — wiring is slices 9-10).

## Verification
apps/web jsdom 318 ✓ (ported 31+ cases + parity guard collected) / browser 58 ✓ / tsc ✓ / mcp typecheck ✓ / mcp-jsdom 285 ✓ (daemon-side safe-storage tests included). Review: 3 LOW (documented, below fix threshold).

### DIFF
--- a/apps/web/src/hooks/useDirtyState.test.ts
+++ b/apps/web/src/hooks/useDirtyState.test.ts
@@ -1,0 +1,106 @@
+import { act, renderHook } from '@testing-library/react'
+import { afterEach, describe, expect, it } from 'vitest'
+import { useDirtyState } from './useDirtyState.js'
+
+// Thin CustomEvent helpers; jsdom can dispatch these directly.
+function dispatchDocChanged(workspaceId: string, slug: string): void {
+  window.dispatchEvent(new CustomEvent('excalidraw:doc_changed', { detail: { workspaceId, slug } }))
+}
+function dispatchVersionSaved(workspaceId: string, slug: string): void {
+  window.dispatchEvent(
+    new CustomEvent('excalidraw:version_saved', { detail: { workspaceId, slug } }),
+  )
+}
+
+afterEach(() => {
+  // renderHook should unmount listeners for us; no extra cleanup is needed here.
+})
+
+describe('useDirtyState', () => {
+  it('starts clean (dirty=false)', () => {
+    const { result } = renderHook(() => useDirtyState('s1', 'c1'))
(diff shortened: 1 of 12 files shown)


kamiazya:
Gemini triage — both findings are the same class (state adjustment in `useEffect` on prop change → one extra render / momentary flash) and are **deferred with rationale rather than applied here**:

- The pattern is pre-existing in the battle-tested daemon originals; this PR is a behavior-preserving port, and fixing it would restructure hook internals **and** force symmetric edits to the frozen `src/app` copies (the port-parity guard requires both sides to change — as the review itself notes).
- The hooks are not consumed in apps/web yet (wiring is slices 9-10), so the one-frame flash is currently unobservable.

Tracked for the slice 9-10 wiring pass, where the React-recommended derive-during-render / key-based-reset refactor can land with an observable no-flash test. Merging on green checks per the port scope.

PR #140, opened 2026-07-06
feat(mcp-server): expose api-client/api-contracts subpaths, port pure libs to apps/web

kamiazya:
UI-unification Stage 2, slice 0 — the shared foundation every component-port lane (5-8) depends on.

## Part A: new subpath exports

- **`@kamiazya/whiteboard-mcp/api-client`** — exposes `apiFetch` (already relocated to `src/shared` in slice 2), cloned from the `./daemon-backend` export pattern (dev types → source, publish types → dist d.ts; both `exports` and `publishConfig.exports`).
- **`@kamiazya/whiteboard-mcp/api-contracts`** — a **deliberately narrow barrel** re-exporting only `branches.ts` and `canvas.ts` Zod schemas (the two contracts lanes 5-8 consume). A new **barrel-scope guard test** asserts the barrel's specifiers equal exactly those two files, so scope creep into runtime/libraries/palette/etc. fails CI — the boundary test alone can't catch that (it scans per-file browser safety, not barrel contents).
- `publish-contract.test.ts` updated red-first and mutation-checked (reverting the package.json addition fails it). No boundary-allowlist change needed (`api-client.js` exact + `api-contracts/` prefix already allowed). Type probes added in `apps/web/src/_type-probe.ts`.

## Part B: pure lib port (red-first)

`error-copy` / `mini-graph` / `format-bytes` / `app-logger` (+ their 4 test suites, incl. fast-check property tests via a local `test-utils/fast-check.ts` helper — no pinned seed) copied to `apps/web/src/lib/`. apps/web copies are canonical; the `src/app` originals are frozen until Stage 5 deletes them. No consumer wiring beyond the type probe — lanes 5-8 own consumption. `fast-check`/`@fast-check/vitest` added to apps/web devDeps at the same versions mcp-server uses.

## Verification

mcp typecheck ✓ / mcp-node 2684 ✓ (publish-contract + barrel-scope + boundary) / apps/web vitest 345 ✓ / tsc ✓. Review: 0 confirmed findings.

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -37,6 +37,7 @@
     "zod": "catalog:"
   },
   "devDependencies": {
+    "@fast-check/vitest": "^0.4.1",
     "@tailwindcss/vite": "^4",
     "@testing-library/react": "^16.0.0",
     "@types/react": "^19",
@@ -44,6 +45,7 @@
     "@vitejs/plugin-react": "^6.0.3",
     "@vitest/browser-playwright": "catalog:",
     "fake-indexeddb": "^6.2.5",
+    "fast-check": "^4.7.0",
     "playwright": "1.59.1",
     "tailwindcss": "^4",
     "typescript": "catalog:",
--- a/apps/web/src/_type-probe.ts
+++ b/apps/web/src/_type-probe.ts
@@ -61,3 +61,21 @@ import type {
 export declare function _useBackend(b: CanvasBackend): void
 export declare function _useHandlers(h: CanvasBackendHandlers): void
 export declare function _usePayload(p: VersionCreatedPayload): void
+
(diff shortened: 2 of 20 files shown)


kamiazya:
Gemini triage — **applied**: the formatBytes rounding edge (values just under a unit boundary displaying as `1024 B` / `1024.0 KiB`) now bumps to the next unit (`1.0 KiB` / `1.0 MiB`; TiB cap keeps raw rounding). Red-first on the apps/web canonical copy, mirrored byte-for-byte to the frozen src/app original so the copies stay identical. Both sides' suites green (5/5 each), both typechecks green. (CodeRabbit was rate-limited on this PR — recorded as a skip.)

PR #141, opened 2026-07-07
feat(mcp-server): harden local-daemon auth surface

kamiazya:
Implements the daemon-side half of the adopted Stage 2 auth design (ADR-0002 addendum): close the gaps the security spike identified, without touching the Stage 4 scope (Bearer-everywhere, `!token` bypass closure).

## Changes (all red-first)

- **Host loopback guard on HTTP `/api/*`** (new `security/api-host-guard.ts`): WS and `/mcp` already rejected non-loopback Hosts; plain HTTP API was the one unguarded surface. `Host: evil.example` → 403 in local-daemon mode; the middleware takes the server-mode exposure decision as a function-boundary argument (Stage 4 branch point), and is wired **before** the CORS middleware so the OPTIONS short-circuit can't bypass it. Mutation-checked (forcing pass-through fails the 403 test).
- **Bind guard**: `daemon-auth-binding.ts`'s documented-but-unimplemented pre-startup guard now exists — `--host=0.0.0.0` (or any non-loopback bind) refuses to start in local-daemon mode; loopback forms (`127.0.0.1`/`localhost`/`::1`/`[::1]`) pinned by test.
- **`ping` pid → `instanceId`**: the unauthenticated ping no longer publishes the OS pid; a per-start `crypto.randomUUID()` instanceId replaces it. `daemonPingResponseSchema` updated schema-first with all consumers via `z.infer` — including all **three** CLI identity paths (`server stop` / `server status` / `server doctor`), which now match on instanceId. Legacy records without instanceId take the safe path: never kill, report identity-unverifiable explicitly. Mutation-checked. `docs/how-to/self-host-with-docker.md`'s documented ping body updated in the same increment.
- **Loopback logic unified** onto `cors-loopback.ts` canonical helpers — ws-auth / mcp-http / daemon-auth-binding private copies removed, `[::1]` handling pinned, behavior-preserving (no existing test weakened).
- **ws-auth negative test**: non-loopback Origin + valid token → 403 (locks the Origin-before-token ordering).
- **Smoke extended** (`scripts/smoke/mcp-e2e-smoke.mjs`): instanceId-shaped ping parse, Host-spoofed GET → 403, loopback GET → 200.

## Out of scope (Stage 4, per adopted design)
`/api/runtime/*` Bearer carve-out unchanged; `!token → true` bypass closure; pairing flow; apps/web wiring (parallel lanes).

## Verification
typecheck ✓ / mcp-node 2734 ✓ / `pnpm smoke:e2e` (real daemon boot, exit 0) ✓ / review: 1 LOW (helper unit-test coverage, documented). Accepted residual risk per design: reflected-CORS reads from other localhost ports until the Stage 4 allowlist.

## Summary by CodeRabbit

* **New Features**
  * The runtime health check now returns a per-start `instanceId` instead of a process ID, improving how active instances are identified.
  * Server status, stop, and doctor flows now use this new instance identity for more reliable checks.

* **Bug Fixes**
  * Prevents misidentifying unrelated processes that may reuse the same OS process ID.
  * Legacy records without an `instanceId` are now handled as unverifiable instead of being treated as a match.

* **Security**
  * Added stronger host validation to block spoofed non-loopback requests and reduce exposure from unsafe daemon binding.

### DIFF
--- a/docs/how-to/self-host-with-docker.md
+++ b/docs/how-to/self-host-with-docker.md
@@ -88,7 +88,10 @@ volume has the correct ownership if you pre-populate it.
 
 The image configures a Docker healthcheck that polls
 `/api/runtime/ping` every 30 s. This endpoint is public (no auth required)
-and returns `{"ok":true,"pid":<number>}`. It does not include any
+and returns `{"ok":true,"instanceId":"<uuid>"}`. `instanceId` is a random
+identifier generated fresh on every process start (not the OS pid — an OS
+pid can be reused by an unrelated process, which would let a stale
+record misidentify it as the daemon). It does not include any
 configuration, credentials, or filesystem paths.
 
 `whiteboard server doctor --json` is intentionally **not** used as the default
--- a/packages/mcp-server/scripts/smoke/mcp-e2e-smoke.mjs
+++ b/packages/mcp-server/scripts/smoke/mcp-e2e-smoke.mjs
@@ -23,9 +23,34 @@
 
 import { spawn } from 'node:child_process'
 import { mkdtempSync, rmSync, writeFileSync } from 'node:fs'
+import http from 'node:http'
 import { tmpdir } from 'node:os'
 import { join, resolve, dirname } from 'node:path'
 import { fileURLToPath } from 'node:url'
+
+// fetch() treats Host as a forbidden header and silently drops any override,
+// so a genuine Host-header spoof for the DNS-rebinding guard test needs a raw
+// node:http request instead.
+function requestWithHostHeader(url, hostHeader) {
+  return new Promise((resolveReq, reject) => {
+    const target = new URL(url)
+    const req = http.request(
+      {
+        hostname: target.hostname,
+        port: target.port,
+        path: target.pathname,
+        method: 'GET',
+        headers: { Host: hostHeader },
+      },
+      (res) => {
+        res.resume()
+        res.on('end', () => resolveReq({ status: res.statusCode }))
+      },
+    )
+    req.on('error', reject)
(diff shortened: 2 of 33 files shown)


kamiazya:
AI review triage (13 comments):
- **[applied] `[::1]` bind crash** (Gemini) — real bug: the guard accepts `[::1]` but `server.listen` throws EINVAL on the bracketed form. `normalizeBindHost` strips brackets before bind, red-first test.
- **[applied] Host-header smuggling** (CodeRabbit, Major) — `evil.example@localhost` / `localhost/path` etc. normalized down to `localhost` and passed the guard; `normalizeHostHeader` now rejects any Host with credentials/path/query/fragment (red-first, 4 cases).
- **[applied] verifyIdentity ×3 duplication** (Gemini ×7 + CodeRabbit) — centralized as `verifyDaemonIdentity` in `daemon-ping-client.ts`; the three `defaultVerifyIdentity` exports remain as aliases so test/injection seams are unchanged.
- **[applied] Hono idioms** (Gemini ×2) — `c.json()` in api-host-guard; `Context` type for `getRequestHost`.
- **[deferred with rationale] WS origin `originHost === requestHost` strictness** (CodeRabbit, Major) — verified **pre-existing on main** (this PR's unification is behavior-preserving and did not introduce it). Relaxing it to loopback-pair matching is exactly the apps/web→daemon connection path (localhost:5173 → 127.0.0.1) that the slice-11 web-side wiring implements next, where it lands with tests driven by the real client. Added to that slice's requirements.

Full mcp-node 2745 ✓ / typecheck ✓ after fixes.

PR #143, opened 2026-07-07
feat(web): port misc UI components (CanvasThumb/ThemeToggle/ErrorBoundary/HeaderSaveDot)

kamiazya:
UI-unification Stage 2, Lane 8a — the four api-client-free misc components, ported with tests from the frozen daemon app.

## Change
- **CanvasThumb** (thumbnail `` + fallback icon; failed-state resets on workspace/slug change — review fix), **ThemeToggleButton** (controlled, cycles system→light→dark via the #139 `ThemeMode`), **ErrorBoundary** (module-local logging seam instead of app-logger to stay disjoint from #140 while it was open; follow-up wire-up tracked in tmp/issues), **HeaderSaveDot** (dirty/saving indicator + Cmd/Ctrl+S hint) → `apps/web/src/components/`.
- Review fix during the loop: the default ErrorBoundary fallback no longer leaks the error message/stack into the UI.
- `packages/mcp-server` zero diff (frozen); components deliberately NOT added to the port-parity guard (byte-parity is for pure logic; components may adapt to IA).
- No page wiring (slices 9-10). New CanvasThumb/ThemeToggle suites are mutation-checked (ported code can't be red-first by construction); ported ErrorBoundary/HeaderSaveDot suites unweakened.

## Verification
apps/web vitest 348 ✓ / browser 58 ✓ / tsc ✓ / mcp typecheck ✓ / packages zero diff ✓. Review: 3 LOW (documented, below threshold).

### DIFF
--- a/apps/web/src/components/CanvasThumb.test.tsx
+++ b/apps/web/src/components/CanvasThumb.test.tsx
@@ -1,0 +1,41 @@
+import { cleanup, fireEvent, render } from '@testing-library/react'
+import { afterEach, describe, expect, it } from 'vitest'
+import { CanvasThumb } from './CanvasThumb.js'
+
+afterEach(() => cleanup())
+
+describe('CanvasThumb', () => {
+  it('renders an img pointed at the latest-thumbnail route, url-encoding the slug', () => {
+    const { container } = render(<CanvasThumb workspaceId="ws-1" slug="my canvas/x" />)
+    const img = container.querySelector('img') as HTMLImageElement
+    expect(img).not.toBeNull()
+    expect(img.getAttribute('src')).toBe(
+      '/api/workspaces/ws-1/canvases/my%20canvas%2Fx/latest-thumbnail',
+    )
+    expect(img.getAttribute('loading')).toBe('lazy')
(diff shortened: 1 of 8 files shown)


kamiazya:
Gemini triage — all 3 applied (red-first where behavioral):
- **HeaderSaveDot disabled-in-TooltipTrigger** — real UX defect (the 'Saving…' tooltip was unreachable on hover); switched to click-guard + `aria-disabled`, matching this repo's CapabilityTeaser pattern. Old native-disabled assertion updated to the new contract; new test locks click-blocking + tooltip reachability.
- **ErrorBoundary non-Error throws** — `getDerivedStateFromError` now accepts `unknown` and normalizes to a real `Error`, so custom fallbacks typed `{ error: Error }` can't crash on a thrown string (red-first).
- **CanvasThumb effect-reset flash** — reset moved to derive-during-render (React-recommended pattern), eliminating the one-frame stale fallback icon. (Applied here rather than deferred like the #139 hook findings because this code is apps/web-only — no frozen-copy symmetry constraint.)

apps/web 364 ✓ / tsc ✓.

PR #144, opened 2026-07-07
feat(web): port merge UI with a Zod-typed merge_committed event contract

kamiazya:
UI-unification Stage 2, Lane 5 — the merge surface, ported from the frozen daemon app with the event contract promoted to Zod. Prerequisite for Lane 6 (branch UI imports MergeDialog).

## Change

- **`merge-committed-event.ts`** (new): `mergeCommittedDetailSchema` (plain non-strict `z.object` so extra fields from the frozen dispatcher never reject), `dispatchMergeCommitted` (compile-checks the payload against `z.infer`), and `parseMergeCommittedEvent` (safeParse; invalid detail → ignored + one app-logger warn). This kills the three hand-written per-subscriber detail interfaces that were drifting independently.
- **MergeDialog / MergeToast / MergeHighlight** → `apps/web/src/components/`, consuming `BranchMeta`/`MergeResponse`/`listVersionsResponseSchema` from the #140 `api-contracts` subpath and `apiFetch` from `api-client`. MergeHighlight validates-then-commits: a malformed detail leaves existing highlight boxes untouched.
- **Config aliases**: vite/vitest configs map the two #140 subpaths to their sources (mirroring the existing browser-shared alias pattern) so dev/vitest resolve without a prior `pnpm build`.
- Tests: contract unit suite (red-first), ported jsdom suites, and a browser-layer `merge-ui.browser.test.tsx` covering real Radix Dialog behavior + the 3-component dispatch→toast/highlight wiring.

`packages/mcp-server` zero diff (frozen; its copies keep hand-written types until Stage 5 deletes them). No page wiring (slices 9-10). apps/web copies are canonical.

## Verification

apps/web vitest 410 ✓ / browser 61 ✓ / tsc ✓ / packages zero diff ✓. Review: 0 confirmed findings (verify lane partially degraded by a credit outage — flagging for extra reviewer attention here).

PR #145, opened 2026-07-07
feat(web): port useBranches (callback-based) and VersionTimeline

kamiazya:
UI-unification Stage 2, Lane 7 — the version-history surface, ported from the frozen daemon app with the sync-event coupling moved to the callback model per ADR-0004.

## Change

- **`useBranches`** → `apps/web/src/hooks/`: same return contract as the original (state/loading/error/refetch + create/delete/rename/setHead/merge/stats via Zod-validated `api-contracts`), with two deliberate behavior changes: (1) the `excalidraw:head_changed` **window listener is removed** — `refetch` is the notification API the unified CanvasPage will call from `useCanvasSync`'s `onHeadChanged` callback (binds lifecycle to backend identity, killing the stale-listener class); (2) an unmount cleanup bumps the existing fetch-seq guard so post-unmount resolutions are discarded (the original had no mounted guard).
- **`VersionTimeline`** → `apps/web/src/components/`: 15s polling, head filtering, restore AlertDialog, mini-graph — plus dev-loop review hardening (restore-failure keeps the dialog open and undo state; network failures don't wedge refresh; branches response validated through `canvasBranchesStateSchema`).
- **Race fix found by the review gate (HIGH, red-first)**: `useBranches` defaults `head` to `'main'` until `/branches` resolves — the timeline no longer renders (or offers restore on) versions filtered by that default; it holds on Loading… until the real HEAD arrives.
- Config aliases for the two #140 subpaths were shared with Lane 5 and already landed in #144 — this PR's configs are byte-identical to main.

`packages/mcp-server` zero diff (frozen). No page wiring (slices 9-10); apps/web copies are canonical.

## Verification

apps/web jsdom 461 ✓ / browser 65 ✓ (one first-run flake retried clean — known Vite first-run dep-optimisation jitter) / tsc ✓ / packages zero diff ✓. Ported suites unweakened; the head_changed window-event test is replaced by callback-model tests (refetch triggers exactly one list fetch; post-unmount refetch is inert; refetch after prop change hits the new canvas URL).

### DIFF
--- a/apps/web/src/components/VersionTimeline.browser.test.tsx
+++ b/apps/web/src/components/VersionTimeline.browser.test.tsx
@@ -1,6 +1,6 @@
+import { cleanup, render, screen, waitFor } from '@testing-library/react'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { page } from 'vitest/browser'
-import { cleanup, render, screen, waitFor } from '@testing-library/react'
 import '../index.css'
 import VersionTimeline from './VersionTimeline.js'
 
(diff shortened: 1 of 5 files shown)


kamiazya:
Gemini triage (5 findings):
- **[applied] restore state survives canvas switch** — real bug (staged restore from the previous canvas would POST its version id to the NEW canvas). The canvas-change effect now also clears pendingRestore/restoreError/isRestoring. Red-first.
- **[applied] useBranches stale state on prop change** — the hook now resets to defaults synchronously during render (not an effect, so there is no one-frame window where loading-ignoring consumers see the previous canvas). Red-first.
- **[applied] clock-skew negative timestamps** — clamped to `0s ago`; mutation-checked.
- **[declined] `gap-6` spacing** — verified identical in the frozen daemon original; this is a faithful port and visual polish belongs to the unified-page slice, not the port. If it looks wrong there, it looks wrong in the daemon UI too — noting for the slice 9-10 pass.
- **[declined] `deleteBranch` URL-builder rename** — matches the frozen original's export name; keeping gratuitous divergence out of the dual-copy window makes the port reviewable against its source. Rename candidate for Stage 5 cleanup when the original is deleted.

apps/web 464 ✓ / tsc ✓.

PR #146, opened 2026-07-07
feat(web): port StorageReportCard with Zod-validated storage responses

kamiazya:
UI-unification Stage 2, Lane 8b — the last misc component, split from Lane 8a (#143) because it needed the #140 `api-client` subpath.

## Change

- **StorageReportCard** → `apps/web/src/components/` (all 6 original test cases ported unweakened: per-category rows, optimize-all aggregation, dangling-file purge aggregation, two unmount-timer safety cases, humanised "Updated …" ticking).
- **One deliberate freeze deviation, per the Zod discipline**: the original component *cast* the optimize-all / prune-sandwiched / purge responses. Instead of porting the casts, this adds three small response schemas (`optimizeAllCanvasesResponseSchema`, `pruneSandwichedVersionsResponseSchema`, `purgeResultSchema`) to `src/shared/api-contracts/canvas.ts` — the shared contracts layer, NOT the frozen `src/app` tree — with contract tests, and the component hydrates through `safeParse`. Server shapes verified against the actual route handlers (`canvas.ts` optimize-all/prune-sandwiched, `files.ts` purge-dangling); schemas are non-strict so the extra `results` detail passes through unvalidated by design (commented).
- The `/api/runtime/*` Bearer carve-out (#141) is a wiring-slice concern — flagged for slices 9-10, the component itself relies on `apiFetch` attaching the token.

No page wiring; apps/web copy is canonical; `src/app` untouched.

## Verification

mcp typecheck ✓ / mcp-node 2754 ✓ (new contract tests included) / apps/web jsdom 425 ✓ / tsc ✓. Review lane was partially degraded by a credit outage — schema-vs-route shapes were re-verified manually by the integrator (noted above).

### DIFF
--- a/apps/web/src/components/StorageReportCard.test.tsx
+++ b/apps/web/src/components/StorageReportCard.test.tsx
@@ -1,0 +1,285 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { act, cleanup, fireEvent, render, waitFor } from '@testing-library/react'
+import { STATUS_CLEAR_MS, StorageReportCard } from './StorageReportCard.js'
+
+const PAYLOAD = {
+  totalBytes: 4096,
+  fileCount: 6,
+  byCategory: {
+    blobs: { bytes: 1024, files: 2 },
+    versions: { bytes: 2048, files: 1 },
+    files: { bytes: 0, files: 0 },
+    libraries: { bytes: 0, files: 0 },
+    db: { bytes: 1024, files: 1 },
+    other: { bytes: 0, files: 2 },
+  },
+}
+
+function jsonResponse(body: unknown): Response {
+  return new Response(JSON.stringify(body), { status: 200 })
+}
+
+beforeEach(() => {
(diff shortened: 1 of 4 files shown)


kamiazya:
Gemini triage — all 3 applied (red-first):
- **removeLib unhandled rejection** — DELETE now wrapped in try/catch routing thrown fetches into `libsError` (callers fire-and-forget, so a network failure previously escaped as an unhandled rejection). Regression test with a rejecting DELETE.
- **misleading empty state** — a failed list fetch no longer renders 'No user libraries installed.'; the error line shows instead. Regression test with a 500 list response.
- **schema strictness** — the three new response schemas now use `z.number().int().nonnegative()` for counts/bytes, with negative/fractional rejection tests; verified against the actual route outputs (fs byte sizes and deletion counts are integer accumulations).

contract 61 ✓ / component 17 ✓ / apps/web 427 ✓ / tsc ✓ / mcp typecheck ✓.

PR #148, opened 2026-07-07
feat(mcp-server): move daemonToken out of runtime config into a one-shot TokenStore

kamiazya:
Implements the ADR-0002 addendum's token-delivery decision (adopted Stage 2 auth design): the daemon token no longer travels inside `__WHITEBOARD_RUNTIME_CONFIG__`.

## Change (red-first, three sides changed together)

- **Injection**: daemon self-serve HTML (`server/app.ts`) and the vite dev plugin inject the token as a dedicated `window.__WHITEBOARD_DAEMON_TOKEN__` script (omitted when no token is configured); the runtime config script is token-free.
- **Consumption**: new `src/shared/token-store.ts` — module-singleton, `readDaemonTokenOnce()` reads the global once, validates with `z.string()`, caches in memory, **deletes the global**. `apiFetch` and `DaemonBackend.openSocket` both take the token from the store. Node-safe / browser-safe (boundary-allowlisted, plus a new transitive-safety scan so allowlisted shared modules can't smuggle Node builtins through indirect imports).
- **Schema hardening**: the shared `runtimeConfigSchema` becomes the strict single source of truth (`{ daemonBaseUrl? }.strict()`) — a payload still carrying `daemonToken` fails parse and falls back to defaults (regression-locked + mutation-checked); this also closed a pre-existing drift where the served config's `daemonBaseUrl` wasn't in the schema. apps/web's schema already rejected the key; now locked by test.
- **Redaction test**: a fake logger sink asserts a sentinel token never appears in any log record across store read, Bearer attachment, and WS auth-failure paths.
- **Smoke**: `smoke:e2e` now fetches the served HTML and asserts config-token absence + dedicated-global presence, alongside the existing authenticated /api + WS steps (exit 0 with a real daemon).

**Framing (per ADR-0002 addendum): this is serialization-surface reduction, not a security boundary** — it keeps the token out of the runtime-config object that logging/error-reporting paths may serialize; it does nothing against a script that runs before the delete.

Frozen-side note: `src/app` production code is untouched; three of its test files changed only their token-seeding mechanism (config-key → dedicated global + `resetTokenStoreForTests`), with auth-failure coverage unweakened.

## Verification
mcp-node 2771 ✓ / mcp-jsdom 286 ✓ / `pnpm smoke:e2e` (real daemon) exit 0 ✓ / apps/web 497 ✓ / both typechecks ✓. Review: 1 LOW (malformed daemonBaseUrl acceptance — noted).

### DIFF
--- a/apps/web/src/runtime-config.test.ts
+++ b/apps/web/src/runtime-config.test.ts
@@ -58,6 +58,22 @@ describe('runtimeConfigSchema', () => {
     expect(() => resolveRuntimeConfig({ Authorization: 'Bearer tok' })).toThrow()
   })
 
+  // Mutation-check target: the daemon auth token travels via its own global
+  // (window.__WHITEBOARD_DAEMON_TOKEN__), never inside this config object.
+  // Reverting that split — putting daemonToken back into the injected config —
+  // must make this assertion fail.
+  it('rejects a payload containing daemonToken (.strict() rejection of the token channel)', () => {
+    expect(runtimeConfigSchema.safeParse({ daemonToken: 'secret' }).success).toBe(false)
+    expect(() =>
+      resolveRuntimeConfig({ daemonBaseUrl: 'http://127.0.0.1:3099', daemonToken: 'secret' }),
+    ).toThrow()
+  })
+
+  it('parses the split shape: a token-free config with daemonBaseUrl only', () => {
+    const config = resolveRuntimeConfig({ daemonBaseUrl: 'http://127.0.0.1:3099' })
+    expect(config).toEqual({ daemonBaseUrl: 'http://127.0.0.1:3099' })
+  })
+
   it('explicit default port 443 is rejected (URL.origin normalizes it away)', () => {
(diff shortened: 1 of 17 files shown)


kamiazya:
Triage:
- **CI verify failure** — root-caused: the verify job runs `smoke:e2e` **before** `pnpm build`, so `dist/app` (and therefore the HTML-injection path) doesn't exist in CI; locally it passed because dist was present. The token-global assertion is now gated on the served HTML actually containing the runtime-config script (skip + log otherwise); the stricter 'no daemonToken anywhere' check remains unconditional.
- **Gemini: strict-mode `delete` TypeError on a non-configurable global** — **applied**: wrapped in try/catch with the rationale inline (value already cached; the scrub is surface reduction, not a boundary). Red-first test with `Object.defineProperty(configurable: false)`, mutation-checked.

token-store 8 ✓ / typecheck ✓ / local smoke (dist present → assert path) ✓.

PR #174, opened 2026-07-11
fix(web): apply theme tokens to page root so dark mode text is readable

kamiazya:
## Summary

Design-audit CRITICAL: in dark mode, apps/web page chrome (canvas title, header, capability-teaser buttons) computed to `color: rgb(0,0,0)` on a near-black background — unreadable.

Root cause: `index.css` defines the light/dark token values and `useThemeMode` toggles the `dark` class correctly, but the shadcn-standard base layer that applies the tokens to the page root was missing — untokened text inherited UA-default black.

Fix: `@layer base { body { background-color: var(--background); color: var(--foreground); } }` (+ regression coverage).

## Test plan
- [x] Red-first web-browser computed-style test (real index.css + html.dark): body + mounted WorkspaceTopBar text resolve to a light color, not rgb(0,0,0); light mode asserted too (guards inversion)
- [x] Mutation check: reverting the base-layer rule sends both assertions red
- [x] Existing useThemeMode / WorkspaceTopBar suites green; typecheck green
- Visual verification screenshot to follow in a PR comment

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -70,3 +70,16 @@
   --input: oklch(1 0 0 / 15%);
   --ring: oklch(0.556 0 0);
 }
+
+/*
+ * Without this, page chrome text (canvas title, header, teaser buttons)
+ * falls back to the UA-default black regardless of --foreground, since
+ * nothing else applies the design tokens to the actual page root. In dark
+ * mode that reads as unreadable black-on-black.
+ */
+@layer base {
+  body {
+    background-color: var(--background);
+    color: var(--foreground);
+  }
+}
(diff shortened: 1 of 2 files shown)


kamiazya:
## Visual repro (verified on the PR branch, real browser)

Before (design audit, dark mode — chrome text `oklch(0.145 0 0)` on the same-value background, contrast **1:1**, effectively invisible):

before-dark

After (dark mode — title/buttons compute to `oklch(0.985 0 0)` on `oklch(0.145 0 0)`, contrast ≈ **18.9:1**, WCAG AAA):

after-dark

Light mode after the fix (no regression):

after-light

PR #204, opened 2026-07-12
feat(web): make canvases addressable — history routing, deep links, and Pages SPA fallback

kamiazya:
## Summary

Closes the other half of the user's dogfood report (「URLにルーティングされない」). apps/web was state-driven with a single URL: reload lost your view, no per-canvas bookmarking or sharing, and back/forward did nothing.

- **History routing** via react-router: `/canvas/:workspaceId/:slug` (daemon canvas), `/w/:workspaceId` (gallery), `/local/:canvasId` (browser-local canvases are addressable too). The R2 `DaemonView` state in App.tsx now syncs both ways — navigate updates the URL, back/forward updates the view, a cold-load deep link restores it.
- **#wb= pairing still works and now self-cleans**: the fragment is consumed once, then the URL is replaced with the canonical route, so a reload doesn't re-consume a stale fragment and the user is left with a shareable URL.
- **Cloudflare Pages SPA fallback** (`public/_redirects`): without it, a direct load or reload of any client-side route 404s in production — the classic way routing ships broken.

This branch also carries #201 (persist the daemon connection target + the working "Open the local app" banner CTA), including its review fixes — that PR's deep link `/canvas/:ws/:slug` now actually resolves to a route instead of silently landing on the default view. **Merging this supersedes #201.**

Critical-path budget raised 108 → 126 KB: react-router must live in the entry (it decides which page to lazy-load), so its ~16 KB is an unavoidable cost of addressable URLs. Measured 114.0 KB — still far under the 300 KB app-page budget.

## Test plan

- [x] apps/web 877/877; `pnpm -r typecheck` clean; `pnpm --filter @kamiazya/whiteboard-web build` green
- [x] Bundle gate passes at the new budget (114.0 KB measured)
- [x] Routes covered: cold-load deep link → correct view; navigate → URL updates; back/forward; #wb= consumed then replaced; unknown route falls back sensibly

## Summary by CodeRabbit

* **New Features**
  * Added client-side routing for workspace, canvas, and local canvas pages.
  * Deep links and browser refreshes now load the correct page instead of returning an error.
  * Canvas selections update the browser URL and can be restored from bookmarked links.
  * Added an “Open the local app” action with direct links to the last connected location.
  * Added an option to forget saved local-app connection details.

* **Bug Fixes**
  * Improved handling of stale or invalid canvas links by falling back to the default canvas.
  * Preserved navigation history across back and forward browser actions.


kamiazya:
Triage of the review comments (all four looked at against the actual code):

**Applied** (fe2072f):
- `decodeURIComponent` URIError in `parseDaemonRoute` / `parseBrowserLocalRoute` — real crash. Both parsers run inside render-phase lazy initializers, so `/canvas/w%1/main` took the whole app down instead of falling back to the index. Undecodable segments are now treated as not-a-route, with a regression test.
- Redundant `setDaemonView` on identical routes — real. The parser returns a fresh object and React compares by reference, so back/forward onto the current view re-rendered for nothing. Now guarded with a functional updater that compares the canonical path.

**Not applied** — adding `location.pathname` to the state→URL effect's deps:
this would break browser-local routing. `App` has no `<Route>` tree; both sync
effects run in *every* mode, including browser-local. `daemonRoutePath` of a
browser-local session's `daemonView` is `/`, so the moment
`BrowserLocalCanvasPage` navigates to `/local/:canvasId`, a pathname-triggered
refire of this effect would see `'/' !== '/local/c1'` and navigate straight back
to `/`, undoing every browser-local canvas URL. There is also no stale closure
in the current form: the effect re-reads `location.pathname` from the render in
which `daemonView` changed, which is exactly the read it needs. Leaving the
deps as-is, with the existing comment explaining why.

kamiazya:
@coderabbitai review

kamiazya:
@coderabbitai review

kamiazya:
@coderabbitai review

kamiazya:
@coderabbitai review

kamiazya:
@coderabbitai review

kamiazya:
@coderabbitai review

PR #205, opened 2026-07-12
fix(mcp-server): document every MCP tool input field and honor annotate_batch per-item groupAs

kamiazya:
## Problem

Every MCP tool declared its input contract twice: a Zod shape (what the SDK actually validates against and what `tools/list` derives its JSON Schema from) and a hand-written JSON-Schema literal in the same file. The literal carried the field descriptions; the Zod shape carried none. Result: **an MCP client — including the LLM calling the tool — saw a schema with no documentation at all**, while the descriptions sat in a literal nothing read.

The drift wasn't only cosmetic. `annotate_batch`'s batch-level `groupAs` description promised that `annotations[].groupAs` overrides it, but the item schema never declared the field and the runtime only ever read the batch-level value. Because the SDK builds a non-strict `z.object()` from the raw shape, a per-item `groupAs` was **silently stripped**, not rejected — the documented feature simply did nothing.

## Changes

- **Descriptions onto the Zod shapes, literals deleted** across all `tools/*.ts` (`.describe()` count 142 → 233). `inputSchema` is now `z.toJSONSchema(z.object(shape))` everywhere, matching the precedent already set by `annotate.ts`.
- **`annotate_batch` per-item `groupAs` implemented.** It applies *in addition to* the batch group (an element can be in both), which is what carving a sub-group out of a larger batch actually needs — and the description now states that shape instead of the override it never had. Red-first test asserts the batch group holds every element while the sub-group holds only its own.
- **`canvas_export_json` handler was dropping `outputPath` / `overwrite`** before forwarding to `execute()`; both are now passed through.
- **Registry-wide guard test**: walks each registered inputSchema's derived JSON Schema (nested objects and array items included) and fails on any field without a non-empty description, plus a count assertion so a newly registered tool can't be silently skipped. Bare `x`/`y` leaves are exempt — the coordinate pair is described at the parent, per existing convention.

## Verification

- `pnpm test --project mcp-node` — 197 files, 3071 passed
- `pnpm -r typecheck` — clean
- `pnpm build` — green
- `pnpm smoke:e2e` — ALL OK (real MCP client over the daemon; the SDK validates `structuredContent` against `outputSchema` at runtime)
- New guard test: 49 tools, every input field described


kamiazya:
@coderabbitai review

kamiazya:
@coderabbitai review

kamiazya:
@coderabbitai review

kamiazya:
@coderabbitai review

kamiazya:
@coderabbitai review

PR #206, opened 2026-07-12
refactor(mcp-server): split the canvas route file into cohesive sub-routers

kamiazya:
## Problem

`packages/mcp-server/src/server/routes/canvas.ts` was one 855-line file handling CRUD, versioning, auto-save debouncing, name/pin persistence, branch/restore, export, thumbnails, and maintenance. Nothing in it could be read — or tested — in isolation from everything else.

## Changes

Pure structural split into `routes/canvas/`, with `canvas.ts` reduced to a 68-line composer that wires the shared deps (`versionStore`, `triggerAutoVersion`, auto-compact registration) and mounts each sub-router:

| module | lines | concern |
| --- | --- | --- |
| `shared.ts` | 42 | broadcast indirection, display-name default, corrupt-data handling |
| `auto-version.ts` | 56 | `createAutoVersionTrigger` + the debounce interval |
| `workspaces.ts` | 101 | list workspaces, list/create canvases |
| `metadata.ts` | 117 | workspace/canvas names, pin |
| `live-doc.ts` | 78 | snapshot + update (auto-version trigger + broadcast) |
| `versions.ts` | 116 | list/save manual versions |
| `maintenance.ts` | 114 | compact, prune-sandwiched, optimize-all |
| `export-json.ts` | 67 | export-json |
| `thumbnails.ts` | 110 | thumbnail PUT/GET, latest |
| `restore.ts` | 141 | in-place reconcile + restore-as-new-canvas |

Every route path, status code, and response body is unchanged. `setBroadcastFn` and `AUTO_VERSION_INTERVAL_MS` stay importable from `./canvas.js` so `ws.ts` and the existing tests keep their import paths.

## Verification

This is behavior-preserving, so the proof is that **no test file changed** — `routes/canvas.test.ts` (1277 lines, exercising every route end to end plus `createAutoVersionTrigger` directly) passed unchanged throughout.

- `pnpm test --project mcp-node` — 197 files, 3021 passed / 2 skipped
- `pnpm vitest run --project mcp-node packages/mcp-server/src/server/routes/` — 18 files, 259 passed
- `pnpm --filter @kamiazya/whiteboard-mcp typecheck` — clean
- `pnpm lint:noconsole` — clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added workspace and canvas management, including creation, listing, naming, and pinning.
  * Added live document snapshots and updates with real-time synchronization.
[...]

PR #208, opened 2026-07-12
fix(mcp-server): make restore overwrite reconcile onto the target instead of replacing it

kamiazya:
> Stacked on #206 (base is `canvas-route-split`). GitHub will retarget this to `main` once #206 merges.

## Problem

`restore` had two branches with different semantics, and one of them was unsound.

The **in-place** restore (no `targetSlug`) reconciles the past version onto the *live* Loro doc, commits, saves, and broadcasts the resulting CRDT delta — history stays continuous and connected clients converge.

The **restore-as-new-canvas** branch (`targetSlug` + `overwrite: true`) instead *replaced persistence wholesale* with a brand-new document and merely evicted the server cache. Whenever the target was open in someone's browser — including `targetSlug === slug`, restoring a canvas onto itself, which silently bypassed the in-place path entirely — that client kept the **old** Loro doc in memory. The replacement is a different CRDT (different peer/version vectors), so the client's next edit arrives based on the old history, gets merged server-side, and **resurrects the very elements the restore removed**. Broadcasting a delta can't fix that: a delta from doc A is meaningless against doc B.

## Change

`overwrite` now means *make the target canvas's content match this version*, not *replace the file on disk*. Reconcile/merge is the one restore semantic:

- `reconcileCommitSaveBroadcast(...)` is extracted as the single reconcile → commit → save → broadcast → `sendRestoreEvent` implementation, used by both paths.
- Target does not exist → created, exactly as before.
- Target exists **without** `overwrite` → 409 `output_exists`, exactly as before.
- Target exists **with** `overwrite` → its live doc is reconciled onto (one CRDT lineage), and the update is broadcast to *that* canvas. `targetSlug === slug` now collapses into the in-place case instead of being a hole.
- The failure-path cache eviction added in #206 is preserved.

Tool descriptions and the request-schema doc comment were corrected — they described the old replace semantics. The wire contract is unchanged.

## Verification (red first)

Five new `mcp-node` tests, each confirmed failing against the old code first:

1. `targetSlug === slug` + `overwrite` reconciles in place — the cached doc is the *same object* afterwards (no lineage swap) and one non-empty update is broadcast.
2. Restoring into a different existing canvas reconciles *that* canvas's live doc, broadcasts to it (not the source), and its content matches the restored version; the source is untouched.
3. **The resurrection bug itself**: an independent `LoroDoc` imported from the target's pre-restore snapshot (modelling a real connected browser client) sends its stale-history update *after* the restore — the tombstoned element stays dead. This is what the old file-replace path allowed.
4. New target still creates the canvas; existing target without `overwrite` still returns 409.

`pnpm test --project mcp-node` 3028 passed / 2 skipped · `pnpm -r typecheck` clean · `pnpm build` green.

## Known follow-up

`versionRestoreOutputSchema.restoredAs` is still `'in-place' | 'new-canvas'`, so a reconcile onto an *existing* target reports `'new-canvas'` even though nothing new was created. No consumer distinguishes them today, so the wire contract was left alone; a third value (`'target-reconciled'`) is worth adding if that distinction ever becomes user-visible.


kamiazya:
Superseded: rebased onto main (the stacked base #206 has merged) and re-opened so the diff is against main. No content change.

PR #211, opened 2026-07-12
fix(mcp-server): redact secrets in log output

kamiazya:
The logger is already pino (a root instance fanned out to stderr, MCP `notifications/message`, and the test capture sink), but it had no `redact` config. So a single careless `log.error({ client, err }, …)` — logging a `DaemonClient`, a request, or a config object wholesale — would have put the local-daemon bearer token straight into stderr *and* into every connected MCP client's log stream.

## What is redacted

Derived from the actual call sites and the objects that cross a process boundary in this repo, not from a generic list:

- `token` / `daemonToken` / `bootstrapToken` — the local-daemon bearer token (`shared/token-store.ts`, `mcp/tools/pairing-link.ts`, the `#wb=` pairing payload). `DaemonClient` carries `token` as a plain field, which is exactly the shape a future wholesale log would hit.
- `accessToken` — OAuth (`security/oauth-resource-strategy.ts`).
- `authorization` — read at every auth-check site; never logged today.
- `cookie`, `password`, `secret`, `apiKey` — standard nets for future call sites.

Each is matched at the top level and one level of `*.<name>` nesting. `censor: '[redacted]'` rather than `remove: true`, deliberately: a visible placeholder tells the reader a credential *was* there, while a silently missing key is easy to not notice.

**No live leak was found.** Every existing `getLogger(...)` call site was audited and none logs a credential today — the codebase already shows deliberate discipline here. This is defense in depth, not a fix to an active bug, so nothing needs rotating.

## Tests (red first)

Four new tests in `log.test.ts`, each confirmed failing before the config landed:
- every secret field is redacted **on both** the stderr destination and an MCP-notification-style destination (redaction happens before the fanout, not per-sink);
- a one-level-nested secret (`{ client: { token } }`) is redacted while its siblings survive;
- a token attached as a custom property on a logged `Error` — which bypasses the err serializer — is still redacted (pino runs serializers *before* redact, so `err.token` lands one level under `err` and `*.token` catches it);
- ordinary diagnostic fields are untouched, so redaction cannot silently eat normal telemetry.

## What redaction cannot do (documented in AGENTS.md, not papered over)

- `fast-redact` has no arbitrary-depth wildcard: a secret nested two or more levels under an unlisted key is not caught.
- Redaction only sees **structured fields**. A secret interpolated into the message string — ``log.info(`token=${token}`)`` — leaks verbatim, and there is no technical guard against it. Call sites must not do that.

`pnpm test --project mcp-node` 3078 passed · `pnpm -r typecheck` clean · `pnpm build` green.

## Summary by CodeRabbit

* **New Features**
  * Added automatic redaction of sensitive token, secret, and credential fields in server logs.
  * Redaction applies to supported top-level and one-level nested fields, including serialized errors.
  * Redacted values appear as `[redacted]` in captured logs and raw output.

* **Documentation**
  * Documented supported redaction coverage and known limitations.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -193,6 +193,28 @@ Why this exists:
 Tests use `resetLoggerForTests({ level, sink })` to install a fake sink and
 assert against structured records instead of spying on `console`.
 
+### Redaction
+
+The root pino instance in `log.ts` redacts a fixed list of field names —
+top-level (`token`, `daemonToken`, `bootstrapToken`, `accessToken`,
+`authorization`, `cookie`, `password`, `secret`, `apiKey`) and one level of
+nesting under any key (`*.token`, `*.daemonToken`, …) — replacing the value
+with `[redacted]` before the record reaches stderr, an MCP
+`notifications/message` subscriber, or a test capture sink. This is what
+stops a call site that carelessly logs a whole request/client/config object
+(e.g. `log.error({ client }, 'request failed')`) from leaking the daemon
+bearer token or an OAuth access token.
+
+Adding a new secret-bearing field anywhere in the server means adding both
+its top-level and its `*.<name>` path to `REDACTED_PATHS` in `log.ts` — pino
+redaction does not infer field names, and `fast-redact` has no
+arbitrary-depth wildcard, so a secret nested two or more levels deep under
+an unlisted key is not caught. The safer habit is still to never log a
+secret-bearing object wholesale in the first place; redaction is the net,
+not the plan. Redaction also cannot help when a secret is interpolated
+directly into a message *string* (e.g. `` log.info(`token=${token}`) ``)
+rather than passed as a structured field — do not do that.
+
 ## Doc Screenshots
 
 Images under `docs/assets/` that are produced from the running UI (canvas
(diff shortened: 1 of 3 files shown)


kamiazya:
Review triage (both findings checked against the actual code):

**Applied** — the three vacuous test assertions. `writeSpy!.mock.calls.at(-1)` is `undefined` when the spy was never called, and `String(undefined)` doesn't contain the secret, so the assertion passed while proving nothing: a dropped log would have kept these tests green. They now assert the call happened and that the expected message is present before checking the secret is absent. Mutation-checked with `redact: { paths: [] }` — all three now fail (`Expected: "[redacted]", Received: "daemon-secret-token"`), so they are load-bearing.

**Not applied** — adding a `*.*.<name>` tier. Every `log.*` call site in `src/server/**` (10 outside tests) passes either a flat field bag or `{ err }` / `{ error }` / `{ cause }`, and pino's `stdSerializers.err` output is flat as well — so the deepest position a credential can actually occupy today is one level (`err.token`, `client.token`), which `*.<name>` already covers. `fast-redact` has no arbitrary-depth wildcard, so each extra tier is a hand-written path list paid for on every log call. Adding one for a shape nothing currently logs is speculative. A comment in `log.ts` now records this as a deliberate stop and names the trigger for revisiting it: the first call site that logs a two-level-nested credential (e.g. a whole `req` with `headers.authorization`). AGENTS.md already documents the one-level limit accurately.

PR #214, opened 2026-07-12
chore: upgrade to pnpm 11 and enable the global virtual store

kamiazya:
## Summary

Upgrades pnpm 10.29.3 → 11.12.0 across `packageManager`, `Dockerfile.server`'s corepack pin, and migrates `package.json`'s `pnpm.overrides` into `pnpm-workspace.yaml` (pnpm 11 no longer reads settings from package.json's `pnpm` field).

**`enableGlobalVirtualStore` is added to `pnpm-workspace.yaml` but shipped OFF.** This is the actual ask of this task (shared virtual store across concurrent agent git worktrees, avoiding the ~1GB-per-worktree `node_modules/.pnpm` copy that has caused real disk exhaustion) — but enabling it breaks `apps/web` typecheck with a reproducible TypeScript structural-typing regression. Full writeup and blocker below.

### Measured with the flag ON (before the blocker was found)

- Fresh `node_modules` after a clean install: **1.5 MB** (all symlinks) vs the pnpm-10 baseline of **~1.1 GB**.
- Clean reinstall after `rm -rf node_modules`: **<1s** (vs ~7-13s for a normal populated install).
- `node_modules/@excalidraw/excalidraw` (and every other package checked) resolves to a single shared hash under `~/Library/pnpm/store/v11/links/...` — confirmed no duplication.

This is a real, working feature — it just isn't safe to ship yet.

### The blocker (`enableGlobalVirtualStore: true`)

`pnpm -r typecheck` fails on `apps/web` only when the flag is on:

```
src/hooks/useCanvasSync.ts(604,13): error TS2740: Type 'MutableAppState' is missing the
following properties from type 'Pick<AppState, keyof AppState>': name, scrollX, scrollY,
zoom, and 85 more.
```

Reproduced in isolation (minimal repro file, not just the real call site). Ruled out:
- **Duplicate package instances** — `pnpm why @excalidraw/excalidraw`, `@excalidraw/utils`, `@types/react`, and `react` each show exactly one physical copy; the global store's own hash-keyed link directories confirm a single hash per package.
- **Byte-identical files** — checksummed the entire `@excalidraw/excalidraw/types` directory between the flag-on and flag-off installs: identical.
- **TypeScript version** — reproduces identically on TypeScript 6.0.3 (the repo's pinned catalog version) and 7.0.2 (current latest).
- **Confirmed root trigger** — switching `apps/web/tsconfig.json`'s `moduleResolution` from `bundler` to `node16` makes the error disappear (replaced by unrelated pre-existing issues from that broader change, so not adopted). This isolates the cause to TypeScript's `bundler` moduleResolution's wildcard `package.json#exports` handling (`"./*": {"types": "./dist/types/excalidraw/*.d.ts"}`) interacting badly with the deep, external symlink chain the global virtual store introduces (`apps/web/node_modules/@excalidraw/excalidraw` → 11 `../` segments → `~/Library/pnpm/store/v11/links/...`), rather than a dependency-hoisting bug pnpm itself could fix.

Per this repo's own instruction ("if any gate regresses in a way you cannot fix cleanly, stop and report rather than papering over it"), I'm shipping with the flag documented but off, rather than forcing a broad `moduleResolution` change or a type-cast workaround in production code. `pnpm-workspace.yaml` has the full writeup inline so the next person doesn't have to re-derive it.

### Two smaller pnpm-11 behavior changes found and fixed along the way

- `allowBuilds` (pnpm 11's replacement for the interactive `pnpm approve-builds` prompt) needed to be populated for `@swc/core`, `canvas`, `lefthook`, `libxmljs2`, `msw`, `sharp`, `workerd` — **all set to `false`**, because a clean install under the *unmodified* pnpm 10.29.3 baseline shows the exact same "Ignored build scripts" warning for these packages. Only `esbuild` was ever load-bearing here, so this preserves existing behavior rather than newly enabling native builds pnpm was already skipping.
- `pnpm add` (not `pnpm install`) now exits non-zero (`ERR_PNPM_IGNORED_BUILDS`) instead of just warning when an added package has an ignored build script. This broke `smoke:tarball` (the packed-tarball install, which pulls in `protobufjs` transitively) — fixed by adding `--ignore-scripts` to that scratch install, since it only needs the entrypoint to run, not any dependency's native build step.

### Lockfile

`pnpm-lock.yaml` is **unchanged** (`git diff --stat pnpm-lock.yaml` is empty) — `lockfileVersion` stayed `9.0` and content is byte-identical. No format bump, so the currently-open `web-noconsole` PR does **not** need to rebase because of this change.

### Migration note for existing worktrees

Existing sibling worktrees (installed under pnpm 10) are unaffected until they sync with `main`. Volta reads `packageManager` from `package.json` and auto-fetches pnpm 11 transparently the next time `pnpm install` runs in a worktree that has this commit — no manual pnpm version management needed, and (per the lockfile point above) no reinstall-format surprises.

## Test plan

- [x] `pnpm install` (clean, `CI=true`) — no ignored-build warnings, no pnpm-field warning
- [x] `pnpm -r typecheck` — green (with `enableGlobalVirtualStore: false`)
- [x] `pnpm build` — green
- [x] `pnpm test` — 288 test files / 4102 tests passed, 2 skipped
- [x] `pnpm lint:noconsole` — green
- [x] `pnpm smoke:e2e` — green
- [x] `pnpm test:e2e:distribution` — green except `smoke:codex`, which spawns the real `codex` CLI against the running MCP daemon and fails inside this nested-agent sandbox; the script's own comment already documents this as a known sandbox limitation ("If that happens, run it outside the sandbox"), unrelated to this change
- [x] `docker build -f Dockerfile.server` (both the `build` stage and the full multi-stage image) — succeeds with `corepack prepare pnpm@11.12.0`
- [x] Fresh-worktree install timing/size comparison (see above)
- [ ] `smoke:codex` — needs to be re-verified outside a nested-agent sandbox by a human or CI

### DIFF
--- a/.github/actions/setup-pnpm/action.yml
+++ b/.github/actions/setup-pnpm/action.yml
@@ -37,9 +37,15 @@ runs:
     # operator to search for 'integrity' in undefined"). corepack never
     # invokes pnpm's self-update — it fetches and shims the target version
     # directly — so it sidesteps the bug entirely.
+    #
+    # Corepack is being unbundled from Node (nodejs/node#57825): Node 25+
+    # ships without it. This action must keep working once the runner's
+    # Node no longer includes it, so install corepack from npm first when
+    # the runner doesn't already have it on PATH.
     - name: Activate pnpm via corepack
       shell: bash
       run: |
+        command -v corepack >/dev/null 2>&1 || npm install -g corepack
         corepack enable
         corepack prepare --activate
 
--- a/.github/actions/setup-pnpm/action.yml
+++ b/.github/actions/setup-pnpm/action.yml
@@ -1,0 +1,30 @@
+name: Setup Node and pnpm
+description: >-
+  Install the packageManager-pinned pnpm via pnpm/action-setup, then Node with
+  pnpm store caching. Centralizes pnpm setup so a pnpm/tooling version bump
+  only needs to land here once instead of at every workflow call site.
+
+inputs:
+  node-version:
+    description: >-
+      Node version to install. Overrides node-version-file when set — used by
+      the publish-mcp release job, which checks out an old tag whose tree may
+      predate .node-version and must pin Node inline instead.
+    required: false
+    default: ''
+  registry-url:
+    description: npm registry URL to configure (passed through to actions/setup-node).
+    required: false
+    default: ''
+
+runs:
+  using: composite
+  steps:
+    - uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6.0.9
+
+    - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+      with:
+        node-version: ${{ inputs.node-version }}
+        node-version-file: ${{ inputs.node-version == '' && '.node-version' || '' }}
+        registry-url: ${{ inputs.registry-url }}
+        cache: pnpm
--- a/.github/actions/setup-pnpm/action.yml
+++ b/.github/actions/setup-pnpm/action.yml
@@ -1,8 +1,8 @@
 name: Setup Node and pnpm
 description: >-
-  Install the packageManager-pinned pnpm via pnpm/action-setup, then Node with
-  pnpm store caching. Centralizes pnpm setup so a pnpm/tooling version bump
-  only needs to land here once instead of at every workflow call site.
+  Install Node per .node-version and activate the packageManager-pinned pnpm
+  via corepack, with pnpm store caching. Centralizes pnpm setup so a pnpm
+  version bump or tooling fix only needs to land here once.
 
 inputs:
   node-version:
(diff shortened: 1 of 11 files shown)


kamiazya:
## Response to Gemini's review findings on `pnpm-workspace.yaml`

Verified all three against https://pnpm.io/settings (whose own page title is literally "Settings (pnpm-workspace.yaml)") and empirically. Summary: **#1 and #2 rejected as stale, #3 not applicable (already correctly quoted).**

### 1. `enableGlobalVirtualStore` — rejected, this IS a valid pnpm-workspace.yaml key

pnpm.io/settings documents `enableGlobalVirtualStore` (added in v10.12.1) as a `pnpm-workspace.yaml` setting, not an `.npmrc`-only one. pnpm 10.6+ moved general config into `pnpm-workspace.yaml`, and pnpm 11 no longer reads `package.json#pnpm` at all — this PR exists precisely because of that migration.

We also have empirical proof it is read, not ignored: toggling `enableGlobalVirtualStore: true` measurably changed the install layout (`node_modules` became a symlink tree into `~/Library/pnpm/store/v11/links/...`, and total worktree size dropped from ~1.1GB to ~1.5MB). A setting pnpm silently ignores cannot do that. (We currently ship it `false` for an unrelated TS2740 symlink-depth issue with `@excalidraw/excalidraw` — documented inline above the key.)

### 2. `allowBuilds` — rejected, this is the *correct* v11 key, not `onlyBuiltDependencies`

Docs confirm the opposite of the finding: `onlyBuiltDependencies` / `onlyBuiltDependenciesFile` / `neverBuiltDependencies` / `ignoredBuiltDependencies` / `ignoreDepScripts` were **removed in pnpm 11** and replaced by `allowBuilds` (https://pnpm.io/settings#allowbuilds, migration section: "The following settings have been removed in v11 and replaced by allowBuilds... run `pnpx codemod run pnpm-v10-to-v11`").

While verifying this we found our `pnpm-workspace.yaml` was actually carrying *both* keys — the now-dead `onlyBuiltDependencies: [esbuild]` alongside the correct `allowBuilds: { esbuild: true, ... }`. Fixed in a follow-up commit (dropped the defunct key).

Empirically confirmed the build script Gemini worried about is NOT silently skipped: a clean `pnpm install` hardlinks esbuild's real native Mach-O binary into `bin/esbuild` (esbuild's `install.js` `maybeOptimizePackage` step, which only runs when the postinstall is allowed to execute). If `allowBuilds` were being ignored, `bin/esbuild` would remain the original JS shim script instead.

### 3. Override key quoting (`@`/`<`/`>`) — not applicable, already correct

Parsed `pnpm-workspace.yaml` with a YAML library and confirmed every override key round-trips correctly, including the mid-string `<`/`>`/`>=` forms (`nanoid@>=4.0.0 <5.0.9`, `qs@>=6.11.1 <6.15.2`, etc.) — these characters are only YAML indicators at specific scalar-start positions, not mid-string, so no quoting is needed for them. The one key that *does* need quoting because it starts with `@` (`@grpc/grpc-js@<1.14.4`) was already quoted in the file before this review. So this is cosmetic/already-handled, not a live bug.

PR #216, opened 2026-07-12
fix(web): confirm copy-canvas-URL success/failure instead of failing silently

kamiazya:
## Summary

- The only clipboard affordance in `apps/web` (canvas actions menu → "Copy canvas URL", `WorkspaceTopBar.tsx`) copied to the clipboard and closed the menu with no confirmation at all — clicking it gave no signal whether it worked. Fixes `tmp/issues/canvas-copy-url-no-toast-feedback.md` and `tmp/issues/copy-canvas-url-no-feedback.md` (duplicate reports of the same bug).
- The menu item itself now reports the outcome instead of introducing a second notification system: it stays open and swaps to a checkmark "Copied!" label for 2s on success.
- A rejected `navigator.clipboard.writeText` (permissions, insecure context, browser refusal) now surfaces a visible `role="alert"` error with the URL rendered as a selectable, read-only fallback input — no silent failure, and no false "Copied" success either.
- An `aria-live="polite"` / `role="status"` region announces both outcomes to screen readers, since focus stays on the menu item and a plain text swap isn't guaranteed to be announced.

## Visual repro

Success (checkmark + "Copied!"):

Failure (inline alert + selectable fallback URL):

## Test plan

- [x] `pnpm --filter @kamiazya/whiteboard-web` unit tests (jsdom) — added two new tests asserting the success confirmation + revert, and the rejected-copy error + fallback input; mutation-checked against the pre-fix implementation (both fail as expected).
- [x] `pnpm -r typecheck`
- [x] `pnpm --filter @kamiazya/whiteboard-web build`
- [x] Manual verification against a real Chromium browser (Playwright, headless) driving the running `vite` dev server at `/local/:canvasId` with a stubbed `navigator.clipboard` for both the success and rejection paths — screenshots above.

## Summary by CodeRabbit

* **New Features**
  * Added clear confirmation feedback after successfully copying a canvas URL.
  * Added accessible status announcements for copy results.

* **Bug Fixes**
  * Copy failures now display an actionable message instructing users to copy the URL manually.
  * The actions menu remains open when copying fails, keeping feedback visible.
  * Improved accessibility by placing alerts and announcements outside the actions menu.

### DIFF
--- a/apps/web/src/components/WorkspaceTopBar.browser.test.tsx
+++ b/apps/web/src/components/WorkspaceTopBar.browser.test.tsx
@@ -564,4 +564,34 @@ describe('WorkspaceTopBar browser mode', () => {
     const headerAfter = screen.getByRole('banner')
     expect(headerAfter.getBoundingClientRect().height).toBeCloseTo(48, 0)
   })
+
+  it('announces a failed copy without nesting the alert inside the real role="menu" element', async () => {
+    // Real Chromium rendering (not jsdom) of the Radix menu — the layer where
+    // the ARIA tree axe/AccessLint inspects actually exists.
+    const writeText = vi.fn().mockRejectedValue(new Error('Clipboard permission denied'))
+    Object.defineProperty(navigator, 'clipboard', {
+      value: { writeText },
+      configurable: true,
+    })
+
+    renderTopBar()
+
+    await page.getByRole('button', { name: 'Canvas actions' }).click()
+    await page.getByText('Copy canvas URL').click()
+
+    const alert = await screen.findByRole('alert')
+    expect(alert.textContent).toContain("Couldn't copy automatically")
+
+    const menu = screen.getByRole('menu')
+    expect(menu.contains(alert)).toBe(false)
+
+    // The announcement is still reachable in the accessibility tree even
+    // though it is no longer a DOM child of the menu.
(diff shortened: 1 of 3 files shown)


kamiazya:
Review triage:

**Applied** — the live region was an invalid child of `role="menu"`. `DropdownMenuContent` renders Radix's `role="menu"` container, and the `aria-live`/`role="status"` and `role="alert"` divs sat directly inside it, which violates `aria-required-children` (AccessLint would flag it). Both blocks now live outside `<DropdownMenuContent>`. The announcement is fully preserved — `aria-live` fires wherever the node sits in the DOM; it does not need to be a menu child. Locked in with a real-Chromium browser-mode test asserting the alert/status text is present *and* is not contained by the `role="menu"` element.

**Not applied** — the `window.location.origin` SSR concern. `apps/web` is a pure client-rendered Vite SPA: no SSR or prerender plugin in `vite.config.ts`, a standard Vite client build, and every render path for this component (grepped across `apps/web/src`) goes through jsdom or real Chromium — there is no Node-side render, no docs-snapshot mount, no daemon-side prerender. `ReferenceError: window is not defined` cannot occur here, so a `typeof window !== 'undefined'` guard would be defending against a scenario this codebase does not have.

PR #220, opened 2026-07-12
feat(mcp-server): unify canvas export behind export_canvas, add SVG support

kamiazya:
## Summary

- Closes the SVG gap noted in `tmp/issues/mcp-no-svg-export-tool.md`: adds an `export_svg` MCP tool and a headless-only `POST /export-svg` daemon route, rendering straight from the persisted document (no browser round trip needed — mirrors `canvas_export_json`'s architecture, not `export_png`'s).
- Adds `export_canvas` as the unified entry point: `format: 'png' | 'svg' | 'json'` in one tool, all fields documented via Zod `.describe()`, output typed via `z.infer<typeof exportCanvasOutputSchema>`. It delegates to the existing per-format implementations (`exportPngTool`, `exportSvgTool`, `canvasExportJsonTool`) instead of reimplementing their HTTP/output-path/error-handling logic.
- **Breaking change (0.0.x, no external users per ADR-0001): `export_png` and `canvas_export_json` are removed as standalone MCP tools.** `export_canvas` is now the only registered export tool. This was a direction change mid-PR — an earlier revision kept both old tools registered as deprecated wrappers, but the project owner decided backward compatibility isn't warranted at 0.0.x, so the dual surface was removed rather than kept. The underlying per-format implementations still live in `export.ts` / `canvas-export-json.ts` and are delegated to by `export_canvas`, so nothing changes for callers already on `export_canvas`.
  - Commit is a plain `feat(mcp-server):` (no `!`, no `BREAKING CHANGE:` footer) so `release-please`'s `bump-minor-pre-major` does not fire and the package stays on 0.0.x per the owner's request. The break itself is called out plainly here and in the commit body so `git log` still tells the truth.
  - Updated the registry-wide description guard, smoke coverage lists, `pnpm smoke:e2e` (folded the old assertions into the existing `export_canvas` format:png/svg/json calls, including the headless-render, theme, and embedded-scene-chunk regression checks that used to hang off `export_png`), and every `SKILL.md` / docs reference that named the removed tools.
- Extends `pnpm smoke:e2e` to call `export_canvas` for all three formats against a real running MCP server; mutation-checked (temporarily broke the svg branch's return type, confirmed the smoke fails with the SDK's runtime structuredContent validation, then restored).

## AI review fixes (this revision)

Gemini flagged three medium findings on this PR; verified each against the code before acting:

1. **Default export filename collisions (valid, pre-existing, fixed for all three formats).** PNG, SVG, and JSON export routes all built their default output filename from a millisecond ISO timestamp alone (`${slug}-${timestamp}.ext`) — two exports issued fast enough to land in the same millisecond would collide and the second write would silently clobber the first. This wasn't unique to the new SVG route; PNG (`export.ts`) and JSON (`export-json.ts`) had the identical pattern already. Fixed all three by appending a short `nanoid(6)` suffix, with a red-first test per route proving the collision and then the fix.
2. **Whitespace-only SVG export body (false positive).** The claim was that a pure-whitespace body passes the `rawText.length > 0` guard and then throws inside `JSON.parse`, producing an unhandled 500 instead of 400. Re-reading `export-svg.ts`, the `JSON.parse` call is already wrapped in its own `try/catch` that returns `{ error: 'invalid_request' }` with a 400 — verified with a direct `JSON.parse('   ')` repro (throws `SyntaxError`, caught locally). Added a regression test (`rejects a whitespace-only body with 400 invalid_request instead of throwing`) that locks in this already-correct behavior instead of a production change.
3. **`export_canvas` format fallthrough (valid, fixed).** The format switch's final `else` branch ran the PNG path for *any* value that wasn't `'svg'` or `'json'`, so an out-of-enum format would silently render PNG instead of failing. The registered tool's Zod `inputSchema` enum already blocks this at the MCP boundary, but the implementation itself should not lie about what it did if that boundary is ever bypassed (same class of bug as the `update_element` fix). Changed the fallthrough into an explicit `format === 'png'` branch with a final `throw new Error('Unsupported format: ...')`, red-first via a direct `execute()` call with a cast-bypassed invalid format.

## PDF finding (no code change)

Confirmed (re-derived independently, matches the existing note in `tmp/issues/canvas-export-pdf-not-discoverable.md`): `@excalidraw/excalidraw`'s own export utilities (`exportToBlob` / `exportToSvg` / `exportToCanvas`) only produce PNG, SVG, and canvas — there is no PDF path anywhere in this stack without adding a new PDF-generation dependency. Left out of the unified tool per that issue's existing guidance ("a menu entry for a format that doesn't work is worse than no menu item").

## apps/web symmetry

apps/web's own header export (`useCanvasSync.exportScene`) already covers png/svg (added recently). It has no JSON (`.excalidraw`) download option — filed as a follow-up in `tmp/issues/2026-07-12-apps-web-json-export-missing.md` rather than implemented here, since this PR is scoped to the MCP/daemon export surface and UI work is a separate increment.

## Test plan

- [x] `pnpm test --project mcp-node` (packages/mcp-server) — 201 files / 3117 tests passed
- [x] `pnpm -r typecheck` (mcp-server + web) — clean
- [x] `pnpm build` — clean
- [x] `pnpm smoke:e2e` — real MCP stdio round trip against a live daemon; `export_canvas` verified for all three formats (headless PNG render + embedded Excalidraw scene chunk + theme light/dark diff, SVG asserted to start with `<svg` on disk, JSON asserted `.excalidraw` extension + element round-trip)
- [x] Mutation-check: broke `export_canvas`'s svg branch return type → smoke failed with a Zod validation error from the MCP SDK → restored → smoke green again
- [x] Red-first tests added for all three review-finding fixes (filename-collision tests failed before the `nanoid` suffix, format-fallthrough test failed before the explicit branch/throw)


kamiazya:
Review triage:

**Applied — default export filenames could collide.** Valid, and broader than reported: the millisecond-timestamp-only default was in *all three* export routes (PNG, JSON, SVG), not just the new SVG one, so two exports of the same canvas in the same millisecond would overwrite each other. All three now carry a `nanoid(6)` suffix, each with a red-first test.

**Applied — unsupported `format` fell through to PNG.** Valid, and the most important of the three: `format` is the entire contract of `export_canvas`, and a tool that silently does something other than what was asked is the same class of lie we removed from `update_element` in #215. The switch now has an explicit `format === 'png'` branch and throws on anything unrecognized, so the Zod enum at the boundary and the runtime branch cannot disagree.

**Not applied — whitespace-only request body.** False positive. `JSON.parse` in `export-svg.ts` is already inside a `try`/`catch` that returns a 400 `invalid_body`; a body of `'   '` passes the `rawText.length > 0` guard, throws inside `JSON.parse`, and is caught — it never becomes a 500. Verified with a direct `JSON.parse('   ')` repro against the real handler, and added a regression test locking that behavior so the guard cannot be removed by accident later.

PR #221, opened 2026-07-12
feat(mcp-server): auto-open the browser at the daemon's own origin

kamiazya:
## Summary

- `whiteboard daemon run` now opens the user's default browser at `http://127.0.0.1:<port>` once the daemon is listening — no pairing link, no AI agent needed, since ADR 0001 R3 already has the daemon inject the token server-side into the HTML it serves.
- Default is **on**, guarded off by: `--no-open` / config-file `openBrowser: false`, a non-interactive stdout, `CI` set, a container environment (Docker/Podman, via `/.dockerenv`, `/proc/1/cgroup`, or the `container` env var), or a non-loopback bind.
- Wired only into the packaged CLI's `whiteboard daemon run` path (`dispatcher.ts`). `server/index.ts` (the `tsx watch` dev entrypoint behind `pnpm mcp:http:dev`) and `whiteboard server run` (server-mode) never call it — so dev-watch restarts never spam tabs, and server-mode never pops a browser on a headless host.

## Why default-on

The problem statement is specifically "a human without an agent has no path in" — the guards (TTY/CI/container/loopback/opt-out) already suppress every context where auto-opening would be wrong or meaningless, so default-on directly solves that gap without extra friction for the interactive case.

## Guards and their tests

| Guard | Test |
|---|---|
| `--no-open` / config `openBrowser: false` | `daemon-run-auto-open.test.ts`, `browser-open-policy.test.ts` |
| non-interactive stdout | same |
| `CI` env var | same |
| container (Docker/Podman/containerd) | `container-detection.test.ts` |
| non-loopback bind | `browser-open-policy.test.ts` (defense-in-depth; structurally also enforced earlier by `assertLoopbackBindHost` in `daemon-run.ts`, which refuses non-loopback binds before this code ever runs) |
| dispatcher wiring (host/port/flag/config threading) | `dispatcher-daemon-run-auto-open.test.ts` |

## Docs

- `docs/reference/configuration.md`: new "Auto-opening the browser" section + `openBrowser` config-file key documented alongside `dataDir`'s existing entrypoint-scoping note.
- `docs/how-to/connect-to-local-daemon.md`: "Opening the daemon's own UI directly" now mentions the automatic open and links to the guard list.

## Manual verification

- Built (`tsc -p tsconfig.server.json`) and ran the compiled CLI (`node dist/cli/index.js daemon run --json`) in this non-TTY sandbox — confirmed the ready JSON is emitted and the auto-open guard correctly logs `reason: "non-interactive"` and skips (this harness's stdout isn't a real TTY, so this is exactly the expected/desired behavior here).
- Directly exercised the compiled `maybeOpenDaemonBrowser` with a stubbed `open()`, forcing `isTTY: true`: confirmed it calls `open('http://127.0.0.1:39099')` when every guard passes, and is suppressed under `CI=true` and `--no-open`.
- Could not visually confirm a real browser tab opening in this headless sandbox (no display) — the integrator should do a final interactive-terminal check on a real machine (`whiteboard daemon run --json` from an actual TTY) to see the tab appear.

## Test plan

- [x] `pnpm test --project mcp-node` (packages/mcp-server) — 202 files / 3119 tests passed
- [x] `pnpm --filter @kamiazya/whiteboard-mcp typecheck` (ran `tsc -p tsconfig.server.json --noEmit` directly) — clean
- [x] `pnpm build` (ran `tsc -p tsconfig.server.json` directly) — clean
- [ ] Integrator: interactive-terminal sanity check that the browser tab actually appears

## Summary by CodeRabbit

* **New Features**
  * `whiteboard daemon run` now automatically opens the daemon UI in the default browser after startup when running interactively on a local machine.
  * Added `--no-open` to disable automatic browser opening.
  * Added the `openBrowser` configuration option, with the command-line flag taking precedence.

* **Documentation**
  * Updated daemon and configuration guides to explain browser-opening behavior and its conditions.


kamiazya:
Review triage:

**Applied** — a throw anywhere in the auto-open flow could take the daemon down. Losing your whiteboard daemon because a browser-opening convenience threw is an absurd failure mode. The whole decision+open flow is now inside one try/catch, and on any failure it logs the daemon URL at `warning` level so the human can still open it themselves — a silent "we tried and gave up" would leave them exactly at the dead end this feature exists to close. Red→green test injects a throwing container-detection.

**Not applied** — `process.stdout?.isTTY`. This code runs only through `cli/dispatcher.ts`'s `dispatchRun`, invoked exclusively from the real CLI entrypoint in a normal Node main process: no worker threads, no SSR-like split, so `process.stdout` is always defined. Same class as the SSR guard rejected on #216 — defensive code for a scenario with no demonstrated occurrence here.

---

Also worth recording, because it changes what we thought we were testing: while mutation-checking the E2E coverage (blanking the server-side token injection), **every GET-only assertion still passed**. `/api/*` gates *mutating* methods only, so reads are unauthenticated — a broken token would have shipped green. The smoke now creates a canvas through the UI, which is what actually exercises token injection. Reads being unguarded is a real finding in its own right and is now called out in ADR-0002 as something the authorization work has to fix.

PR #223, opened 2026-07-12
fix(web): keep the unsupported-browser notice and its escape hatch on one line

kamiazya:
The banner's parent lays its children out with `gap-3`, so the notice `<span>` and the "Open the local app" anchor were two separate flex items — the link read as a detached chip and could wrap onto its own line, away from the sentence that explains why it is there. It is now one flex item, so the sentence and its way out stay together.

Follow-up to #222, from a Gemini review comment that landed after the merge.

### DIFF
--- a/apps/web/src/components/migration/DaemonDetectedBanner.tsx
+++ b/apps/web/src/components/migration/DaemonDetectedBanner.tsx
@@ -173,12 +173,15 @@ export function DaemonDetectedBanner({
   return (
     <>
       {showUnsupportedNotice && (
-        <>
-          <span className="text-xs text-muted-foreground">{UNSUPPORTED_BROWSER_NOTICE}</span>
-          <a href={openLocalAppUrl} className="text-xs font-medium underline">
+        // One flex item, not two: the parent lays its children out with a gap,
+        // so a sibling anchor would read as a detached chip and could wrap onto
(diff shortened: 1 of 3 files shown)


kamiazya:
**Applied — and the finding was valid.** `localDaemonBaseUrl` was `z.string().optional()`: no URL validation, no protocol constraint. Anything able to write localStorage (a same-origin script, a browser extension) could have stored `javascript:…` and turned a click on the escape-hatch link into script execution.

Fixed at the **schema**, not at the render site as suggested. The Zod schema is this repo's single source of truth for stored settings, so rejecting the value there protects every consumer — guarding the one `href` that renders it today would leave the next consumer to rediscover the hole.

Tests: three hostile values (`javascript:`, `data:`, a non-URL) are rejected and the whole payload falls back to defaults, plus a positive case proving an otherwise-identical payload with an `http:` URL still loads.

Worth recording how close this came to being a fake test: my first version of it passed **whether or not the constraint existed** — the payload omitted `migration` and `capabilities`, which the strict top-level schema requires, so `safeParse` failed for an unrelated reason and the settings fell back to defaults regardless. Mutation-checking caught it (removing the constraint left the suite green). The committed test now fails for exactly the right reason when the constraint is removed — verified.

PR #224, opened 2026-07-12
docs(adr): re-measure WebSocket LNA gating, add reusable cross-engine harness

kamiazya:
## Summary

ADR-0002's addendum (2026-07-08) measured hosted-HTTPS-to-loopback `fetch` across Chromium/Firefox/WebKit, but explicitly left WebSocket-upgrade behavior under Local Network Access unmeasured ("must be re-verified when it lands"). ADR-0005's connection-ticket design depends on the answer.

Re-measured with a new, committed, reusable harness (`pnpm --filter @kamiazya/whiteboard-mcp smoke:lna-transport`) driving real Chromium/Firefox/WebKit against the real hosted origin (`https://kamiazya-whiteboard.pages.dev`), probing both `fetch` (baseline re-confirmation) and a `WebSocket` upgrade to a throwaway, fully CORS-permissive loopback probe server.

### Measured table

| Engine (build) | fetch (no LNA grant) | fetch (LNA granted) | WS upgrade (no LNA grant) | WS upgrade (LNA granted) |
|---|---|---|---|---|
| Chromium 147.0.7727.15 | Fails fast (~25ms, `permissions.query` → `prompt`) | Succeeds | **Succeeds** | Succeeds |
| Firefox 148.0.2 | Succeeds | n/a | Succeeds | n/a |
| WebKit 2272 (macOS 26.4) | Blocked | n/a | Blocked | n/a |

### Key finding

**The WebSocket upgrade is not currently gated behind LNA on Chromium**, even with an undecided (`prompt`) permission state — a page whose `fetch` to loopback is blocked can still open a raw loopback WebSocket. This means the connection-ticket `fetch` and the socket open do **not** currently need to sit inside the same granted permission — no permission is required for the socket at all today. This is a currently-observed gap in Chromium's rollout (tracked upstream, not yet shipped), not a property the design should assume permanently: the ticket-minting `fetch` must still surface a denied/blocked LNA grant as an explicit error, and the harness should be re-run whenever Chromium moves.

Also noted: the undecided-permission `fetch` failure is now fast (~25ms) rather than the addendum's previously observed 5-second hang — plausibly an automation-specific fast-deny rather than the real interactive prompt UX, which remains unverified by this harness (flagged as such in the doc).

## Changes

- `packages/mcp-server/scripts/smoke/mcp-lna-transport-smoke.mjs` — new committed harness (the one the ADR-0002 addendum promised but never checked in), runnable via `pnpm smoke:lna-transport`.
- `docs/contributing/adr/0002-browser-to-daemon-transport.md` — adds the re-measured WS row/table to the existing addendum.
- `docs/contributing/adr/0005-hosted-origin-authorization.md` — replaces the "must be re-measured" WS note in the LNA constraint section with the actual measured answer.

## Test plan

- [x] `pnpm --filter @kamiazya/whiteboard-mcp smoke:lna-transport` (manual, real browsers, real hosted origin) — see table above.
- [x] `pnpm -r typecheck`
- [x] `pnpm test --project mcp-node` (206 files, 3154 passed)
- [x] `pnpm lint:noconsole`
- [x] `biome check` on new/changed files

### DIFF
--- a/docs/contributing/adr/0002-browser-to-daemon-transport.md
+++ b/docs/contributing/adr/0002-browser-to-daemon-transport.md
@@ -139,9 +139,56 @@ The remaining blocker for hosted pairing is therefore **this repo's own
 daemon policy**, not the browser: with a `pages.dev` Origin the daemon
 returns no `Access-Control-Allow-Origin` on `/api/*`, 403 on `/mcp`, and
 rejects the WS upgrade (loopback-hostname Origin gate) — all verified
-against a running daemon. Chromium's LNA gating for WebSocket is not yet
-shipped (tracked upstream), so WS behavior must be re-verified when it
-lands.
+against a running daemon.
+
+### Re-measured (2026-07-12): WebSocket is still not LNA-gated
+
+The remaining unknown from the addendum above — whether Chromium's LNA
+gating had been extended to the WebSocket upgrade — has been re-measured
+with a committed, reusable harness
+(`packages/mcp-server/scripts/smoke/mcp-lna-transport-smoke.mjs`, run via
+`pnpm smoke:lna-transport`). It drives all three engines against the same
+real `https://kamiazya-whiteboard.pages.dev` origin, this time probing both
+a `fetch` (baseline re-confirmation) and a `new WebSocket(...)` upgrade to a
+throwaway, fully CORS-permissive loopback probe server — not this repo's own
+daemon, so the daemon's own origin gate (F4 above) cannot mask browser-level
+behavior.
+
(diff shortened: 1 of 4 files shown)


kamiazya:
**Applied.** All three comments are the same defect: `server.close()` only stops *new* connections and waits for live ones, so the open probe sockets kept the harness alive. `startProbeServer()` now resolves `{ server, wss }` and the cleanup terminates the WebSocket server first.

Verified by running it: the harness completes and exits cleanly in ~7s (exit 0). That matters more than it looks — the whole point of this harness is that the next person can re-run it in one command when a browser version moves, and a harness that hangs is one nobody re-runs.

PR #310, opened 2026-07-26
feat(canvas-render): add scene graph, pure layout, SVG backend, and sceneDigest

kamiazya:
## What

Adds `packages/canvas-render` — the shared-layer rendering core for OpenCanvas (slice 3). This is the runtime-agnostic engine that turns the canvas data model into visual output on Node, the browser, and Cloudflare Workers alike.

- **Scene graph** — plain-TS node types that retain semantic provenance (a node knows it *is* an ellipse / a heading / a link, not just a shape).
- **Layout** — pure functions with an injected `measure` seam (opentype.js / Canvas `measureText` supplied by the composition root, never imported here). Covers spatial edges, embed recursion, and mdast block/inline flow.
- **SVG backend** — serializes the laid-out scene to an SVG string. `SvgFragmentNode.role` is wired through for a11y.
- **sceneDigest** — bounded spatial digest (what is where / what overlaps) for AI-facing consumption.

The package imports **no `node:*`, no DOM globals, no inversify** — enforced by an import-guard test.

## Rendering determinism / regression coverage

- **SVG string golden snapshot** (`svg/determinism.test.ts` + `.browser.test.ts`) — same input → identical SVG across the node and browser projects. This is the primary, deterministic VRT layer; pixel VRT is deferred to the theme slice (needs a Linux-fixed baseline).

## Security findings fixed during review

Multi-dimension review + adversarial verify surfaced and fixed, each with a dedicated test:

- **math SVG injection** — the escaped-text math *fallback* interpolated untrusted math source into `<text>…</text>` unescaped; now escaped. The trusted SVG-fragment seam keeps its "caller supplies a validated fragment" contract.
- **sceneDigest OOM** — free-region grid resolution scaled with coordinate span (two distant nodes → unbounded allocation); now capped to a fixed max grid, free regions become a documented approximation under the cap.
- **href scheme sanitization** — only `http`/`https`/`mailto` and in-doc `canvas:` refs pass into SVG navigation; `javascript:`/`data:` neutralized.
- **`resolveEmbeds` totality** — bundle lookups guarded against inherited object keys (`__proto__`) so totality can't be broken.
- **XML escaper** — strips XML-1.0-forbidden control chars and lone surrogates.

## Inline layout fix

Inline runs now advance a per-line horizontal cursor (was hardcoded `bbox.x = 0`, so sibling runs overlapped). A hard break resets x and bumps y one line-height (was silently dropped); paragraph/heading `bbox.h` grows to cover every line so following blocks don't overlap. Word-wrap at `maxWidth` stays deferred to the theme/line-layout slice.

## Test plan

- [x] `pnpm test --project canvas-render-node` (81 tests green)
- [x] `pnpm test --project canvas-render-browser` (determinism, green)
- [x] `pnpm test --project mcp-node` (docs-contract test; full suite green)
- [x] `pnpm -r typecheck` (all workspace packages green)
- [x] `pnpm biome check` + `pnpm knip` clean
- [x] each security fix mutation-checked (revert → test fails)

Ships its path-scoped rule `.claude/rules/package-canvas-render.md` in the same increment (architecture-map policy).

## Summary by CodeRabbit

* **New Features**
  * Added the Canvas Render package with Markdown layout, scene graph generation, spatial edge routing, text measurement, scene analysis, and SVG rendering.
  * Added safe XML escaping and URL handling for generated SVG content.
  * Added deterministic output suitable for consistent rendering across environments.

* **Documentation**
  * Documented Canvas Render architecture, constraints, and testing expectations.
  * Updated development and browser-testing guidance to include Canvas Render.

* **Tests**
  * Added comprehensive Node and browser coverage for layout, embeds, SVG output, safety, determinism, and rendering edge cases.


kamiazya:
## CodeRabbit review triage (integrator)

All 8 findings verified against the code and dispositioned:

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | `sanitizeHref` whitespace scheme bypass (CRITICAL) | **Fixed** `6a9184e` — normalize (strip ASCII tab/LF/CR, trim leading C0/space) before the scheme regex. Note: CodeRabbit's suggested regex `[\^@- ]` has a reversed range (U+0040→U+0020) and throws in V8, so a corrected normalization was used. Mutation-checked. |
| 2 | U+FFFE/U+FFFF not stripped | **Fixed** `6a9184e` — added noncharacters to the XML invalid-char set. Mutation-checked. |
| 3 | sceneDigest pairwise O(n²) unbounded | **Fixed** `de7c812` — `PAIRWISE_MAX_ENTRIES` cap mirroring `FREE_REGION_MAX_CELLS`. |
| 4 | self-edge loop only routes right/down | **Fixed** `07b0883` — outward normal per side (left/top/bottom cases added). |
| 5 | xml-well-formed helper accepts unescaped attrs | **Fixed** `4849b13` — validate attribute values; new test file. |
| 6 | architecture-map row (themes/zod) | **Fixed** `408fa1d`. |
| 7 | stale vitest project counts in docs | **Fixed** `408fa1d` — verified 11 projects / 3 browser against `vitest.config.ts`. |
| 8 | table-cell/list-item render at page origin (Major) | **Deferred** — verified real, but a correct fix requires threading an x-origin through every `layoutBlock` case (a naive `<g transform>` double-counts the already-absolute `LIST_INDENT_PX * depth` on nested list items). This is layout-model rework beyond this foundational slice, tracked for the theme/line-layout slice. |

Docstring-coverage warning: not enforced in this repo; not actioned.

kamiazya:
## CodeRabbit triage (integrator)

Triaged all 8 findings against the code. **7 fixed** on this branch (commits `6a9184e`, `de7c812`, `07b0883`, `4849b13`, `408fa1d`):

- **CRITICAL** `sanitizeHref` whitespace-obfuscated scheme bypass (`\tjavascript:` etc.) — fixed by normalizing (strip tab/newline/CR + trim leading C0/space) before the scheme allow-list. (Note: the suggested regex had a reversed `@- ` range that throws in V8, so a corrected normalizer was written instead.)
- U+FFFE/U+FFFF stripped in the XML escaper.
- `sceneDigest` pairwise overlap/containment/cluster now capped (mirrors `FREE_REGION_MAX_CELLS`).
- self-edge loops routed along the selected side's outward normal.
- xml-well-formedness test helper now rejects unescaped attribute values.
- architecture-map row (drop "themes", add `zod`) + stale vitest project counts.

**1 deferred (valid, tracked):** table-cell / list-item content renders at the container's x-origin instead of its `bbox.x` (columns collapse; list indent lost). This is a coordinate-model change spanning table + list + nested-list and belongs with the deferred word-wrap / line-layout work — canvas-render has no downstream consumer yet (its output is unused until the View/UI slice), so no user-facing impact now. Tracked to be fixed in the layout slice, with a table/list golden fixture added so the fix is regression-guarded.

PR #431, opened 2026-08-07
fix(tooling): reach the dev MCP daemon through a retrying stdio proxy

kamiazya:
## Summary

Sessions regularly ran without the whiteboard MCP tools: the client attempts **one** HTTP connection at session start and never retries, so it lost the race against the SessionStart hook spawning the dev daemon; a tsx-watch restart mid-session stranded the connection the same way.

Both clients (`.claude/settings.json`, `.codex/config.toml` `whiteboard_dev`) now register a local **stdio proxy** (`packages/mcp-server/scripts/dev/mcp-http-stdio-proxy.mjs`) instead of the HTTP URL. The spawn always succeeds; the proxy runs the ensure hook, waits for readiness, and forwards each stdin JSON-RPC line as one authenticated `POST /mcp` with a retry budget that absorbs watch restarts. This is sound because the daemon's `/mcp` is already **stateless per request** (fresh server + transport, no `Mcp-Session-Id`, JSON responses) — one line in = one POST = one line out, no protocol session to lose.

Deliberate dev-only tradeoff: server-initiated notifications (tools listChanged) do not traverse the proxy; reload the client session to pick up a changed tool list.

Note on the 2026-07-28 MCP spec / SDK v2: not pursued here — the server already behaves statelessly under the v1 spec, and the connection failures were client-side timing, which a spec/SDK bump would not fix. v2 adoption becomes interesting once clients speak 2026-07-28 natively.

## Verification

- Subprocess tests (mcp-node): forward + auth, retry across a late-starting backend, notification passthrough (no stdout line), clean exit on stdin close — full `scripts/dev` suite 105/105.
- Live: spawned the proxy against the running dev daemon — `initialize` → whiteboard 0.0.19, `tools/list` → 16 tools.
- Docs synced: AGENTS.md (MCP Development Mode), docs/contributing/mcp-debugging.md.

## Summary by CodeRabbit

* **New Features**
  * Added a local connection proxy for development integrations, improving startup handling and resilience during service restarts.
  * Added automatic request retries while the local service becomes ready or restarts.
  * Improved handling of notifications and connection failures during local development.

* **Documentation**
  * Updated development guidance to explain proxy-based connections and direct diagnostic tools.

* **Tests**
  * Added coverage for request forwarding, authentication, retries, notifications, and clean shutdown.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -2,25 +2,27 @@
   "$schema": "https://json.schemastore.org/claude-code-settings.json",
   "_comment": [
     "Repo Claude Code MCP override.",
-    "Connects every reload to the long-lived daemon launched by",
-    "  pnpm mcp:http:dev",
-    "which serves http://127.0.0.1:3099/mcp. The HTTP transport sidesteps the",
-    "stdio + daemon-registry path that would otherwise reuse stale daemons",
-    "recorded in ~/.whiteboard/daemon.json.",
+    "Registers the whiteboard MCP as a LOCAL STDIO PROXY",
+    "(packages/mcp-server/scripts/dev/mcp-http-stdio-proxy.mjs) that forwards",
+    "each request to the long-lived dev daemon at http://127.0.0.1:3099/mcp",
+    "(launched by `pnpm mcp:http:dev`).",
     "",
-    "The SessionStart hook starts the dev daemon automatically. If hooks are",
-    "disabled, Claude Code will see a connection error on the whiteboard MCP",
-    "server. Start it in another terminal first:",
-    "  pnpm mcp:http:dev"
+    "Why a proxy instead of a direct HTTP entry: the client attempts ONE HTTP",
+    "connection at session start and never retries, so it regularly lost the",
+    "race against the SessionStart hook spawning the daemon, and a tsx-watch",
+    "restart mid-session stranded the connection the same way. The stdio spawn",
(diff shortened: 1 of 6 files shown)


kamiazya:
CodeRabbit triage (both reviews, 8 findings):

- **Fixed in 4aad8d8f**: retry-timeout validation (non-finite override no longer wedges the retry loop); spawnProxy env-override merge.
- **Fixed in 33d2e73**: non-JSON HTTP error bodies now fail the request (5xx retried as transient) instead of being forwarded to stdout as protocol responses; notification-test fake/assertion tightened; nextStdoutLine cleans up its listeners/timer on settle; **and the Codex `whiteboard_dev` stdio-proxy config, which the sparse checkout had silently dropped from an earlier commit** — good catch, the AGENTS.md/config mismatch was real.
- **Already addressed** (marked by CodeRabbit): fixed-port wording in settings.json / AGENTS.md transport distinction.
- **Skipped with reason** — `git rev-parse --show-toplevel` → `${CLAUDE_PROJECT_DIR}` in the AGENTS.md registration one-liner: that command runs in the developer's shell at registration time, so it expands to a stable absolute path, which is exactly what a machine-private local-scope entry should store. An unexpanded `${CLAUDE_PROJECT_DIR:-.}` would also work but adds a runtime-expansion dependency for no benefit here.

PR #435, opened 2026-08-07
fix(web): keep the /pair consent route from being rewritten away

kamiazya:
## Summary

Reported from real use: **"Use here" appeared to do nothing.** Driving Chrome against the deployed app and the real daemon found the cause, and the flow works end to end with this fix.

**The bug**: the daemon-served app rewrote `/pair` to `/` on mount — dropping the `origin`/`challenge`/`state` query and landing the user on the daemon's canvas gallery instead of the consent prompt. A grant could therefore never be created, so pairing silently never happened.

**Cause**: the consent page is not a daemon *view*, but the `daemonView → URL` sync effect still ran for it. `parseDaemonRoute('/pair')` is `null`, so the view defaulted to the index and the effect navigated to `/`. The `/pair` short-circuit in render never got a chance because the pathname had already changed. Both sync effects now exclude the route.

**Same dead end, second place**: the multi-responder picker only offered `Open <baseUrl>`, i.e. leave the hosted app — the very thing the hosted-app-first model exists to avoid. Each responder now offers **Use here** (pair in place) with `Open` kept as the fallback.

## Verification

- Regression test at the App level (this is where the defect lived): rendering App at `/pair?…` with local-daemon provider state must keep the pathname and query and render the consent page. It fails on the old code with `expected '/' to be '/pair'` — the exact observed symptom.
- Picker test asserts a `Use here` per responder and that it starts the grant for **that** daemon's baseUrl.
- apps/web jsdom 115 files / 1343 green, typecheck 0, knip clean, bundle gate OK.
- **Live in Chrome, real daemon**: hosted origin → *Use here* → consent page showing the requesting origin → *Approve* → back on `https://latest.kamiazya-whiteboard.pages.dev/` with the daemon's workspace list, then opened `/canvas/<ws>/opencanvas-mcp-server-plan` with the editor mounted. Daemon data operated in place from the hosted app.

## Note on how this escaped

`PairConsentPage.test.tsx` rendered the page component directly, so it never exercised the surrounding App effects; `App.test.tsx` had no `/pair` case. The new test closes exactly that seam.

## Summary by CodeRabbit

* **New Features**
  * Added a “Use here” action for each discovered daemon, allowing users to start pairing without leaving the page.
  * Kept a separate “Open” option for navigating directly to a daemon.

* **Bug Fixes**
  * Fixed pairing consent links so the `/pair` route and challenge parameters remain intact while the consent page is displayed.

### DIFF
--- a/apps/web/src/App.test.tsx
+++ b/apps/web/src/App.test.tsx
@@ -107,6 +107,35 @@ const INVALID_CONFIG_STATE: ProviderState = {
   message: 'Runtime configuration is invalid.',
 }
 
+describe('/pair consent route', () => {
+  it('renders the consent page and does NOT rewrite the URL away from /pair', async () => {
+    // Regression: the daemonView -> URL sync effect ran on mount, saw
+    // parseDaemonRoute('/pair') === null (so daemonView defaulted to the
+    // index), and navigated to '/' — dropping the origin/challenge/state
+    // query and dumping the user on the daemon gallery instead of the
+    // consent page. Observed live on the daemon origin, 2026-08-07.
+    const router = createMemoryRouter(
+      [
+        {
+          path: '*',
+          element: <App providerState={LOCAL_DAEMON_STATE} />,
+        },
+      ],
+      {
+        initialEntries: ['/pair?origin=https%3A%2F%2Fexample.com&challenge=chal&state=st'],
+      },
+    )
+    await act(async () => {
+      render(<RouterProvider router={router} />)
+    })
+
+    expect(router.state.location.pathname).toBe('/pair')
+    expect(router.state.location.search).toContain('challenge=chal')
+    await screen.findByText(/allow this web app to use your local daemon/i)
+    expect(screen.queryByText(/Configured for local daemon/)).toBeNull()
+  })
+})
+
 describe('App backend configuration chip', () => {
   it('shows "Browser only" when configured for browser-local storage', () => {
     render(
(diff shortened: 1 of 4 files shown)


kamiazya:
## Preview verification (per the new merge gate)

Driven in Chrome against the PR preview `https://fix-pair-route-rewrite.kamiazya-whiteboard.pages.dev` and the real local daemon (serving this branch's build):

- **The fixed leg is proven on the preview origin**: `connect anyway` → daemon `/pair?origin=<preview>&challenge=…&state=…` → the consent page renders with the full query intact and the preview origin shown — the exact step that previously rewrote to `/` and dead-ended. Approve persisted the grant and redirected back to the preview origin with the `#wb-grant` fragment.
- **The exchange-and-render leg is proven on the latest origin** (earlier today, same session): Use here → consent → Approve → daemon workspaces listed and a daemon canvas opened in place on the hosted origin. That leg shipped in #433 and is untouched by this PR.
- The preview-origin run could not complete the final exchange in the automated context: each new preview origin needs its own Chrome **Local Network Access** permission, and the browser prompt cannot be accepted by automation. Environment constraint, not this change (a human clicking gets the prompt).

**Found while verifying** (pre-existing, filed as `grant-exchange-failure-invisible`): when the code→token exchange fails, the app falls back to browser-local with no message — the same silent-dead-end class as the probe. Fix queued on the in-flight silent-renewal branch.

PR #436, opened 2026-08-07
feat: silent daemon reconnect for paired hosted origins, grant list/revoke API

kamiazya:
## Summary

Closes the last gap in the hosted-PWA-first pairing loop: **opening the hosted app now reconnects to the stored daemon silently** — no redirect, no re-approval — completing "open latest and your daemon data is just there".

- **Silent renewal (apps/web)**: a cold load on a hosted origin with a stored daemon and no pairing fragment attempts one `grantType: 'origin'` renewal (`POST /api/pairing/token`); the browser-enforced Origin header matched against the daemon's persisted grant is the whole credential. Success renders daemon mode in place; a 403 (revoked) or unreachable daemon collapses to `none` and the app falls back to browser-local exactly as before, with the banner as the way back to a fresh consent. An in-flight `#wb=`/`#wb-grant` flow always wins over renewal.
- **Visible exchange failure**: the silent dead end found during #435's preview verification — Approve on the consent page, then a blocked code exchange (typically Chrome's local-network permission still closed) dropped the user on browser-local with zero feedback — now shows a dismissible alert naming the failure and the next step. Canvas `grant-exchange-failure-invisible`.
- **Grant management API (mcp-server)**: `GET /api/pairing/grants` and `DELETE /api/pairing/grants/:grantId` behind the same `runtime:admin` bar as creation; revocation also kills the origin's live session tokens (a revoked origin keeping a working 24h token would make revoke a lie). The management **UI** lands as the immediate follow-up PR.

## Verification

- App-level tests: renewal renders daemon mode with the minted token; `none` falls back to browser-local; no attempt when nothing is stored; a failed exchange shows the dismissible alert (the alert path uses the REAL `consumeGrantFragment` against an empty transaction, not a mock).
- Module tests: `renewPairingToken` success/403/unreachable; server route tests incl. revoke killing live tokens; route-scope registry entries pinned.
- apps/web jsdom 115 files / 1350 green; mcp-node pairing+security 543 green; typecheck 0; knip clean; bundle gate OK.
- Preview verification per the merge gate to follow on this PR's preview URL (note: each new preview origin needs Chrome's local-network permission granted once by a human; evidence will be posted before merge).

### DIFF
--- a/apps/web/src/App.test.tsx
+++ b/apps/web/src/App.test.tsx
@@ -52,6 +52,15 @@ vi.mock('./hooks/useDaemonConnection.js', () => ({
   useDaemonConnection: () => mockDaemonConnectionResult,
 }))
 
+// Silent-renewal seam: everything else in pairing-grant stays real (the
+// /pair tests exercise the true fragment/PKCE code paths).
+let mockRenewResult: import('./lib/pairing-grant.js').GrantConsumeResult = { status: 'none' }
+const renewPairingTokenMock = vi.fn(async () => mockRenewResult)
+vi.mock('./lib/pairing-grant.js', async (importOriginal) => ({
+  ...(await importOriginal<typeof import('./lib/pairing-grant.js')>()),
+  renewPairingToken: (...args: unknown[]) => renewPairingTokenMock(...(args as [])),
+}))
+
 let receivedDaemonPageProps: Record<string, unknown> | undefined
 // Toggled by the error-boundary test: throwing from the lazily-resolved page
 // exercises the paired branch's boundary, which must sit OUTSIDE Suspense to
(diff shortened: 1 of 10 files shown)


kamiazya:
Merging on the user's instruction with preview verification deferred: the preview-origin run was blocked only by Chrome's per-origin Local Network Access prompt (no human present to click Allow). Groundwork completed against the preview: grant seeded, stored daemon seeded, renewal request confirmed in flight and parked on the permission. **Full verification will be completed on the `latest` alias right after this deploys** — that origin already holds the LNA permission and a persisted grant, so the silent-renewal path can be exercised end to end autonomously; evidence will be posted here.

kamiazya:
## Deferred verification completed (latest alias, post-deploy)

Opened `https://latest.kamiazya-whiteboard.pages.dev/` in Chrome with **zero clicks**: silent renewal fired against the stored daemon and rendered daemon mode in place — workspace listed, tabs (Canvases / Files / Storage) live, canvas grid populated from the real daemon.



**Found while verifying** (filed as `pwa-sw-serves-stale-shell`): the PWA service worker keeps serving the previous app shell after a deploy — reloads stayed on the old bundle until the SW was unregistered manually. This also retroactively explains part of today's 'latest does nothing' report: the tab was likely on a SW-cached shell without the newest fixes. Needs an update strategy (skipWaiting/clients.claim or an update-available toast).

PR #438, opened 2026-08-07
fix(web): re-check for a new service worker on interval and tab focus

kamiazya:
## Problem

After a Cloudflare Pages deploy, an already-visited browser keeps loading the old app shell: the registered service worker serves the cached bundle, and nothing ever re-checks `sw.js`, so the update toast never gets a chance to appear. Observed 2026-08-07 on `latest`: reloads stayed on the old entry bundle while a cache-busted fetch of `/` returned the new one; only unregistering the SW picked up the deploy.

## Fix

The prompt-mode toast infrastructure already existed — the missing piece was the update *check*. `register-sw.ts` now passes `onRegisteredSW` and starts a small scheduler (`sw-update-scheduler.ts`, dynamically imported, 0.42 kB chunk) that calls `registration.update()`:

- every 60 minutes, and
- on `visibilitychange` → visible (tab refocus),

so a deploy becomes an UpdateToast within one focus event or one interval tick. `update()` rejections (offline etc.) are caught and logged; one failure never kills the schedule. `registerType: 'prompt'`, the toast wiring, and `vite-pwa-options.ts` are untouched, and the scheduler stays out of the eager entry chunk (bundle-size smoke passes).

## Test plan

- [x] Unit (jsdom): interval ticks call `update()`; visible-transition triggers a check; hidden does not; rejection is caught + logged and the schedule survives; dispose clears interval + listener; dynamic-import failure path logged
- [x] `register-sw.test.ts`: `onRegisteredSW` wired; undefined registration schedules nothing; all pre-existing cases unmodified
- [x] `pnpm build` green incl. PWA generation; bundle smoke passes
- [ ] Preview verification: two deploys on this PR's preview URL — confirm the toast appears on tab refocus without unregistering the SW (evidence to follow in a comment)

No browser-mode/E2E layer: vitest browser mode cannot install a real production SW, so the unit layer plus the manual two-deploy preview check is the nearest honest coverage.

### DIFF
--- a/apps/web/src/pwa/UpdateToast.test.tsx
+++ b/apps/web/src/pwa/UpdateToast.test.tsx
@@ -21,6 +21,18 @@ describe('UpdateToast', () => {
     expect(onReload).toHaveBeenCalledTimes(1)
   })
 
+  it('renders overlaid on the viewport, not in document flow', () => {
+    // The app shell fills 100dvh, so a static toast appended to <body> lands
+    // BELOW the viewport and is never seen — the exact failure observed on
+    // the deployed preview (toast at y === viewport height). Fixed
+    // positioning with a stacking z-index is what keeps it visible.
+    render(<UpdateToast onReload={vi.fn()} onDismiss={vi.fn()} />)
+
+    const toast = screen.getByRole('status')
+    expect(toast.className).toMatch(/\bfixed\b/)
+    expect(toast.className).toMatch(/\bz-\d+\b/)
+  })
+
   it('hides the toast and calls onDismiss when Dismiss is clicked', () => {
     const onDismiss = vi.fn()
     render(<UpdateToast onReload={vi.fn()} onDismiss={onDismiss} />)
--- a/apps/web/src/pwa/UpdateToast.tsx
+++ b/apps/web/src/pwa/UpdateToast.tsx
@@ -12,10 +12,21 @@ export function UpdateToast({ onReload, onDismiss }: UpdateToastProps) {
   const [dismissed, setDismissed] = useState(false)
   if (dismissed) return null
 
+  // Fixed + z-index, not document flow: the app shell fills 100dvh, so a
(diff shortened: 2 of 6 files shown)


kamiazya:
## Preview verification (two-deploy SW update cycle)

Verified end-to-end on this PR's Cloudflare Pages preview (`https://fix-sw-stale-shell.kamiazya-whiteboard.pages.dev`, Chrome, production build, SW-controlled tab):

1. **Update check works**: with an older deploy's shell loaded and its SW controlling the page (`controller: true`), triggering the scheduler's visibility path made `registration.update()` run and the newer deploy's worker reached `waiting: "installed"` — a deploy became detectable without unregistering anything.
2. **Bug found during verification, fixed in this PR**: the toast this whole flow ends in had *never been visible* — `.pwa-update-toast` has no CSS anywhere, so it rendered as an unstyled static `div` after the 100dvh app shell, at exactly `y == viewport height` (measured `{y: 857, viewport: 857}`). The scheduler fired, `onNeedRefresh` fired, and the user still saw nothing. Fixed by styling the toast with the app's tokens and fixed positioning (bottom-center, `z-50`); regression-guarded by a jsdom test asserting the positioning classes.
3. **After the fix**: clicking the (previously invisible) toast's Reload swapped the tab to the new shell (`index-DINV0MDS` → `index-Crbb_X8S`), the scheduler then detected the *next* deploy on this branch, and the styled toast rendered visibly at bottom-center (`{y: 797, h: 44, viewport: 857}`):



Note on method: the automated Chrome window stays unfocused, so `visibilitychange → visible` was triggered by overriding `document.visibilityState` and dispatching the event — the deployed scheduler code (listener + visibility check + `update()` call) is exactly what ran; only the OS-level focus was synthetic. The 60-minute interval path is unit-tested with fake timers.

The intermediate "deploy marker" commit used to force distinct builds was dropped before the final push; the verified running shell is the current head's build.

PR #439, opened 2026-08-07
fix(web): show an in-flight state and a local-network hint on daemon check

kamiazya:
## Problem

Clicking "Check for local daemon" on the hosted origin sometimes appears to do nothing. Two compounding cases:

1. Chrome's Local Network Access permission prompt lives in browser chrome, not the page — while it is pending, every loopback probe is parked on the user's decision and the page shows zero feedback.
2. Even without the prompt, the probe sweep runs up to its timeout with the button static, inviting a useless second click (the in-flight map dedupes silently).

## Fix

In `DaemonDetectedBanner.tsx`:

- While a sweep is outstanding, the trigger button is disabled and a "Checking…" indicator renders (announced via `role="status"`).
- On https: origins only, if the check has been outstanding for more than ~1s, a hint appears: the browser *may* be asking for permission to reach local devices (phrased as a possibility — the same stall also happens from plain probe timeouts). The hint clears the moment the sweep settles, and the timer is cleaned up on settle/unmount.
- A stale failure notice is now cleared when a retry starts.

The discovery-layer dedupe is unchanged; this is purely presentation of the already-existing in-flight window. No files outside `components/migration/**` are touched.

## Test plan

- [x] jsdom component tests: disabled button + Checking… during an unresolved sweep; hint after ~1s on https:; hint never on http: (auto-probe and manual); everything resets on settle; early settle never flashes the hint; unmount with an armed timer is clean; all pre-existing cases pass
- [x] `pnpm test` green in the worktree
- [ ] Preview verification: on the PR preview origin with no daemon running, click the check and observe Checking… + hint (screenshot to follow)

No browser-mode layer: the real LNA prompt cannot be automated, and jsdom + the injected probe seam fully models the in-flight window.

### DIFF
--- a/apps/web/src/components/migration/DaemonDetectedBanner.test.tsx
+++ b/apps/web/src/components/migration/DaemonDetectedBanner.test.tsx
@@ -754,5 +754,38 @@ describe('DaemonDetectedBanner', () => {
 
       expect(screen.queryByText(LNA_HINT_TEXT)).toBeNull()
     })
+
+    it('clears the stale failure notice once a retry sweep starts', async () => {
+      const { promise: firstProbe, resolve: resolveFirst } = deferred<DaemonProbeResult>()
+      const probeFn = vi.fn().mockReturnValueOnce(firstProbe)
+      render(
+        <DaemonDetectedBanner
+          settingsStore={makeStore()}
+          fetch={vi.fn()}
+          locationProtocol="https:"
+          probeFn={probeFn}
+        />,
+      )
+
+      const button = screen.getByRole('button', { name: /check for local daemon/i })
+      act(() => {
+        fireEvent.click(button)
+      })
+      await act(async () => {
+        resolveFirst({ detected: false, reason: 'network' })
+        await vi.advanceTimersByTimeAsync(0)
+      })
+      expect(screen.getByTestId('daemon-check-failed-notice')).not.toBeNull()
+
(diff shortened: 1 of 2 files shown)


kamiazya:
## Preview verification (https://fix-daemon-check-pending.kamiazya-whiteboard.pages.dev)

Verified on the PR preview origin (production build, Chrome). The in-flight window is bounded by the 2s probe timeout, so a static screenshot can't reliably catch it — evidence below is DOM sampling at 250ms intervals right after clicking "Check for local daemon" (deduplicated):

```json
[
 {"t":250,  "text":"Check for local daemon", "disabled":true,  "status":"Checking…", "hint":false},
 {"t":750,  "text":"Check for local daemon", "disabled":true,  "status":"Checking…", "hint":true},
 {"t":1000, "text":"Check for local daemon", "disabled":false, "status":"",          "hint":false}
]
```

- `disabled` = the trigger button's disabled attribute; `status` = the `role="status"` live region; `hint` = the local-network-permission hint text present in the DOM.
- Confirms: button disables + "Checking…" announces immediately on click → the LNA hint joins while the sweep is still outstanding → everything resets the moment the sweep settles, after which the existing "No daemon reachable from this origin…" failure notice renders (screenshot attached in the thread).

Sequence matches the jsdom test expectations 1:1 on the deployed bundle.

PR #443, opened 2026-08-07
feat(web): pin the daemon identity at consent and verify renewals against it

kamiazya:
## What

Browser half of the approved mutual-auth design (canvas `daemon-mutual-auth-design`; server half landed in #441/#442). The web app now **pins the daemon's Ed25519 public key at /pair consent time and verifies every later signed response against the pin** — a port-squatting local process advertising its own key gains nothing.

- **`lib/daemon-identity-pin.ts`** (new): localStorage pin store keyed by daemon baseUrl (corrupt store degrades to unpinned, re-pins on next consent), WebCrypto Ed25519 verification byte-matching the daemon's JSON-array payload encoding, challenge nonces, and an `XXXX-XXXX` base32 fingerprint (first 40 bits of sha256(key)).
- **/pair consent page**: fetches its own origin's ping (daemon-served → the daemon introducing itself over the approved top-level navigation), shows the fingerprint to the user, and embeds the key in the `#wb-grant` return fragment.
- **Token exchange** (`consumeGrantFragment`): challenges with a nonce; the response must be signed by the approved fragment key over `["wb-token-v1", nonce, origin, sha256(token), expiresAt]` — otherwise `daemon identity verification failed` and no pin. Success pins.
- **Silent renewal** (`renewPairingToken`): verifies against the **pin**, never the advertised key. Mismatch → new `identity-mismatch` result: renewal refused (fail closed), pin kept as evidence, and App shows a "daemon's identity changed — re-approve" warning alert.
- **Storage tab**: the Paired web apps card shows the daemon's fingerprint for out-of-band cross-checking against the consent page.
- Legacy daemons without an identity keep working unpinned (no behavior change).

## Test plan

- [x] Pin store round-trip/normalization/corrupt-degrade; WebCrypto verify accepts genuine + rejects tampered message, different-key (squatter) signature, malformed key; fingerprint determinism/format; sha256 vector matches the daemon encoding
- [x] Exchange: nonce always sent; fragment-key verification pins on success; different-signer response refused with no pin
- [x] Renewal: pinned+valid → paired; pinned+wrong-key → `identity-mismatch` with pin retained; unpinned → no nonce (legacy path)
- [x] Consent page: fingerprint rendered; grant call unaffected; Deny still touches no pairing API
- [x] Full apps/web suite 1388 passed; `-r typecheck`, biome, knip green
- [ ] Preview verification: full pair → pin → renewal loop against a local daemon running the #441 server code (evidence to follow)

PR #445, opened 2026-08-07
fix(web): actually surface the identity-mismatch renewal warning

kamiazya:
## What

Live verification of #443 on the deployed hosted app (planted a wrong pin for the dev daemon, reloaded) found the silent-renewal callback in `App.tsx` only committed `'paired'` results to state — an `identity-mismatch` outcome was computed by `renewPairingToken`, then **dropped**, so the fail-closed "daemon's identity changed" warning could never render. The alert JSX and the lib-level fail-closed path were both unit-covered; the one un-covered wiring line between them was the bug.

- `App.tsx`: commit both `'paired'` and `'identity-mismatch'` renewal outcomes to `grantConnection`.
- `App.test.tsx`: new scenario test — renewal returns `identity-mismatch` → stays on browser-local AND shows the warning alert.

## Test plan

- [x] New App-level test red on the old code, green now; full App suite 43 passed
- [ ] Post-merge: re-run the planted-wrong-pin scenario on the deployed `latest` and confirm the warning renders (evidence to follow on this thread)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved silent pairing renewal handling when an identity mismatch occurs.
  * The app now remains in browser-local mode and displays a warning when a connected identity changes.
  * Successful renewals continue to be saved correctly.

### DIFF
--- a/apps/web/src/App.test.tsx
+++ b/apps/web/src/App.test.tsx
@@ -175,6 +175,31 @@ describe('silent renewal on a hosted origin', () => {
     expect(screen.queryByTestId('daemon-index-page')).toBeNull()
   })
 
+  it('surfaces the identity-mismatch warning when renewal fails closed', async () => {
+    localStorage.setItem(
+      STORAGE_KEY,
+      JSON.stringify({
+        version: 1,
+        storage: { localDaemonBaseUrl: 'http://127.0.0.1:3099' },
+        migration: {},
+        capabilities: {},
+      }),
+    )
+    mockRenewResult = { status: 'identity-mismatch', daemonBaseUrl: 'http://127.0.0.1:3099' }
+    await act(async () => {
+      render(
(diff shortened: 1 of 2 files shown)


kamiazya:
## Post-merge verification on the deployed `latest`

Re-ran the planted-wrong-pin scenario against the production deploy of this fix (`69dee0df`): unregistered the SW + cleared caches, planted a wrong pin for `http://127.0.0.1:3099` in `whiteboard:daemon-identity-pins`, reloaded. The silent renewal challenged the real dev daemon, its signature failed against the planted pin, and the fail-closed warning now renders:



- The app stays on browser-local (no daemon connection was made with the unverified responder).
- The pin was retained (the warning's evidence), per the approved design.
- Same scenario on the pre-fix bundle rendered nothing — the exact gap this PR closes.

Cleaned the planted pin afterwards; the next real /pair approval re-pins normally.

PR #447, opened 2026-08-07
feat(export): render real dark chrome when theme:'dark' is requested

kamiazya:
## Problem

`theme: 'dark'` swapped only the document background while the scene was always built light-themed — edge strokes and edge labels resolved to `#333333` and sat directly on the `#121212` background, effectively invisible. Nodes looked fine (light fills), which made the loss easy to miss: a diagram whose meaning lives in its arrows lost exactly that when exported dark. (Canvas `dark-export-loses-edge-labels`.)

## Decision

The canvas framed the choice: make the option honest, or remove it as half-implemented. **Made it honest** — the dark palette already exists and is WCAG-contrast-tested (theme layer, decision #8), so wiring it is small and the option is already documented API surface. The load-bearing invariant is preserved: `theme` is an explicit per-request argument; ambient UI theme still never influences exported bytes, and **light exports are byte-identical to before**.

## Implementation

- `headless-renderer`: the scene is built with `createSpatialTheme({ mode })` for the requested theme (dark node chrome: `#9BA3AF` outlines/edges, `#E6E8EB` labels, transparent fills — same as the editor's dark mode).
- Dark node fills are transparent, and markdown body runs carry no fill of their own (the canvas's "adjacent, worth checking" finding) — so `canvas-render` gains `SvgDocumentOptions.textFill`: an inheritable `fill` on the root `<svg>`, the document-level analogue of the editor host's inherited CSS fill. Additive by contract: absent/empty → byte-identical output; existing goldens untouched. Fixed root attribute order extends to `xmlns width height viewBox fill`.
- Palette comment + `package-canvas-render.md` decision #8 note updated to record the taken behavior decision.

## Verification

Real dark PNG rendered through the actual export pipeline (opentype measure + resvg raster):

Node outlines, body text, edge stroke, and the edge label are all legible on the dark background.

## Test plan

- [x] canvas-render: `textFill` emitted after `viewBox`; absent/empty byte-identical; escaping; envelope activation (red first)
- [x] headless-renderer: dark SVG contains dark edge stroke + label fill + root fill and no light chrome; light SVG unchanged with no root fill (red first)
- [x] Full mcp-node + canvas-render-node suites (3339+), typecheck, biome, knip

### DIFF
--- a/.claude/rules/package-canvas-render.md
+++ b/.claude/rules/package-canvas-render.md
@@ -211,12 +211,15 @@
    to delete. The editor's dark palette (contrast-tested against the WCAG
    1.4.11/1.4.3 floors) is the shared theme's dark palette; viewer and export
    both pin `mode: 'light'` at their call sites, preserving the invariant
-   that a user's UI theme can never change exported bytes. NOT fixed by this
-   decision: `mcp-server`'s `headless-renderer`'s `theme: 'dark'` export
-   option still only changes the document background, not node chrome —
-   giving export a real dark chrome variant is a behavior decision (export
-   gaining new visual output as a function of a request option), not a
-   convergence one, and is deliberately out of scope here.
+   that a user's UI theme can never change exported bytes. The dark-export
+   behavior decision was later taken (2026-08-08): `headless-renderer`'s
+   `theme: 'dark'` now builds the scene with `createSpatialTheme({ mode:
+   'dark' })` and sets `SvgDocumentOptions.textFill` (an inheritable root
+   `fill` — the document-level analogue of the editor host's inherited CSS
+   fill) so body runs stay legible on the dark background. `theme` is an
+   explicit per-request argument; the invariant that ambient UI theme never
+   changes exported bytes still holds, and light exports are byte-identical
+   to before.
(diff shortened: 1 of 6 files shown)


kamiazya:
## CodeQL triage (alerts 33-35, js/html-constructed-from-input)

Verified against the code and judged **false positive** for this sink:

- The flagged construction interpolates `textFill` only through `escapeXmlAttr`, which escapes all five XML-significant characters (`& < > " '`, `packages/canvas-render/src/svg/format.ts:35-41`) and strips XML-invalid control chars — breaking out of the double-quoted attribute value is not possible.
- This is byte-for-byte the same pattern as the adjacent, previously-shipped `background` attribute (same escape, same sink), and the serializer's escaping guarantee is exactly what `canvas-viewer`'s `dangerouslySetInnerHTML` is documented to rely on (`.claude/rules/package-canvas-viewer.md`).
- The behavior is pinned by a unit test: a `"&`-bearing `textFill` must serialize as `fill="&quot;&amp;"`.

CodeQL's taint pass flags any string-built markup reaching an HTML sink and cannot see the escaping contract; no code change made.

PR #448, opened 2026-08-07
feat(web): label a pinned daemon 'identity verified' after a live challenge

kamiazya:
## What

Completes the mutual-auth design's discovery-labeling leg (canvas `daemon-impersonation-port-squatting`, the last remaining item): when the single detected responder's baseUrl carries a **pinned identity** (approved on /pair before), the banner challenges it live via `POST /api/runtime/verify` — fresh nonce, signature over `["wb-verify-v1", nonce, origin]` — and verifies against the **pin**, never the advertised key. Only a passing challenge earns "(identity verified)" in the copy.

Fail-cautious by design: a pinned responder that cannot answer its own challenge (wrong key, missing route, network failure) is **downgraded to the cautious unverified copy even when it matches the stored paired target** — a daemon we once pinned must be able to prove itself, so absence of proof is treated as no proof. Unpinned responders keep the existing unverified behavior byte-for-byte.

- `lib/daemon-identity-pin.ts`: `challengeDaemonIdentity` → `'verified' | 'failed' | 'unpinned'` ('unpinned' never fetches).
- `DaemonDetectedBanner`: injectable `challengeFn` seam (same pattern as `probeFn`), challenge keyed on the detected baseUrl, three-way label.

## Test plan

- [x] Lib: unpinned short-circuits with no fetch; genuine WebCrypto-signed answer verifies; wrong key / 404 / network failure all fail (red first)
- [x] Banner: pinned+verified renders the verified label and challenges the right baseUrl; pinned+failed drops the trust copy; all 60 pre-existing banner tests unchanged
- [x] Full apps/web suite 1391, typecheck, biome, knip
- [ ] Live check on latest after merge (pin the dev daemon via /pair, observe the verified label) — will report on this thread

Backend routes are #441's, already live-verified with WebCrypto end-to-end.

### DIFF
--- a/apps/web/src/components/migration/DaemonDetectedBanner.test.tsx
+++ b/apps/web/src/components/migration/DaemonDetectedBanner.test.tsx
@@ -334,6 +334,51 @@ describe('DaemonDetectedBanner', () => {
     })
   })
 
+  it('a pinned, challenge-verified daemon earns the "identity verified" label', async () => {
+    const probeFn = vi.fn().mockResolvedValue(DETECTED)
+    const store = makeStore()
+    store.update((current) => ({
+      ...current,
+      storage: { ...current.storage, localDaemonBaseUrl: 'http://127.0.0.1:3099' },
+    }))
+    const challengeFn = vi.fn(async () => 'verified' as const)
+    render(
+      <DaemonDetectedBanner
+        settingsStore={store}
+        fetch={vi.fn()}
+        locationProtocol="http:"
+        probeFn={probeFn}
+        challengeFn={challengeFn}
+      />,
+    )
+
+    await screen.findByText(/identity verified/)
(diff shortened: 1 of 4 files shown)


kamiazya:
## Post-merge live verification on the deployed `latest`

Against the production deploy of this commit (`3530a73d`) and the real dev daemon:

1. Pinned the daemon's actual identity key for `http://127.0.0.1:3099`, revoked the latest-origin grant server-side so silent renewal would not auto-connect (leaving the app browser-local with a pin in place).
2. Clicked "Check for local daemon" — the sweep found the daemon, the banner challenged it via `POST /api/runtime/verify`, and the signature verified against the pin:



> "A local whiteboard daemon is running at http://127.0.0.1:3099 **(identity verified)**."

Restored the grant afterwards (201) so the normal renewal flow keeps working.

PR #450, opened 2026-08-07
test(web): pin the editor's single content path and command boundary (S10)

kamiazya:
Closes the OOUI initiative's opportunistic S10 audit slice (test-only):

- **Single content path**: `dangerouslySetInnerHTML=` may appear only at the two documented scene-svg sinks (`SpatialEditor.tsx`, `DragPreviewLayer.tsx`), both fed exclusively by canvas-render's escaping serializer. Any third injection point fails the directory-wide source scan.
- **Command boundary**: every `onChange` call site in the editor hands over a canvas produced by `applyCommand` (directly or via the running accumulator) — a hand-rolled canvas mutation bypassing the typed `EditorCommand` layer fails the scan.

Audited the shipped palette/tool code against both properties first (clean); the tests pin them. Test-only change — no user-visible surface (preview verification N/A). Full spatial-editor suite green.

## Summary by CodeRabbit

* **Tests**
  * Added coverage to verify that raw HTML injection is limited to approved components.
  * Added checks ensuring canvas changes use the supported command pathway rather than bypassing it.

### DIFF
--- a/apps/web/src/components/spatial-editor/purity-guard.test.ts
+++ b/apps/web/src/components/spatial-editor/purity-guard.test.ts
@@ -37,3 +37,38 @@ describe('theme resolver purity (no ambient DOM read)', () => {
     expect(source).not.toMatch(/#[0-9a-fA-F]{3,8}\b/)
   })
 })
+
+describe('single content path (S10 guardrail)', () => {
+  it('raw HTML injection exists ONLY at the two documented scene-svg sinks', async () => {
+    // Every user-visible string in the editor renders either as plain React
+    // text (palette labels, notices) or through canvas-render's escaping
+    // serializer injected at exactly these two places. A third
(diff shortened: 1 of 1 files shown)


kamiazya:
Applied all three findings in the latest commit: .spec files excluded from the scan; the raw-HTML tripwire now also catches innerHTML/outerHTML assignment and insertAdjacentHTML (allowed nowhere) and its comment states explicitly that the escaping guarantee lives in canvas-render's serializer tests, not this scan; the command-boundary matcher drops the comment-count subtraction and excludes prose by requiring code punctuation before the call. Full AST data-flow analysis was deliberately not adopted: these are tripwires whose job is to force a reviewed diff when a new sink or bypass appears — the behavioral guarantees are carried by the serializer tests and the browser-tier command-replay tests, and an AST pass would duplicate those at much higher maintenance cost.

PR #453, opened 2026-08-07
feat(web): select an edge by clicking its line, Delete removes it

kamiazya:
## What

Create/delete symmetry for edges (dogfood finding `edges-cannot-be-selected-or-deleted`): the Connect tool made edge **creation** a two-click flow, but a misclicked edge could not be removed at all. Now:

- Clicking an edge's line (6px screen-space tolerance, zoom-aware via `distanceToPolyline`) selects it and shows a highlight overlay.
- `Delete`/`Backspace` removes the selected edge (`delete-edge` command through the `applyResult` boundary); `Escape` and empty-space clicks clear the selection without deleting.
- Node presses clear any edge selection; text-entry targets never trigger the delete.

## Follow-up caught by live verification (second commit)

The browser tests dispatched synthetic keydowns directly on the editor root, which bypasses focus. Live verification showed the real keyboard's Delete never arrived: an edge has no focusable element (node shapes carry `tabIndex`; polylines do not), so focus stayed on `<body>`. Fix: the root is click-focusable (`tabIndex={-1}`, outline suppressed) and takes focus at the **release** of an edge-selecting press — the browser's default mousedown focus handling runs after the pointerdown listener and undoes a focus taken there. A new regression test routes the key through `document.activeElement` via `userEvent.keyboard`, pinning the real path.

## Visual repro (live dev app, real pointer + real keyboard)

Edge selected by clicking its line (highlight; `document.activeElement` = editor root):

After pressing Delete — the edge is gone (4 → 3 polylines in DOM):

## Test plan

- [x] New real-pointer browser tests (`edge-select-delete.browser.test.tsx`): click-select + Delete removes (nodes untouched); near-miss selects nothing; Escape/empty-click clears without deleting; real-keyboard Delete after a cold edge click via `userEvent.keyboard` (red first — it reproduced the live focus gap)
- [x] `delete-edge` command unit coverage via `applyCommand`
- [x] All 217 apps/web browser tests, 1393 jsdom tests, typecheck, biome, knip
- [x] Live verification in the running dev app with real pointer/keyboard (screenshots above)

### DIFF
--- a/apps/web/src/components/spatial-editor/SpatialEditor.tsx
+++ b/apps/web/src/components/spatial-editor/SpatialEditor.tsx
@@ -478,6 +478,10 @@ export const SpatialEditor = forwardRef<SpatialEditorHandle, SpatialEditorProps>
       // re-renders because it compares node ids, not DOM identity.
       // Shift-click builds a multi-selection instead of starting a gesture.
       if (e.shiftKey && hitId !== undefined) {
+        // Node and edge selection are mutually exclusive: Delete processes a
+        // selected edge FIRST, so an edge left selected here would be what a
+        // Delete on the node multi-selection actually removes.
+        setSelectedEdgeId(null)
         if (selectedId === null) {
           setSelectedId(hitId)
         } else if (hitId === selectedId) {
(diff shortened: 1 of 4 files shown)


kamiazya:
Both CodeRabbit findings addressed in 802a47de:

- **Mutual exclusivity (Major, outside diff)**: valid — Shift-click returned before clearing `selectedEdgeId`, and a marquee starting on an edge line kept the press-time edge selection; Delete would then remove the edge instead of the nodes. Both paths now clear the edge selection, with red-first browser regressions for each.
- **Empty-space deselection test (Minor)**: the test now exercises the pointer path too (reselect → click empty space → highlight cleared, edge intact).

PR #455, opened 2026-08-07
feat(web): edge context menu with Edit label and Delete

kamiazya:
## What

Dogfood finding, following #453/#454: right-clicking an edge's line fell through to the **empty-space** menu ("Add note here") — the object under the pointer got no actions, unlike nodes, which offer Edit text / Delete. The context-menu hit-test now includes edges at the same zoom-aware tolerance as the click path (`EDGE_HIT_TOLERANCE_PX`, hoisted to module scope now that two paths share it), selects the edge for visual feedback, and offers:

- **Edit label** — opens the inline label editor from #454
- **Delete** (danger) — `delete-edge` through the `applyResult` boundary

Node and empty-space menus are unchanged.

## Visual repro (live dev app)

Right-click on the edge line — the edge highlights and gets its own menu instead of "Add note here":

After choosing Delete, the edge is gone (3 → 2 polylines in DOM), nodes untouched:

## Test plan

- [x] Real-pointer browser tests (`context-menu.browser.test.tsx`, red first): edge right-click offers Edit label + Delete and NOT "Add note here"; Edit label opens the editor prefilled; Delete removes the edge, emits `delete-edge`, leaves both nodes
- [x] Existing node / empty-space context-menu tests unchanged and green (6/6 in file)
- [x] Full apps/web browser suite (225 passed; the 1 failure is the known unrelated `source-pane-keymap` load-flake — passes in isolation and in CI, now tracked as canvas `source-pane-softwrap-load-flake`), 1397 jsdom, typecheck, biome, knip
- [x] Live verification in the running dev app (screenshots above)

## Summary by CodeRabbit

* **New Features**
  * Added context-menu actions for editing edge labels and deleting edges.
  * Right-clicking an edge now selects it and displays relevant actions without offering node creation.

* **Bug Fixes**
  * Improved edge detection for click and context-menu interactions.
  * Edge labels and connected nodes are preserved correctly during context-menu operations.

### DIFF
--- a/apps/web/src/components/spatial-editor/SpatialEditor.tsx
+++ b/apps/web/src/components/spatial-editor/SpatialEditor.tsx
@@ -130,6 +130,9 @@ export interface SpatialEditorHandle {
 
 const EDGE_LABEL_EDITOR_WIDTH_PX = 160
 const EDGE_LABEL_EDITOR_HEIGHT_PX = 28
+/** Screen-space px within which a press/right-click counts as hitting an
+ * edge line; divided by the zoom for the canvas-space comparison. */
+const EDGE_HIT_TOLERANCE_PX = 6
 const DEFAULT_TEST_ID = 'spatial-editor'
 /**
  * Window for OUR double-press detection (see handlePointerDown). Matches the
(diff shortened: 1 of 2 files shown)


kamiazya:
CodeRabbit finding addressed: valid — context-menu targeting now clears the other object type's selection in both directions (node right-click clears the edge selection; edge right-click clears node selection + extras), matching the pointerdown path's mutual-exclusivity rule from #453. Browser regression added (red first, mutation-checked against the fix).

PR #460, opened 2026-08-08
feat(web): canvas-only top level — Storage and pairing move into Settings

kamiazya:
## What

**Design refactor D2** (approved brief `design-refactor-brief-2026-08-08`): the top level becomes **one canvas surface**.

- The daemon index's Canvases / Files / Storage tablist is gone. **Files was never a different place** — it showed the same canvases as an alias tree + preview — so it becomes a **view toggle** (Grid ⧉ / Tree ☰) of the one surface instead of a sibling tab whose name contradicted its content.
- The Storage tab's operational surfaces (**Paired web apps**, **Storage usage**) move under **Settings** (new gear on the index toolbar), via a new `extraSections` slot on `SettingsPanel`. Security/ops management is a different audience at a different frequency; it no longer occupies top-level navigation.
- Legacy copy purged: "Excalidraw library packs" → "Library packs" (the feature survives; the era branding doesn't), and the raw `Error: HTTP 401` line is now a sentence ("Couldn't load storage usage (…)").

## Visual repro (live dev app, paired via the D1 chip popover flow)

Before — three parallel tabs, Files claiming to be a place, Storage top-level:

After — one canvas surface; Grid/Tree toggle + Settings gear on the right:

Settings hosts Capabilities + Paired web apps + Storage usage (with the new error copy):

## Test plan

- [x] Red first: index has **no tab roles at all**; Storage reachable via the Settings dialog (canvas grid stays mounted behind it); Grid/Tree buttons carry `aria-pressed`
- [x] Former Files-tab test rewritten to the Tree-view toggle (alias tree + OKF preview unchanged underneath)
- [x] StorageReportCard tests green under the copy changes (19/19)
- [x] Full apps/web: 1404 jsdom, typecheck, biome, knip
- [x] Live verification: paired localhost:5199 through the D1 chip → grid, tree, and Settings all exercised (screenshots above)

## Notes

- The `HTTP 401` from the storage report endpoint visible in the screenshot is pre-existing (also present on the old Storage tab) and out of D2's scope — tracked separately.
- Chrome styling (button shapes, raw workspace id in the select) is D3 (design tokens), deliberately untouched here.

### DIFF
--- a/apps/web/src/components/PairedOriginsCard.tsx
+++ b/apps/web/src/components/PairedOriginsCard.tsx
@@ -145,7 +145,7 @@ export function PairedOriginsCard() {
           <ul className="space-y-1.5">
             {state.grants.map((grant) => (
               <li key={grant.grantId} className="flex items-center justify-between gap-2 text-xs">
-                <span className="truncate font-mono">{grant.origin}</span>
+                <span className="min-w-0 break-all font-mono">{grant.origin}</span>
                 <button
                   type="button"
                   aria-label={`Revoke ${grant.origin}`}
--- a/apps/web/src/components/StorageReportCard.tsx
+++ b/apps/web/src/components/StorageReportCard.tsx
@@ -71,7 +71,7 @@ const CATEGORIES: CategoryDescriptor[] = [
   {
     key: 'libraries',
     label: 'User libraries',
-    description: 'Excalidraw library packs you installed',
+    description: 'Library packs you installed',
     softCapBytes: USER_LIBRARIES_SOFT_CAP_BYTES,
   },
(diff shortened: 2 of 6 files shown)


kamiazya:
Fixed the Settings dialog overflow (user-reported): the operational cards were page-width components inside a `sm:max-w-md` dialog — origins in `font-mono` couldn't wrap and pushed rows past the dialog edge, and there was no vertical scroll cap. The dialog now widens to `sm:max-w-xl` when it hosts the extra sections, scrolls within `max-h-[85dvh]`, and grant origins `break-all`.

Verified live:

PR #464, opened 2026-08-08
fix(server): accept pairing and OAuth read credentials on runtime read routes

kamiazya:
## What

Fixes the Settings dialog's Storage usage card failing with **"Couldn't load storage usage (HTTP 401)"** for paired web origins, while the neighbouring Paired web apps card loaded fine.

**Root cause**: `routes/runtime.ts` mounts its own defense-in-depth middleware on `/api/runtime/*` that accepted **only the daemon token**. A paired web origin's session bearer — already validated by the global `/api/*` auth middleware (`createDaemonAuthMiddleware`) — was then rejected by this second layer, so every runtime route (storage report, status) returned 401 to paired origins. `/api/pairing/*` routes have no such second layer, which is why grant listing worked and made the failure look card-specific.

**Fix**: the per-router guard now mirrors the global middleware through the route-scope registry (single authority for public vs read vs admin):

- public routes (`ping`, `verify`) pass — now derived from the registry instead of a hardcoded path list
- the daemon token passes everything, as before
- scope-checked OAuth grants and origin-bound pairing session tokens pass routes whose registry scope is `runtime:read`
- admin routes (`shutdown`, `touch`, `logs/prune`) stay daemon-token-only: a paired web origin can inspect the daemon but never stop it or delete its logs

The pairing-origin check is extracted into `isAuthorizedPairingOrigin` (auth.ts) and shared by both layers, so the two can no longer drift.

## Verification (live dev app, paired `http://localhost:5199`)

Before: Settings > Storage usage rendered `Couldn't load storage usage (HTTP 401)`.
After (same paired session, watch-restarted daemon) the card loads — live dialog DOM:

```
Total 5.0 MiB · 11 files
Updated just now
Canvas snapshots  318.3 KiB  2 files  …  Auto-optimised 12 days ago
Versions / Uploaded files / Exports / Logs rows all populated
```

(The screenshot persistence path of the browser tooling was unavailable this run; the dialog text above is captured from the live DOM.)

## Test plan

- [x] Red first: `runtime.test.ts` — pairing bearer + paired Origin on GET `/api/runtime/storage` expected 200, got 401 pre-fix
- [x] Negative coverage: wrong origin → 401, missing Origin header → 401, pairing bearer on `POST /api/runtime/shutdown` and `POST /api/runtime/logs/prune` → 401 with no side effect
- [x] mcp-node project: 252 files, 3127 passed
- [x] `pnpm --filter @kamiazya/whiteboard-mcp typecheck`, `pnpm lint:noconsole`, `pnpm knip` clean
- [x] Live verification against the running daemon from the real paired web app

## Summary by CodeRabbit

* **New Features**
  * Added origin-bound pairing token authentication for eligible runtime read operations.
  * Added OAuth grant authentication support for scoped runtime routes.
  * Administrative operations continue to require daemon-level authorization.

* **Bug Fixes**
  * Pairing tokens are rejected when the Origin header is missing or does not match.
  * Unauthorized requests now consistently return a 401 response.


kamiazya:
Applied the CodeRabbit finding: added route-level OAuth grant coverage (runtime:read success, insufficient-scope rejection with handler-not-run assertion, admin-route rejection) in `runtime.test.ts`. 21/21 green locally.

PR #470, opened 2026-08-08
feat(web): two-finger touch pan and pinch zoom on the spatial canvas

kamiazya:
## What

User report: on a phone the canvas could not be panned or zoomed at all. Root cause: the editor root sets `touch-action: none` (disabling the browser's own touch scrolling/zooming — necessary so one-finger gestures reach the editor) but supplied **no touch navigation of its own**; pan/zoom existed only as wheel, Space-drag, and middle-button drag, none of which exist on a phone.

**Fix — the standard whiteboard convention:**

- **One finger** keeps the existing semantics: tap to select, drag a node to move it, drag empty space for marquee select, double-tap to create/edit.
- **Two fingers** navigate: dragging both pans by the centroid's movement; pinching zooms by the finger-distance ratio, anchored at the centroid so the content under your fingers stays under them.
- A second finger landing mid-gesture cancels whatever the first finger started (marquee, node move) and converts the sequence to navigation; the lone finger left after a pinch stays inert until it lifts — no accidental marquee or pan tail.

`touch-pinch.ts` is the pure per-frame math (`computePinchUpdate`: pan delta + zoom factor + anchor); `SpatialEditor` wires it through the existing pointer handlers keyed on `pointerType === 'touch'`. Mouse/pen paths are untouched.

## Verification

- Live dev app: dispatched real touch PointerEvent sequences against the running editor — pinch spread took the viewport from `scale(1) translate(0px, 0px)` to `scale(2) translate(-275px, -113px)` (anchored at the finger centroid), and a two-finger +40/+30px drag panned exactly +20/+15 canvas units at zoom 2.
- The **Cloudflare Pages preview URL** below (browser-local mode) can be opened directly on a phone to feel the real gesture.

## Test plan

- [x] Red first: pure-math unit tests (pan-only, spread ×2, pinch ×0.5, degenerate distance, and the pinned invariant that the canvas point under the centroid stays fixed when applied pan-then-zoom against the real viewport math)
- [x] Red first: real-PointerEvent browser tests — two-finger pan, pinch zoom, second-finger cancels an in-flight marquee, lone finger after pinch does not keep panning
- [x] Full web-browser project 43 files / 233 passed (all existing mouse-path suites green); jsdom 1416 passed
- [x] typecheck / lint / knip clean

## Summary by CodeRabbit

* **New Features**
  * Added two-finger touch navigation to the spatial editor, including panning and pinch-to-zoom.
  * Automatically switches from one-finger selection or manipulation to touch navigation when a second finger is detected.
  * Improved handling of touch releases and cancellations to prevent unintended editor movement.

* **Tests**
  * Added coverage for panning, zooming, gesture cancellation, and edge-case pinch behavior.

### DIFF
--- a/apps/web/src/components/spatial-editor/SpatialEditor.tsx
+++ b/apps/web/src/components/spatial-editor/SpatialEditor.tsx
@@ -486,7 +486,15 @@ export const SpatialEditor = forwardRef<SpatialEditorHandle, SpatialEditorProps>
           if (gestureState.kind !== 'idle') {
             applyResult(reduceGesture(gestureState, canvas, { type: 'pointercancel' }))
           }
-          capturePointer(root, e.pointerId)
+          // Capture BOTH fingers, not just the one that arrived second: an
+          // uncaptured first finger crossing outside the root would stop
+          // delivering its move/up events here, leaving a stale entry in
+          // touchPointsRef that would misread a later one-finger press as
+          // a pinch participant.
+          for (const pointerId of touchPointsRef.current.keys()) {
+            trySetPointerCapture(root, pointerId)
+          }
+          activePointerIdRef.current = e.pointerId
           return
         }
       }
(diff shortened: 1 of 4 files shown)


kamiazya:
Applied the CodeRabbit finding: pinch activation now captures every active touch pointer (was: only the second finger), with a browser regression test spying on setPointerCapture. 5/5 green locally.

PR #474, opened 2026-08-08
feat(canvas): JSON Canvas edge arrowheads, ends, and side pinning (J1)

kamiazya:
## What

First slice (J1) of the JSON Canvas 1.0 full-spec plan. The spec defaults every edge to a **destination arrowhead** (`toEnd: arrow`), but no layer ever rendered ends — edges were bare polylines — and `edge_patch` could not even set them.

- **canvas-render**: `ResolvedEdgeNode` carries `fromEnd/toEnd` (required — `routeEdge` resolves the spec defaults, so no scene constructor can forget). The SVG backend draws filled-triangle arrowheads in the edge's stroke color, oriented along the terminal segment. Geometry lives in one shared helper (`edge-arrows.ts`) that `sceneBounds` also walks, so a derived viewBox never clips the wings. Degenerate zero-length directions skip the arrow (total, never throws). The shape/appearance **golden is regenerated deliberately** — its edge now exercises the arrow serialization format.
- **server-core**: `edge_patch` accepts `fromEnd/toEnd` (schema-tested — the strict schema was silently locking agents out of arrowheads even though the model supported them).
- **editor**: the edge context menu gains one-tap arrow direction states (`→ / ↔ / ← / none`, current one check-marked; ends equal to the spec defaults are removed so stored documents stay canonical) and per-endpoint **side pinning** cycles (auto/top/right/bottom/left). `arrow-side-pinning` leaves `SPATIAL_EDITOR_UNSUPPORTED`.

One renderer invariant holds: editor, viewer, and headless export all pick the arrows up from the same scene builder.

## Verification (live dev app)

Real canvas with three edges: three destination arrowheads render (`<polygon … fill="#737373">`, the theme stroke color). Right-clicking an edge shows `Edit label / ✓ Arrow → / Arrow ↔ / Arrow ← / No arrows / From side: auto / To side: auto / Delete`; selecting `Arrow ↔` immediately renders the fourth polygon (both ends on that edge).

## Test plan

- [x] Red first: routeEdge threads spec-default and explicit ends (incl. the degraded missing-endpoint path); backend emits tip-first polygons with exact coordinates, stroke-color fill, and none for zero-length edges; sceneBounds includes arrow wings
- [x] Red first: `edgePatchFieldsSchema` admits ends and rejects invalid values; tool round-trip
- [x] Red first: commands — `set-edge-ends` canonical default-removal, `set-edge-side` pin/auto; real-browser context-menu flow (checkmark on the default state, `Arrow ↔` → both polygons, side pin auto→top)
- [x] canvas-render 246 (golden updated deliberately), server-core 226, apps/web jsdom 1427 + browser 243, full `pnpm test` (the two knowns: stale-dist artifact smoke green after `pnpm build`; /pair test passes in isolation)
- [x] typecheck / lint / knip clean

## Summary by CodeRabbit

* **New Features**
  * Added configurable arrowheads on both ends of spatial editor edges.
  * Added edge options for arrow direction and endpoint side pinning.
  * Added accessible option pickers that keep menus open while selecting.
  * Edge arrows now render correctly, follow edge colors, and remain within scene bounds.
  * Edge editing supports saving and updating arrow and pinning settings.

* **Bug Fixes**
  * Prevented arrowheads from being clipped or rendered on zero-length edges.

* **Tests**
  * Expanded coverage for edge options, rendering, validation, and reset behavior.

### DIFF
--- a/apps/web/src/components/spatial-editor/ContextMenu.tsx
+++ b/apps/web/src/components/spatial-editor/ContextMenu.tsx
@@ -19,12 +19,14 @@
  * discrete pointerdown loses the focus fight with mousedown's default
  * action, while click fires after those defaults.
  */
-import { useEffect, useRef } from 'react'
+import { type ReactNode, useEffect, useRef } from 'react'
 import { cn } from '@/lib/utils'
 
 export interface ContextMenuActionItem {
   readonly kind?: 'action'
   readonly label: string
+  /** Leading icon (a lucide element); decorative — the label carries the name. */
+  readonly icon?: ReactNode
   readonly onSelect: () => void
   /** Visually separates destructive entries (Delete). */
   readonly danger?: boolean
@@ -33,6 +35,8 @@ export interface ContextMenuActionItem {
 export interface ContextMenuOption {
   /** Visible content of the option button (often a glyph). */
   readonly label: string
+  /** Icon rendered INSTEAD of the label text when provided (decorative). */
+  readonly icon?: ReactNode
   /** Accessible name when the visible label is a glyph; defaults to label. */
   readonly ariaLabel?: string
   readonly selected: boolean
@@ -123,7 +127,13 @@ export function ContextMenu({ x, y, items, onClose }: ContextMenuProps) {
                   // reopen for every adjustment.
                   onClick={option.onSelect}
                 >
-                  {option.label}
+                  {option.icon !== undefined ? (
+                    <span aria-hidden="true" className="[&>svg]:size-3.5">
+                      {option.icon}
+                    </span>
+                  ) : (
+                    option.label
+                  )}
                 </button>
               ))}
             </span>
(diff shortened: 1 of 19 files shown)


kamiazya:
Reworked the property UX per user feedback: cycling items are gone. The context menu now uses **inline option rows** (segmented pickers — one tap to any value, and the menu stays open so arrows + both sides adjust in one visit) with **section separators** grouping action / properties / destructive. Live-verified: `Edit label | Arrows [→ ↔ ← −] / From side / To side | Delete`, pick applies immediately with the menu open. Browser tests updated (3× green).

kamiazya:
Applied the CodeRabbit finding: stroke-less arrowheads now emit `fill="none"` (matching the invisible polyline) instead of the SVG default black fill — no more detached black arrows on appearance-free edges. Byte-pinned in the appearance-free arrow tests; canvas-render 246 green.

PR #476, opened 2026-08-08
feat: JSON Canvas link nodes — Add link dialog, follow, and Edit URL

kamiazya:
## Summary

J3 of the JSON Canvas 1.0 full-spec plan: link nodes become creatable and followable in the spatial editor. (Rendering already existed — canvas-render draws link nodes as chrome + URL label.)

- **Palette "Add link"**: a new palette button opens a URL dialog; a valid submit creates a `type: 'link'` node at the viewport center (free-spot cascade, shorter default box since the node is label-only chrome) and selects it.
- **Follow the reference**: double-click a link node opens its URL in a new tab (`noopener,noreferrer`), mirroring the text node's double-press-edits rule; the context menu adds **Open link** and **Edit URL** alongside the shared Color row and Delete.
- **`set-node-url` command**: rewrites a link node's destination; a non-link target or missing id is a no-op. Canonical storage matches the other `set-*` commands.
- **One validation authority**: the dialog validates with the same `z.url()` rule the canvas-model schema enforces, so an accepted URL can never fail schema validation downstream. Invalid input keeps OK disabled and submit inert; Escape dismisses.

## Visual repro

Add link dialog → created link node (URL label, selected):

Link node context menu (Open link / Edit URL / Color / Delete), clamped inside near the edge:

Live-verified in Chrome: double-clicking the created node opened https://jsoncanvas.org/spec/1.0/ in a new tab.

## Notes for review

- A live-caught bug during this slice: the dialog initially rendered inside the pan/zoom transform, so its percentage position resolved against the transformed container and it opened far off-center. It now renders at the root overlay layer (same layer as the context menu), with positioning inline so browser-mode tests (which run without the app stylesheet) exercise the same geometry.

## Test plan

- [x] `commands.test.ts`: set-node-url applies / no-ops on non-link or missing targets (22 passed)
- [x] `link-node.browser.test.tsx`: create via dialog, invalid-URL blocked + Escape dismiss, double-click opens URL (window.open spy, noopener), Edit URL prefilled + rewrite (4 passed)
- [x] Full suite green (one unrelated App.test.tsx lazy-route flake under full-suite load; passes 43/43 in isolation); `pnpm -r typecheck`, `pnpm lint`, `pnpm knip` clean
- [x] Live verification in Chrome (screenshots above)

### DIFF
--- a/apps/web/src/components/spatial-editor/LinkUrlDialog.tsx
+++ b/apps/web/src/components/spatial-editor/LinkUrlDialog.tsx
@@ -1,0 +1,90 @@
+/**
+ * URL entry surface for link nodes — used both by the palette's "Add link"
+ * (create) and the context menu's "Edit URL" (rewrite).
+ *
+ * Validation delegates to the same rule the canvas-model schema enforces
+ * (`z.url()`), so a URL this dialog accepts can never fail schema
+ * validation downstream — one authority, no drift.
+ *
+ * Marked `data-editor-overlay` so the canvas root's gesture handlers ignore
+ * presses inside it. Positioning is inline for the same reason as the
+ * context menu: it must behave identically where the app stylesheet is
+ * absent (browser-mode component tests).
+ */
+import { useState } from 'react'
+import { z } from 'zod'
+
+const urlSchema = z.url()
+
+export interface LinkUrlDialogProps {
+  readonly title: string
+  readonly initialUrl?: string
+  readonly onSubmit: (url: string) => void
+  readonly onCancel: () => void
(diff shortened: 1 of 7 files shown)


kamiazya:
Security follow-up (0522e4e5): `z.url()` accepts any parseable URL — including `javascript:` — and `window.open` on one executes in the app's origin. Since canvases arrive via sync/import (not only via our own dialog), the http/https allowlist now holds at the open sink (`openLinkNode`) as well as in the dialog validation. Regression test pins both directions.

PR #484, opened 2026-08-08
fix: allow https iframes in the Pages CSP

kamiazya:
## Summary

User report: link-node iframe embeds render as Chrome's このコンテンツはブロックされました page on the deployed app (mobile screenshot). Root cause: `public/_headers` declares a CSP with **no `frame-src`**, so child iframes fall back to `default-src 'self'` and every external embed is blocked by our own policy. Local dev serves no `_headers`, which is why only production surfaced it.

Fix: add `frame-src https:` to the global CSP. `https:` is sufficient — `http:` iframes on the https origin are mixed content and are blocked by the browser before CSP is consulted, and the editor's followable-url gate already restricts embeds to http/https.

## Test plan

- [x] `public-headers.test.ts` gains a red-first assertion that the CSP contains `frame-src https:` (4 passed)
- [ ] Verify on the Cloudflare Pages preview deployment that a link-node embed actually renders before merging (local cannot reproduce — headers only apply on Pages)

### DIFF
--- a/apps/web/public-headers.test.ts
+++ b/apps/web/public-headers.test.ts
@@ -50,4 +50,16 @@ describe('public/_headers', () => {
     const csp = globalHeaders?.get('Content-Security-Policy') ?? ''
     expect(csp).toContain("worker-src 'self'")
   })
+
+  it('permits https frame-src so link-node iframe embeds are not blocked by our own CSP', () => {
+    // Without an explicit frame-src, child iframes fall back to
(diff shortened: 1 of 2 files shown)


kamiazya:
Preview verification done on https://fix-embed-frame-src.kamiazya-whiteboard.pages.dev — created a link node, activated the embed facade, and the sandboxed iframe now renders jsoncanvas.org's actual content (no `securitypolicyviolation` event fired; the deployed CSP carries `frame-src https:`):

PR #499, opened 2026-08-09
feat: node lock stored in the Loro layer, honored by the editor and by agents

kamiazya:
## Summary

Lock as the user framed it: **kept in the Loro layer, never emitted on export**, because it is an editor experience rather than canvas content.

**canvas-workspace** — a `nodeLocks` sidecar map keyed by node id (`readNodeLocks` / `setNodeLock`). Since `readSpatialCanvas` reads only nodes/edges, and every export, render, and version path goes through it, the lock cannot reach a JSON Canvas file, an OKF export, or an SVG — no stripping step needed anywhere. It also keeps the SoT unambiguous: nodes/edges are the canvas, the sidecar is editor state.

**Decisions implemented** (yours, 2026-08-09):
1. **Agents respect the lock, but can still lock/unlock.** `node_patch` refuses a locked node with `NodeLockedError` *before* any write; the new `node_lock` tool is the one mutation a locked node still accepts, so an agent can lift its own mistake without a human at the keyboard.
2. **Copies do not inherit the lock** — the sidecar is keyed by the source id and paste remints.
3. **Deleting a node cascades its entry away**, so no orphans accumulate and a reminted id can never inherit a stranger's lock.

**Editor** — locked nodes are invisible to pointer hit-testing, marquee, and select-all, so they cannot be selected, moved, resized, or deleted. Right-click still finds them and offers exactly one action, **Unlock**. `Cmd+Shift+L` toggles the selection. Without the host seam the whole affordance hides and nothing is blocked.

## Reload persistence

An earlier revision of this PR disclosed that a lock did not reappear after a reload in browser-local mode. The cause was in the composition, not in any one layer: the session published its lock set only on a toggle, so a lock already in the doc at hydrate was never announced, and the hook subscribed without seeding from the session. Both are fixed — hydration and the initial subscribe each publish the persisted set.

Verified in a real browser against IndexedDB: lock a node, reload, right-click — the menu offers **Unlock**; unlock, reload, right-click — the full menu with **Lock** is back. `BrowserLocalCanvasPage.node-lock.browser.test.tsx` is the regression guard for both directions, and it is mutation-checked (removing the hydrate notification fails it with an empty locked set). Daemon-mode persistence remains unverified.

## Visual repro

After locking a node and reloading the page, right-clicking it still offers only **Unlock** — the lock survived the reload.

## Test plan

- [x] `loro-bridge.test.ts` (red-first, 6 cases): set/clear/read, invisible to `readSpatialCanvas`, survives a full `writeSpatialCanvas` resync, delete cascade (helper + batch), no-op write, and snapshot+update replay survival — 50 passed
- [x] `node-lock.test.ts` + `node-patch.test.ts` (server-core): lock/unlock persists, ghost id rejected, idempotent, `node_patch` refuses a locked node and writes nothing, patches again once unlocked — 342 passed across the package suites
- [x] `node-lock.browser.test.tsx` (real Chromium, 8): not selectable, not draggable, skipped by select-all and marquee, Unlock-only menu, Lock entry, `Cmd+Shift+L`, inert without the seam, no canvas mutation
- [x] `canvas-sync-session.test.ts`: hydration reports the persisted set; `setNodeLock` pushes to peers and notifies without publishing a canvas
- [x] `BrowserLocalCanvasPage.node-lock.browser.test.tsx` (real IndexedDB, 2): a lock and an unlock each survive a page remount — mutation-checked against the hydrate notification
- [x] Manual: locked and unlocked both survive a real browser reload (screenshot below)
- [x] Full: web-browser 323/61, jsdom 1478/123, package suites 342/39; `pnpm -r typecheck` 0; knip clean

## Summary by CodeRabbit

* **New Features**
  * Added node locking to prevent selection, movement, resizing, deletion, connections, and edits.
  * Lock or unlock nodes from the context menu or with Cmd/Ctrl+Shift+L.
  * Node lock states persist and synchronize across sessions.
  * Added a server tool for locking and unlocking nodes.
* **Bug Fixes**
  * Prevented modifications to locked nodes through editing tools and server-side updates.
  * Ensured deleted nodes no longer retain lock states.
* **Tests**
  * Added comprehensive coverage for locking, synchronization, persistence, and protected actions.


kamiazya:
Triaged all six comments against the code. Five were real and are fixed on this branch; one is real but not this PR's to fix.

**Fixed**

- `SpatialEditor` — confirmed on all three counts, and all three were red-first. A lock arriving from a peer or an agent while the node was already selected or mid-drag left every selection-driven path open; rather than guarding each command site, the selection is now pruned when the locked set changes, which closes nudge/delete/resize/z-order/colour/cut at one point (a locked primary promotes a surviving extra instead of deselecting everything, and a locked in-flight move/resize gesture is cancelled). Group frames no longer carry locked members. The Tab-reachable connect-target overlay now uses `selectableBoxes`, so keyboard and pointer agree. (e918080d)
- `useCanvasSync` — confirmed; teardown reset canvas and restore state but not locks. (5c9603ce)
- `writeSpatialCanvas` orphan lock — confirmed, and it contradicted this PR's own stated invariant: the resync is the second node-removal path, so it owes the same cascade `deleteSpatialNode` does. (666a7264)
- `NodeLockedError` export — confirmed. (150dd0f2)

**Not fixed here**

- Serializing lock and patch persistence. The race is real, but it is not `node_lock` vs `node_patch` — `saveDocSnapshot` passes a frontier that no store implementation compares, so the same lost update sits under `node_patch` vs `node_patch`, `edge_patch`, `body_patch`, `facet_set`, and `canvas_import_okf` equally. A fix belongs in the `CanvasDocStore` contract (a compare-and-swap on the stored frontier, plus a retry in the shared tool path) and both implementations, not in these two tools. Tracked in the local backlog with the required tests; adding a CAS to the store contract inside a lock PR would be a bigger and less reviewable change than the feature it rides on.

PR #506, opened 2026-08-09
fix(mcp): serialize mutating tool calls per canvas so neither loses an update

kamiazya:
Second item from the recommended order (backlog canvas `canvas-doc-store-lost-update-race`), found while triaging CodeRabbit on #499.

## The bug, demonstrated rather than argued

Every mutating MCP tool is a load-modify-save against `canvasDocStore`, and `saveSnapshot` writes **unconditionally** — it carries the new frontier, but no implementation compares it against the stored one. Two calls that load the same base before either saves silently drop one of the changes.

The first test in this PR is the demonstration: two concurrent `node_patch` calls to different nodes of the same canvas, on the real tool, and exactly **one** of the two survives. It is written as an assertion on the hazard, so if a future change ever makes the unserialized path safe on its own, that test failing is the signal this lock can go.

This mattered more once node lock shipped: a patch that loaded before a lock landed can apply to a locked node *and* erase the lock.

## Why an in-process queue rather than a compare-and-swap

I filed this originally proposing CAS in the `CanvasDocStore` contract. Reading the code changed my mind about the right size for **today**:

- The daemon is a single process, and `workspace-lock.ts` already says so in its own header — it exists for exactly this class of problem and already carries the reentrancy handling (`AsyncLocalStorage`) that a naive queue gets wrong.
- The shared layer cannot host it: `node:async_hooks` is banned there by the architecture rules, and serialization is a platform concern that belongs in a composition root.
- CAS is a port-contract change across two implementations and nine call sites, and the retry it needs means restructuring each tool into a re-runnable mutate callback.

So: reuse the existing machinery, keyed per canvas document. The commit and the module comment both state plainly that a multi-instance deployment (the Cloudflare direction) cannot rely on this and needs the CAS — with the specific shape written down, so the next person does not have to re-derive it.

## Details worth reviewing

- **Read-only tools are deliberately NOT wrapped.** Serializing a render or digest behind an unrelated patch buys no correctness and turns every agent write into a stall.
- **The lock wraps `execute()`, not the registration.** `opencanvas-tools.ts` documents at length why registration is not routed through a generic wrapper (a generic body makes `ToolHandlerReturn<O>` unresolvable and silently degrades the type check into a cast). Wrapping the inner promise leaves that binding concrete.
- **The key is namespaced** (`canvas-doc:<id>`) because workspace and canvas ids are separate spaces; an unlucky collision would silently share a queue.

## Test plan

- [x] `canvas-doc-write-lock.test.ts` (5): the lost update on real tools, both surviving once serialized, different canvases not blocking each other, and the queue still draining after a holder throws
- [x] Wiring conformance guard: every mutating tool wrapped, every read-only tool not — **mutation-checked** (unwrapping `nodePatch` fails it by name)
- [x] `pnpm smoke:e2e` green (18 tools, lock/patch/export paths unchanged)
- [x] Full `pnpm test` 583/583 files, 6187 passed, 0 failed; `pnpm -r typecheck` 0; `pnpm build` 0

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when multiple canvas edits occur at the same time.
  * Prevented concurrent changes to the same canvas from overwriting one another.
  * Ensured canvas updates continue processing even if an earlier operation fails.
  * Kept independent canvas documents from blocking each other.

### DIFF
--- a/packages/mcp-server/src/server/mcp/opencanvas-tools.ts
+++ b/packages/mcp-server/src/server/mcp/opencanvas-tools.ts
@@ -18,6 +18,7 @@ import {
 import type { McpServer } from '@modelcontextprotocol/server'
 import type { z } from 'zod'
 import { getLogger } from '../log.js'
+import { withCanvasDocWriteLock } from '../store/workspace-lock.js'
 import { registerToolWithAnnotations, structuredJsonResult } from './tool-support.js'
 
 // server-core is a shared layer and cannot depend on this composition
@@ -42,6 +43,18 @@ setServerCoreLogSink((record) => {
 export function registerOpenCanvasTools(server: McpServer, deps: ServerDeps): void {
   const { tools } = createServer(deps)
 
+  // Every MUTATING tool below runs inside withCanvasDocWriteLock, keyed on
+  // the canvas it targets. Each is a load-modify-save against
+  // canvasDocStore, whose saveSnapshot writes unconditionally, so two calls
+  // that load the same base before either saves silently drop one of the
+  // changes (see canvas-doc-write-lock.test.ts, which demonstrates the loss
+  // on the real tools). Read-only tools are deliberately NOT wrapped:
+  // serializing reads behind writes would make a render or digest wait on
+  // an unrelated patch for no correctness gain.
+  //
+  // The lock wraps the execute() call rather than the registration, so the
+  // concrete outputSchema/O binding described below is untouched.
+
   // Each call below is a direct registerToolWithAnnotations invocation (not
   // routed through a shared generic wrapper) so outputSchema and O are
   // concrete at every call site: TypeScript checks this handler's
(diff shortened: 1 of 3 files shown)


kamiazya:
Both findings were valid and both are fixed in ca95cf15.

**Force both patch calls to load the same snapshot** — right, and worse than a generic miss: this is the exact shape I rejected in #504 a few hours ago as "a test that passes whenever the collision it meant to create never happened," and then wrote here anyway. A barrier now holds the first two canvas loads until both have arrived, so the shared base is constructed rather than left to the scheduler.

One thing the fix surfaced that is worth recording: the barrier **cannot** be used in the serialized case. Once the calls are serialized the second never loads until the first finishes, so waiting for both to arrive deadlocks — my first attempt did precisely that and hung for 10s. That case now asserts the store traffic instead, which expresses the property more directly anyway: the second load must land after the first save. The failure message prints the sequence, so a regression reads as `load,load,save,save` (the lost update) against `load,save,load,save` (the fix).

**Test MCP handler wiring at runtime** — also right. The source-text check would pass on a wrapping kept syntactically but bypassed semantically, and would break on reformatting. It is replaced by a test that runs the registered handlers: two mutating calls must serialize, and a read-only handler must complete while a write is held mid-flight (that second half is what keeps a future change from "fixing" this by queueing reads behind writes too).

Both are mutation-checked — unwrapping `node_patch` fails the runtime test with the interleaved sequence in the message.

PR #508, opened 2026-08-09
feat(web): snap a dragged node to neighbour edges and the grid

kamiazya:
Dragging a node now snaps to nearby content and to a background grid, with the guide line that justifies each snap drawn while the gesture is in flight. Cmd/Ctrl suspends it for one gesture (the Figma / Excalidraw / tldraw convention).

Both snap targets are live: a neighbour's leading edge, centre and trailing edge on either axis, plus a grid. The two axes resolve independently, so a box can land on a neighbour's edge horizontally while taking the grid vertically.

## Visual repro

The dragged node's left edge stopped 3px short of the node above it; the snap closes the gap and the guide shows why.

## Decisions worth reviewing

**The pointer is nudged, not the emitted command.** The reducer derives both the preview and the commit from the same `point`, so adjusting only one of them would draw the box in one place and land it in another.

**The grid pitch (20) is deliberately wider than twice the threshold (6).** At or below that, every lattice line is reachable from everywhere — silent rounding rather than snapping — and it would out-pull the neighbour edges the user was aiming at.

**Anything travelling with the drag is excluded from the candidate set**: a multi-selection member, or a frame's contained members. Left in, a carried node attracts its own carrier and pegs the gesture at a fixed offset.

**Snapping is move-only.** A resize would have to snap the dragged edge, a different candidate set — deferred rather than approximated, since an approximate resize snap fights the handle.

## Tests

`snap.ts` has 16 unit tests (candidate selection, threshold, axis independence, totality on non-finite input). `snap.browser.test.tsx` pins the wiring in a real browser: the landing position, the guide appearing and retiring with the gesture, Cmd suspension, and the carried-member exclusion. The last two passed before the wiring existed, so both were mutation-checked — removing either guard fails its test.

Three existing tests asserted 1:1 pointer tracking, which snapping deliberately breaks. The preview-tracking one now holds Cmd, since raw tracking is exactly what Cmd preserves; the two commit-shape ones move to a drop point no candidate is near, so they stay on the default path.

## Also here

A separate commit narrows a canvas-model property that a rare seed failed on `{ __proto__: {} }`. The finding is real but the schema is right: writing `__proto__` onto the result sets the prototype rather than creating an own key, so it comes back dropped — and that drop is what stops a hostile facets payload reaching a downstream spread as a prototype. The property is narrowed and the drop pinned as deliberate behaviour, rather than the schema widened to carry it.

## Verification

`pnpm test` 585/585 files green (6213 passed), `pnpm -r typecheck` clean, `pnpm knip` clean. Manually verified in Chrome against the running app: a 3px near-miss snapped to the neighbour edge with the guide at the right position, and a separate real-pointer drag landed on the grid line as predicted.

## Summary by CodeRabbit

- **New Features**
  - Added snapping for dragged nodes and groups to nearby edges, centers, and a 20-unit grid.
  - Added alignment guides during dragging, with consistent preview and final placement.
  - Snapping now works independently along each axis and can be suspended with Cmd/Ctrl.

- **Bug Fixes**
  - Improved handling of special `__proto__` data keys to prevent prototype pollution while preserving valid data.

- **Tests**
  - Added coverage for snapping behavior, guides, modifier-key controls, grouping, and edge cases.

### DIFF
--- a/apps/web/src/components/spatial-editor/SpatialEditor.browser.test.tsx
+++ b/apps/web/src/components/spatial-editor/SpatialEditor.browser.test.tsx
@@ -119,20 +119,23 @@ describe('SpatialEditor (browser)', () => {
         button: 0,
       }),
     )
+    // +29/+29 lands "a" at (49,49): 9 from the nearest grid line and well
+    // clear of every line node "b" offers, so this stays a plain drag on the
+    // default path rather than a snap (snap.browser.test.tsx covers those).
     await editor
       .element()
       .dispatchEvent(
-        new PointerEvent('pointermove', { bubbles: true, clientX: 70, clientY: 55, pointerId: 1 }),
+        new PointerEvent('pointermove', { bubbles: true, clientX: 69, clientY: 69, pointerId: 1 }),
       )
     await editor
       .element()
       .dispatchEvent(
-        new PointerEvent('pointerup', { bubbles: true, clientX: 70, clientY: 55, pointerId: 1 }),
+        new PointerEvent('pointerup', { bubbles: true, clientX: 69, clientY: 69, pointerId: 1 }),
       )
 
     expect(onChange).toHaveBeenCalledTimes(1)
     const [next, command] = onChange.mock.calls[0] as [SpatialCanvas, unknown]
-    expect(command).toEqual({ kind: 'move-node', id: 'a', x: 50, y: 35 })
+    expect(command).toEqual({ kind: 'move-node', id: 'a', x: 49, y: 49 })
     expect(next.nodes[1]).toEqual(canvasValue.nodes[1])
     // input untouched
     expect(canvasValue.nodes[0]).toEqual({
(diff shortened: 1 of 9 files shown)


kamiazya:
All three findings applied — thanks, the third one was a genuinely wrong claim on my part.

**Locked group member (4af9f424)** — valid, fixed. The commit path already filters `!isLocked`, so a locked contained member never travels with its frame; it is stationary content and a legitimate alignment target. Excluding it silently discarded one. The exclusion now mirrors the commit path's own filter so the two cannot drift. Regression test added and it was red before the fix: `DOES snap a group frame to a LOCKED member, which stays put`.

**Ctrl coverage (4af9f424)** — added. The two modifiers reach the handler through different event fields, so testing one left the other free to regress. Mutation-checked: narrowing the guard to `metaKey` fails the Ctrl case.

**`__proto__` rationale (de054027)** — you were right, and both halves of my comment were wrong. I verified each by running it rather than reasoning:

- Native spread does NOT invoke the setter — `{...hostile}` yields an own `__proto__` data property with the prototype intact. Naming it as the unsafe sink was simply incorrect. The sinks that DO pollute are `Object.assign` and a hand-written `for (k) out[k] = ...` merge.
- The mechanism was also wrong. Zod does not accidentally write the prototype; it deliberately **skips** the key. The parsed object has `keys: ['keep']`, no own `__proto__`, prototype still `Object.prototype`, and no global pollution.

Both comments now say that, and the test asserts the point as behaviour rather than prose: the parsed value survives `Object.assign` without polluting.

PR #510, opened 2026-08-09
feat(web): make canvas embeds and image nodes work in daemon mode

kamiazya:
Canvas embeds (J5a) and image nodes (J5b) only ever worked in browser-local mode. `DaemonCanvasPage` passed none of the four file seams to `SpatialEditor`, so in daemon mode a canvas-reference node rendered as a bare card that never expanded, and pasting or dropping an image did nothing at all.

The daemon side was already complete. `PUT`/`GET /api/canvas/:workspaceId/:slug/file/:fileId` and the snapshot route have been there all along — MCP's `load_image` uses them. Nothing but the web wiring was missing.

## Why nothing failed

Each page's tests only ever exercised its own mode. That is the whole reason this survived, and it is why the anti-recurrence guard here is a **conformance test across both pages** rather than another per-page test. Mutation-checked: dropping the daemon page's seam spread fails it.

## Commits

1. **`refactor(web)`** — extract the seams behind `CanvasFileAdapter` (behaviour-preserving). The caching rules are why this is a shared hook and not a second copy: staleness stamps, the same-map-instance guard that stops failed image reads spinning forever, and object-URL revocation on unmount.
2. **`fix(repo)`** — the pre-commit gate was not a gate (see below).
3. **`feat(web)`** — the daemon adapter, which is now just four functions.

## The reference convention

`asset:<id>` is deliberately the **same** in both modes. A canvas has to mean the same thing by the same `file` value wherever it is opened, or moving content between modes would reinterpret every image node as a canvas reference. The colon keeps it unambiguous: a daemon slug and a browser-local canvas id both match `/^[a-zA-Z0-9_-]+$/`, so neither can collide. The daemon's file route validates `fileId` against that same pattern, so the adapter strips the prefix before it travels in a path.

A failed upload returns `undefined` rather than the reference — reporting success would leave a node pointing at bytes the daemon never stored.

## The pre-commit gate was not a gate

Found while committing the first change: `.git/hooks/pre-commit` held two concatenated scripts, lefthook's followed by a block appended by code-review-graph ending in `|| true`. lefthook's exit status was never checked, so the script always exited 0 and git committed anyway. secretlint reported its failure correctly on the way past, which is exactly what let this survive — the output looks identical to a gate that works.

Verified by breaking it: a staged file containing an absolute home-dir path committed cleanly. After the fix the same commit exits 1 and leaves HEAD alone.

`verify-git-hooks.mjs` detects the shape (executable commands after lefthook's invocation; comments cannot change an exit status, so they are ignored) and `--fix` hoists the block ahead of it. Wired into pre-push, because the question it answers is "were the gates actually enforced on the way here?". It targets the general shape, not this one tool — several editor and code-intelligence integrations install by appending.

`CONTRIBUTING.md`'s hook section was also describing a pre-commit that does not exist: secretlint was missing entirely and the hook was called non-blocking. Corrected to what runs.

## Verification

`pnpm test` 589/589 files green (6242 passed), `pnpm -r typecheck` clean, `pnpm knip` clean.

**Manual daemon-mode verification is still outstanding** — the adapter is unit-tested against a fake fetch, and the route contract was read from the daemon source, but I have not yet driven a real paired daemon session end-to-end. Holding as draft until that is done.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -62,8 +62,10 @@ This is not yet enforced as a hard CI gate because the existing codebase still h
 
 [Lefthook](https://lefthook.dev) installs git hooks automatically on `pnpm install` (via the `prepare` script):
 
-- **pre-commit** (fast, non-blocking): formats the staged files with Biome and re-stages them. It deliberately does NOT run lint/typecheck/tests, so it never slows the many automated commits the dev flow makes.
-- **pre-push** (the gate): runs `pnpm -r typecheck`, `pnpm test --project mcp-node`, and `pnpm lint:noconsole` in parallel — mirroring CI so a broken build, failing test, or stray `console.*` in server code is caught before it leaves your machine.
+- **pre-commit** (fast): formats the staged files with Biome and re-stages them, then runs **secretlint**, which BLOCKS the commit on a secret or an absolute home-dir path. It deliberately does NOT run lint/typecheck/tests, so it never slows the many automated commits the dev flow makes.
+- **pre-push** (the gate): runs `pnpm -r typecheck`, `pnpm test --project mcp-node`, `pnpm lint:noconsole`, `pnpm lint`, and `pnpm verify:git-hooks` in parallel — mirroring CI so a broken build, failing test, or stray `console.*` in server code is caught before it leaves your machine.
+
+`pnpm verify:git-hooks` checks the hooks themselves rather than your code. A tool that appends its own block to `.git/hooks/<name>` — several editor and code-intelligence integrations install that way — makes *its* exit status the script's, which silently reduces every lefthook command above to advisory while still printing failures as if they blocked. Run `pnpm verify:git-hooks --fix` to move an appended block ahead of lefthook's invocation.
 
 Hooks are a local safety net, **not** a replacement for CI (the authoritative gate). Bypass a run when you must with `LEFTHOOK=0 git …` or `git push --no-verify` (e.g. a docs-only push). If the hooks didn't install, run `pnpm lefthook install`.
 
--- a/apps/web/src/hooks/use-canvas-file-seams.test.tsx
+++ b/apps/web/src/hooks/use-canvas-file-seams.test.tsx
@@ -1,0 +1,149 @@
+/**
+ * The backend-agnostic half of the editor's file seams. This logic used to
+ * live inline in BrowserLocalCanvasPage, which is why the daemon page shipped
+ * without any of it; the caching rules below (staleness stamps, the
+ * same-instance guard, URL revocation) are subtle enough that a second
+ * hand-written copy is exactly what should not happen.
+ */
+import type { SpatialCanvas } from '@kamiazya/whiteboard-canvas-model'
+import { renderHook, waitFor } from '@testing-library/react'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { type CanvasFileAdapter, useCanvasFileSeams } from './use-canvas-file-seams.js'
+
+const canvasWith = (...files: string[]): SpatialCanvas => ({
+  nodes: files.map((file, i) => ({
+    id: `n${i}`,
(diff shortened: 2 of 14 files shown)


kamiazya:
## Chrome verification against a real daemon

Both seams confirmed live in daemon mode (dev server + `pnpm mcp:http:dev`, paired via `#wb=`):

**Canvas embed (J5a)** — adding a canvas-reference node and reloading fires `GET http://127.0.0.1:3099/api/canvas/<ws>/untitled/snapshot` → **200**. Before this PR the daemon page made no such request at all, because it passed no `resolveFileCanvas`.

**Image node (J5b)** — pasting a PNG uploads it to the daemon and reads it back: `GET .../file/19b455dc-…` → **200**, and server-side confirmation that the stored file is **70 bytes, `content-type: image/png`** — the exact pasted image. The editor then emits `<image href="blob:…">`.

## A pre-existing defect this surfaced, NOT caused by this PR

Image nodes **do not actually display** — the `<image>` element's object URL is already dead when the browser tries to load it (`fetch(href)` → `TypeError: Failed to fetch`), so it renders the broken-image placeholder.

This reproduces in **both** modes and, critically, on **`main` at e07c2e5c** before any of this work — two image nodes, both decode-error. The extraction is behaviour-preserving, and that faithfully includes this bug.

It is **not** root-caused yet, and I would rather say so than guess: the only `revokeObjectURL` in the seam runs on unmount, and no unmount is expected between creation and paint. The obvious candidate (a Chrome IndexedDB-backed Blob going invalid when its connection closes) would explain browser-local but not daemon mode, whose Blob comes from `Response.blob()`. No single explanation covers both yet.

Filed as its own issue canvas (`image-node-blob-url-dead-on-render`), with the reproduction and the first debugging step.

## Overlap with #509

`apps/web/public/_headers` currently sets `img-src 'self' data:` — **no `blob:`**. So on deployed builds every image node is CSP-blocked regardless of the above; only the dev server (which serves no `_headers`) escapes it. #509 (`fix-csp-blob-img`) is fixing exactly that, so **image nodes shipping visibly depends on it**. Write scopes are disjoint — #509 touches `_headers`, `CanvasThumb`, `VersionThumbnail`; this PR touches `pages/`, `hooks/`, `lib/`.

kamiazya:
## Correction: there is no separate image-display bug

My earlier comment on this PR claimed image nodes never display and that it was a pre-existing defect. **That was wrong on the cause, and the "bug" does not exist.** Images render correctly in both modes on current `main`.

What actually happened:

1. The broken-image rendering was **CSP** — `img-src 'self' data:` with no `blob:`. I had identified that directive as a deployed-only problem because I assumed the dev server serves no CSP. **It does**: `http://localhost:5173` sends the same policy. So the symptom was the CSP issue all along, on localhost too.
2. #509 landed `img-src 'self' data: blob:` shortly before this PR merged. The page I was testing had been loaded under the *old* header, and CSP is fixed per-document at load — so it kept failing until a reload.
3. My supporting evidence was also invalid: I read "the object URL is dead" from `fetch(href)` failing inside the browser-automation evaluation context. A control with a plain in-memory Blob fails there too, so that context simply cannot resolve `blob:` URLs. The instrumentation did show one thing correctly — `revokeObjectURL` was never called — which should have made me doubt the revocation story sooner than it did.

I also chased an IndexedDB-blob-dies-on-close hypothesis and wrote tests for it. Those tests **passed**, which was the signal the hypothesis was wrong; they have been discarded rather than kept, since they were justified by a defect that does not exist.

Verified after reload, judged by rendering alone (no probing):

- **browser-local**: a pasted red/green PNG renders as red/green.
- **daemon**: a pasted blue/yellow PNG renders as blue/yellow, alongside the canvas-reference node.

So this PR's feature is confirmed working end to end, and both issue canvases I filed (`image-node-blob-url-dead-on-render`, `csp-blocks-blob-images-on-deployed-builds`) have been deleted as resolved/incorrect.

PR #514, opened 2026-08-09
feat: sync over SSE when the page cannot open a WebSocket to the daemon

kamiazya:
The hosted app has no working sync transport to a local daemon: a page served over https cannot open a `ws://` socket to loopback. Mixed content blocks the upgrade before auth is attempted, and the loopback exemption that lets plain `http://` fetches through does **not** extend to WebSocket. The canvas therefore showed "Sync off" and no content on the deployed origin while the same build synced fine from an http dev origin.

Measured on one page, everything else equal:

| From `https://…pages.dev` | Result |
| :-- | :-- |
| `fetch('http://127.0.0.1:3099/api/runtime/ping')` | 200 |
| `new WebSocket('ws://127.0.0.1:3099/ws/…')` | ERROR |

## Why SSE and not the alternatives

WebRTC DataChannel and WebTransport with `serverCertificateHashes` both dodge mixed content — DTLS is mandatory in WebRTC, and WebTransport authenticates against a certificate hash instead of the Web PKI. Neither was adopted:

- Chrome 142+ gates a public origin reaching loopback behind a **Local Network Access permission**, which none of these avoid. The constraint that actually bites is unchanged by the transport.
- Both need substantial daemon-side machinery for a same-machine link (a WebRTC stack; HTTP/3 plus 14-day certificate rotation).
- The future hosted/Durable-Object leg is ordinary `wss://` between two https origins and has no problem at all. Only the loopback leg is broken, so it is the wrong place to spend the largest dependency.

A publicly-trusted certificate for a domain resolving to 127.0.0.1 (the Plex `*.plex.direct` pattern) was considered and rejected: it ships a private key to every user's machine and needs DNS, so it breaks offline use.

## What this adds

**Server.** `GET /api/sync/stream` carries incremental Loro updates (base64 — SSE is a text protocol) and server text messages; `POST /api/sync/subscribe` adjusts which documents a stream receives, because SSE itself is one-way; `POST /api/sync/message` takes the client→server messages a WebSocket sends as text frames. Both broadcast paths in `ws.ts` now fan out to SSE subscribers, so an MCP tool edit and a WebSocket peer edit are indistinguishable to a subscriber.

The stream id is minted by the daemon and announced as the stream's first frame, never chosen by the caller — a client-chosen key into a server-side registry would let one client name another's stream. Binding a stream to its opener's credential instead was considered and rejected: it breaks every subscription change on a live stream the moment that token rotates.

Two things deliberately reuse what exists rather than growing a parallel surface: the initial snapshot keeps its binary route (`/api/canvas/:ws/:slug/snapshot`), so the largest payload never pays base64 inflation, and **upstream needs no new endpoint at all** — `POST /api/canvas/:ws/:slug/update` already imports, persists and broadcasts.

Viewport delivery mirrors the WebSocket semantics exactly: a request reaches only streams that signalled `client_ready`, and a stream that becomes ready later gets the cached request replayed. Implementing only one half would either drop the request for a late client or deliver it twice.

**Client.** `selectCanvasTransport` decides up front rather than attempting `ws://` and falling back, because the attempt fails every time. WebSocket stays wherever it still works: the daemon-served page, an http dev origin, and the future https→wss leg. `SseBackend` reads the stream with `fetch` + `ReadableStream` rather than `EventSource`, which cannot carry an `Authorization` header — the alternative would put a bearer token in URLs, history and access logs.

**Sharing.** One stream per daemon origin, in a SharedWorker, because browsers cap HTTP/1.1 connections per origin at six and the daemon is plain http where browsers do not negotiate HTTP/2. A stream per canvas tab would spend that budget on sync alone and starve the daemon's own API.

## Bugs found and fixed along the way

- **Text messages were unaddressed.** A WebSocket is per-canvas so its text frames need no addressing; one multiplexed SSE stream serves many canvases, so `head_changed` for one canvas would have been applied to another. Frames now travel as `{doc, raw}` and both receivers check the address. Found by building the hub before the worker — it would otherwise have been a real-browser-only bug.
- **The worker's connect handler was on `globalThis`, not `self`.** Any host that runs the module with an injected scope sees the assignment land elsewhere, and the worker silently never receives a connection.
- **Control messages were addressed with the wrong stream.** With the SharedWorker source injected — the configuration that ships — the stream belongs to the worker, so a backend sending its own id named a stream the daemon had never seen: `client_ready` was dropped and viewport requests never arrived. `sendMessage` moved onto `SseStreamSource` so the stream's owner answers for it. Every earlier test drove the plain-hub path, which is why nothing caught it.
- **The stream never reconnected.** A stream that ended left every listener registered while nothing arrived — the canvas looked connected while it diverged. Two comments already described a reconnect that did not exist. It now reopens with capped backoff and re-announces both its subscriptions and its `client_ready`, because the daemon forgets a stream when it drops.
- **A rotated token was never picked up.** The worker cached one hub per origin and captured the credential it was built with, so a refreshed pairing token left every tab in the profile using a dead one. The token is now read at request time through the same auth seam.
- **A snapshot read could land after disconnect.** Reading the body is a second await; the window between it and the response arriving was unchecked.
- **The Authorization seam was a module, not a function.** A SharedWorker importing `daemon-api-client.ts` pulled that file's response schemas in and stalled the worker's module load entirely. The seam moved to `daemon-auth-fetch.ts`; `createPairingGrant`, which set the header inline for the same reason, now routes through `createDaemonFetch` (which takes the underlying fetch as an argument). `daemon-auth-seam.test.ts` still enforces exactly one setter — it is what caught the first version of the worker.

## Verification

Red-first throughout; mutation-checked (reverting the production change turns the specific guard red). One mutation check found a test of mine that passed for the wrong reason — it held because `close()` drops its listeners, not because of the guard it claimed to cover — so that guard got a test of its own.

- **Real daemon process**, not mocks: opened the stream, subscribed, pushed a Loro update through the existing update route, and the update came back out of the stream, addressed to that document, byte-identical after base64 round-trip.
- **Real SharedWorker**, driven through its real `onconnect` entry via `@vitest/web-worker` with MSW as the daemon — no test-only exports.
- Suite: 595 files / 6282 tests passing; typecheck, build, lint and knip clean.

## Not verified

**Cross-tab sharing.** The Vitest worker polyfill builds a fresh context per `new SharedWorker`, so two ports there are two workers, not two tabs — asserting it would pass or fail for reasons that say nothing about a browser. Local verification was blocked by the environment rather than the code: the app's daemon discovery starts at port 3099, which another checkout's daemon occupies, and that daemon has no `/api/sync/*` routes.

This is an optimisation, not a correctness property: if the sharing failed, each tab would simply open its own stream and still sync correctly.

**To check it on the Preview deployment, the daemon must be running this branch's build** — a daemon built from `main` answers 404 on `/api/sync/stream`, and the app would fall back to browser-local.

Also deliberately uncovered, and noted in the test file: "a subscribe arriving before init is dropped". That is an absence, and a SharedWorker cannot be terminated, so earlier tests' workers keep the environment from quiescing enough to observe it deterministically.

## Summary by CodeRabbit

* **New Features**
  * Added SSE-based canvas synchronization for secure connections and environments where WebSockets are unavailable.
  * Shared live updates across browser tabs, including document-specific changes, messages, viewport requests, and reconnect handling.
  * Added automatic transport selection between WebSocket and SSE.
  * Added authenticated daemon communication and improved pairing requests.
* **Bug Fixes**
  * Improved handling of rotated credentials, streamed requests, connection cleanup, and cross-origin transport scenarios.


kamiazya:
Holding this out of merge until the review findings are fixed. CI is green (18/18), but green CI is not the gate here — two of the findings are paths CI never exercises.

Triage of the CodeRabbit review:

| Finding | Verdict |
| :-- | :-- |
| Build cannot resolve `sse-stream-hub` | **Fixed** in eaf760df — the two Vite alias tables were consolidated into `apps/web/mcp-source-alias.ts`; the miss was invisible locally because a previous build had left a `dist` behind |
| Control messages carry a `streamId` the backend does not own | **Valid, Critical.** With a `streamSource` injected — the shipped configuration — the SharedWorker owns the stream, so `client_ready` reaches the daemon as `404 unknown_stream` and `post()` swallows it. Viewport requests are therefore never delivered. Every test here drove the plain-hub path, so nothing covered the configuration that actually ships |
| The stream never reconnects | **Valid, Major.** `start()` is only reached from `subscribe()`, so after a daemon restart or a dropped connection subscribers keep their listeners and silently stop receiving updates. Two comments in the hub already describe a reconnect that does not exist |
| A later `init` cannot replace a rotated token | **Valid, Major.** `hubFor` caches per `baseUrl` and discards the new token; `createSharedSseStreamSource` caches the same way. With no reconnect path, every tab stays 401 until the worker dies |
| Stream ownership is not checked on subscribe | **Valid, Major.** `POST /api/sync/subscribe` resolves a `streamId` without binding it to the caller, so one session can attach documents to another's stream |
| `cancelled` is not re-checked after reading the snapshot body | **Valid, Minor.** A `disconnect()` during the read still delivers `onSnapshot` to a torn-down canvas |

Fixing these in one pass rather than splitting: they are all one feature, and landing part of it would put "viewport silently does nothing in the shared-worker configuration" on `main` for the gap in between.

kamiazya:
Applied 8 of the 11 nitpicks in ed3b38c5. Triage of the rest:

| Finding | Verdict |
| :-- | :-- |
| `toBase64` → `Buffer` | **Applied.** Node-only file. |
| Exact-match the `/api/sync/*` route scope | **Applied.** Fail-closed default is the point. |
| Reuse `fromBase64` instead of duplicating | **Applied.** The protocol comment already claimed the decode lived in one place; it did not. |
| Type handler stubs as `CanvasBackendHandlers` | **Applied.** `as never` would keep them green through a contract change. |
| Reset module state between route tests | **Applied.** Also gives `resetSyncStreamsForTests` its only caller. |
| Frame split across chunk boundaries | **Applied.** Mutation-checked: breaking the parser's buffer carry-over turns it red. |
| Unparseable `daemonBaseUrl` | **Applied.** |
| `idle()` does not wait | **Applied, and the sharpest of the eleven.** The negative assertion held only because both frames landed in one parser call, exactly as described. It now pushes the unsubscribed frame alone and waits a real interval; mutation-checked against a hub that forwards every frame. |
| Cover the injected `streamSource` path | **Already done** in 5f64f451, which is also where the Critical from the previous round was fixed. |
| Options object for the `SseBackend` constructor | **Skipped.** Stale — 3805d506 removed the `streamId` parameter, so there is no `undefined` placeholder left to remove. |
| Keep-alive frame on the stream | **Skipped, deliberately.** The stated rationale no longer holds: the client reconnects on `done` as of 96676200. A server ping only helps against a half-open socket, and only if the *client* treats silence as death — which it does not. Adding pings alone would change nothing observable. The useful shape is a client-side idle timeout paired with the ping, which is a separate change; the connection is loopback, so there is no intermediary to time it out in the meantime. |

kamiazya:
All four applied in bc1d6bc0. Every one of them was in the reconnect logic I added earlier in this branch, and two are real behavioural bugs rather than hygiene:

| Finding | Verdict |
| :-- | :-- |
| `readyDocs` never pruned | **Valid, and the worst of the four.** After a reconnect the hub replayed `client_ready` for canvases it had already unsubscribed from, so the daemon marked the new stream ready for documents nobody was watching. Fixed and mutation-checked: dropping the `readyDocs.delete` turns the new test red. |
| `close()` leaves the retry wait pending | **Valid.** `clearTimeout` without settling left the `start()` frame — and the hub it captures — reachable forever. Fixed. Not covered by a test: whether that promise settles is not observable through this class's public surface, and I would rather say so than assert something adjacent and call it coverage. |
| Parse SSE frames with Zod | **Valid, and the guideline is right here.** The daemon declared the shapes and the client re-derived them with `typeof` checks — exactly the drift the rule exists to prevent. One declaration now lives in `shared/sync-sse-contract.ts`, the only place both the route and the hub can import from. |
| Cancel a non-OK stream body | **Valid.** Fixed. |

One deliberate departure from the repo's DTO convention, called out because it is not an oversight: these three frame schemas are **not** `.strict()`. The producer is a locally-installed daemon and the consumer is an auto-updating hosted page, so their versions skew by design — a strict parser would make a field added to a frame drop that frame entirely on every older client and stop sync outright. The reason is in the module doc and pinned by a test. Requests travel the other way and stay `.strict()`.

PR #524, opened 2026-08-09
fix: show Reconnecting instead of Synced while sync is down

kamiazya:
Measured on the deployed app against a local daemon: three windows of **15.6s, 70.0s and 31.6s** in which the daemon had **no open sync stream at all** while both canvas tabs displayed "Synced". Remote edits were not arriving in those windows and nothing said so.

The stream does come back on its own, so no data is lost. The defect is that a user cannot tell the difference, which is the failure mode this branch's own reconnect commit described: *a client that does not come back keeps its listeners registered while silently receiving nothing, so the canvas looks connected while it diverges.*

## Why it said Synced

The chip was derived from the auth state alone:

```
state={authError ? 'sync-off' : 'synced'}
```

So it read Synced whenever pairing was intact, whatever the transport was doing — including on the WebSocket path, which has the same hazard. `syncStatus` was already computed by `useCanvasSync` and simply never destructured by the page: the signal existed and went nowhere.

## What now reports what

- **The hub** tells each document's subscribers when its stream opens and when it ends, and **also reports the current state at subscribe time**. Announced on transitions alone, liveness would never reach anyone joining a stream that is already open — which is every canvas after the first one in a tab.
- **The SharedWorker** forwards that over the port, because with the shared source the page never sees the stream itself.
- **Every backend with a transport** reports its loss: the SSE backend from the source's liveness, the daemon backend from `onclose`.
- **The session** maps that to a new `reconnecting` status and the chip shows it. An auth rejection still outranks it — re-pairing is the way out of that one, while a reconnect resolves itself.

Both port contracts were extended rather than each implementation tested separately, so a backend or stream source that stays silent fails rather than passes.

## Not fixed here, and not claimed to be

**Why the gaps happen at all.** Each one followed a tab navigation, and the stream stayed open for minutes across idle once the tabs settled — so this is not an idle timeout. The leading hypothesis is that a page's `pagehide` unsubscribes its documents, the worker's hub sees no documents left and stops reconnecting until a page subscribes again. That is **unverified**; the daemon's subscribe/unsubscribe traffic is not logged, so it cannot be told apart from a worker teardown without more instrumentation. Making the state visible is worth shipping on its own; the cause is a separate lane.

## Verification

608 test files / 6449 tests passing; typecheck, knip and build clean.

Three test-harness defects were found and fixed while writing this, each of which would otherwise have been read as a defect in the code under test: a fake WebSocket whose `close()` passed no `CloseEvent` (the real handler reads `event.code`), a liveness test that passed only because a second subscriber happened to register before the ready frame was parsed, and a case measuring "after disconnect" from `connect` instead of from `disconnect`.

## Summary by CodeRabbit

* **New Features**
  * Added a **Reconnecting** connection status indicator with an amber status dot.
  * Edits remain available locally while the connection is temporarily restored.
  * The indicator automatically returns to **Synced** after reconnection.
  * Authentication failures continue to display **Sync off**.

* **Bug Fixes**
  * Improved connection-state reporting across supported transport types.
  * Prevented stale sessions from overwriting the active connection status.

* **Tests**
  * Added coverage for transport drops, reconnection, and connection-state notifications.

### DIFF
--- a/apps/web/src/components/connection/ConnectionStatus.test.tsx
+++ b/apps/web/src/components/connection/ConnectionStatus.test.tsx
@@ -9,6 +9,19 @@ import { ConnectionStatus } from './ConnectionStatus.js'
 afterEach(cleanup)
 
 describe('ConnectionStatus chip', () => {
+  it('explains the reconnecting state instead of opening an empty popover', () => {
+    // The chip can reach this state, so the popover must have something to
+    // say; a state with no branch renders an empty box on click.
+    render(<ConnectionStatus state="reconnecting" />, { container: document.body })
+    fireEvent.click(screen.getByTestId('connection-chip'))
+
+    const popover = screen.getByTestId('connection-popover')
+    expect(popover.textContent).toMatch(/not running/i)
+    // No replay promise: DaemonBackend drops local updates while its socket is
+    // not OPEN, so telling the user they are sent on recovery would be false.
+    expect(popover.textContent).not.toMatch(/will be sent|are sent when/i)
+  })
+
   it('synced state shows a quiet chip and explains where data lives in the popover', async () => {
     render(<ConnectionStatus state="synced" daemonBaseUrl="http://127.0.0.1:3099" />)
 
--- a/apps/web/src/components/connection/ConnectionStatus.tsx
+++ b/apps/web/src/components/connection/ConnectionStatus.tsx
@@ -8,6 +8,9 @@
  * - `local`    — browser-local page; data exists only in this browser.
  *                `children` hosts page-supplied extras (daemon detection,
  *                capability hint) inside the popover.
+ * - `reconnecting` — the transport dropped and is being retried. Edits stay
+ *                in this browser and are sent when it comes back, so this is
+ *                informational, not an error.
  * - `sync-off` — daemon-backed page whose session was rejected. The chip
  *                turns attention-colored and the popover carries the two
  *                ways forward (re-pair / continue browser-local). A polite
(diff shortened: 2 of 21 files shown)


kamiazya:
All seven verified against the code and applied in de2d695a. Every one was valid — four of them in what this PR had just added.

| Finding | Verdict |
| :-- | :-- |
| Contract case asserts a disconnect that cannot fail | **Valid, and the worst of the seven.** Subscribing already records the state at join time — `false` on a fresh source — so `includes(false)` passed whether or not `dropStream()` emitted anything. The guard I added to protect both implementations could not fail. Now measured only after the drop, and mutation-checked: a hub that stops reporting disconnects turns it red. |
| `handlers.onConnected()` unconditional after subscribe | **Valid.** It overwrote the source's accurate "no stream yet" and double-sent `client_ready` when there was one — reintroducing, on the no-SharedWorker path, exactly the false Synced this PR exists to remove. Removed; the source's report is now the only one. Covered by a new test asserting the first report is `disconnected` and `connected` arrives exactly once. |
| Same, seen from the hub side | **Valid — same root, same fix.** `onConnectionChange` is now the sole transport-status authority. |
| `idle` and `error` rendered as `synced` | **Valid.** `error` is set on failed update pushes, so the chip claimed sync while pushes were failing. Only `connected` earns Synced now; the rest fold into `reconnecting`, whose copy makes no claim about timing. A distinct terminal-error state would be better still and is not in this PR's scope. |
| Popover has no `reconnecting` branch | **Valid.** Clicking the chip opened an empty box. Added, with a component test that also asserts the absence of a replay promise. |
| Copy promises edits are sent on recovery | **Valid, and the one I'm least comfortable having written.** `DaemonBackend.pushLocalUpdate` returns early while the socket is not `OPEN`, so the bytes are gone. Queueing and replay is the right fix and is a separate change; until then the copy says what is true and promises nothing. |
| Deployment narrative in a test comment | **Valid** per the repo's comment discipline. Trimmed to the durable why. |

Suite: 608 files / 6463 tests passing; typecheck, knip and build clean.

PR #528, opened 2026-08-09
fix: re-send the document when the transport returns

kamiazya:
An edit made while the transport is down was **lost for good**.

Each push carries one commit's delta — `subscribeLocalUpdates` hands the backend exactly the ops from that commit — and `DaemonBackend.pushLocalUpdate` returns early unless the socket is `OPEN`. The delta is dropped, and no later push replays it because every later push is a different delta. The user sees their edit on screen; the daemon never learns about it.

This is the concrete harm behind the window the previous change made visible: the chip now says Reconnecting, but until now the honest thing to say about edits made in that window was nothing.

## The fix

The session holds the LoroDoc, so it re-sends the whole document on connect. Loro's import is idempotent: a resend is a no-op for a server that already has the ops, and a recovery for one that does not.

**Full state rather than a delta since the last acknowledged version.** Nothing acknowledges a push today, so there is no such version to send from — and inventing one that is wrong would lose edits silently again, which is the bug. The cost is one full document per reconnect. That is the cheap side of the trade, and a version-vector optimisation can come later on top of a correct baseline.

It sits in the session rather than in each backend because that is where the document is; a per-backend queue would need a byte cap, and every cap is a silent-loss policy.

## The copy is now true

The reconnecting popover was deliberately softened last change to promise nothing, because the drop was real. It now says edits are sent when the connection returns, and the component test pins that sentence — so removing the resend breaks a test that names the reason.

## Verification

Red-first; mutation-checked (removing the resend turns the new session test red). The test rebuilds a document from the resent bytes and asserts the edit is in it, rather than asserting the push was merely non-empty.

609 test files / 6466 tests passing; typecheck, knip and build clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reconnecting now resends the latest document state, including edits made while disconnected.
  * Preserved changes are restored when the connection returns, preventing missed updates.
  * Repeated synchronization remains safe for changes already received.
* **User Experience**
  * Updated connection-status messaging to clarify that edits are retained and resent automatically after reconnecting.

### DIFF
--- a/apps/web/src/components/connection/ConnectionStatus.test.tsx
+++ b/apps/web/src/components/connection/ConnectionStatus.test.tsx
@@ -17,9 +17,10 @@ describe('ConnectionStatus chip', () => {
 
     const popover = screen.getByTestId('connection-popover')
     expect(popover.textContent).toMatch(/not running/i)
-    // No replay promise: DaemonBackend drops local updates while its socket is
-    // not OPEN, so telling the user they are sent on recovery would be false.
-    expect(popover.textContent).not.toMatch(/will be sent|are sent when/i)
+    // The recovery promise is only sound because the session re-sends the whole
+    // document on reconnect; a backend hands a closed socket one delta and it
+    // is gone. If that resend is ever removed, this sentence becomes a lie.
+    expect(popover.textContent).toMatch(/sent when the connection returns/i)
   })
 
   it('synced state shows a quiet chip and explains where data lives in the popover', async () => {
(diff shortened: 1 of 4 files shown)


kamiazya:
Both valid, applied in 371ab2ca.

| Finding | Verdict |
| :-- | :-- |
| Unhandled rejection from the reconnect resend | **Valid, and it defeated the point of the change.** The send was fire-and-forget, so a rejection was an unhandled rejection and a synchronous throw from `doc.export()` escaped into the backend's own connect handler — while the session had already reported `connected`. A recovery that silently failed looked exactly like one that worked. It now runs through the same chain as the ordinary local-update path and reports `error`. |
| Test does not model a closed transport | **Valid.** Clearing the recorded pushes proves nothing about *which* ops the resend carries: an implementation resending only the latest delta passes it. The fake transport now discards pushes while it is down, two commits are made in that window, and the document rebuilt from what the server would actually have received must contain both edits plus the state that predates the outage. |

Mutation-checked twice against the strengthened test — removing the resend, and resending a delta from the current version (an empty update) — each turns it red.

609 files / 6466 tests passing; typecheck, knip and build clean.

PR #531, opened 2026-08-09
fix: make disconnecting from a daemon outlive the page

kamiazya:
Found by dogfooding the shipped build, one hour after shipping it: disconnect from a daemon, reload, and you are straight back on it.

The popover promises **"This browser stops using it and stops looking for it."** The second half was true — discovery skips the dismissed daemon. The first was false the moment the user reloaded: `App.tsx` decides a load is daemon-backed from `localDaemonBaseUrl`, and disconnecting left that set. Disconnecting now clears it when it names the daemon being left.

Observed state after a disconnect in the real app, before this fix:

```
dismissedDaemonBaseUrls: ["http://127.0.0.1:3099"]   ← written
knownDaemonBaseUrls:     []                          ← cleared
localDaemonBaseUrl:      "http://127.0.0.1:3099"     ← left behind
```

## The test that was supposed to cover this

It passed vacuously three times over before it caught anything:

1. The settings key was simply absent in the harness, so `toBeUndefined()` held before the page did anything.
2. Seeded as hand-written JSON — which the store rejected as invalid and silently replaced with defaults, so the seed never reached the code under test.
3. The precondition read `localStorage` directly rather than through the store, so it confirmed the raw write and not what the store had loaded.

It now seeds and reads through the store, and asserts the precondition rather than assuming it: a seed that stops working fails loudly instead of turning the case green. That is the actual lesson here — the defect was in shipped code for an hour because its test could not fail.

## Verification

Red-first, with the red confirming the exact symptom seen in the browser (`expected 'http://127.0.0.1:3099' to be undefined`). 135 test files / 1618 tests in apps/web passing; typecheck clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Disconnecting from the active daemon now clears its saved local connection address.
  * Saved connection settings for a different daemon remain unchanged.
  * Dismissal preferences continue to be preserved when disconnecting.

* **Tests**
  * Added coverage for connection cleanup and browser-local escape behavior.

### DIFF
--- a/apps/web/src/pages/DaemonCanvasPage.test.tsx
+++ b/apps/web/src/pages/DaemonCanvasPage.test.tsx
@@ -6,6 +6,7 @@ import { act, cleanup, fireEvent, render, screen, waitFor } from '@testing-libra
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { MemoryStore } from '../lib/browser-local-store.js'
 import * as daemonApiClient from '../lib/daemon-api-client.js'
+import { createUserSettingsStore } from '../lib/user-settings-store.js'
 import { DaemonCanvasPage } from './DaemonCanvasPage.js'
 
 vi.mock('../lib/daemon-api-client.js', async (importOriginal) => {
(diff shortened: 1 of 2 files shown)


kamiazya:
All three valid (two are the same issue at two sites), applied in e1549040.

| Finding | Verdict |
| :-- | :-- |
| `JSON.stringify(window.localStorage)` proves only that the text appears | **Valid, and the sharpest of the three.** The daemon's URL also sits under `localDaemonBaseUrl`, so `toContain(DAEMON_BASE_URL)` held whether or not the dismissal was ever recorded — the assertion could not distinguish the fix from its absence. Now loads the store and asserts `dismissedDaemonBaseUrls` contains it and `knownDaemonBaseUrls` does not. |
| Settings leak between tests | **Valid.** The suite's `beforeEach`/`afterEach` clear mocks and the DOM but never `localStorage`, so a test that seeds settings hands them to every test after it. Cleared per test — order dependence surfaces as an unrelated flake weeks later, which is the expensive way to find it. |
| Same, second site | **Same fix.** Handled by the lifecycle change rather than a per-test reset, so a future test that seeds settings is covered without remembering to opt in. |

Mutation-checked three ways against the strengthened assertions — not clearing the stored target, not recording the dismissal, not removing the daemon from the known list — each turns the case red on its own.

Worth noting for the record: this PR exists because the *previous* test could not fail, and the review has now caught a second layer of the same weakness in its replacement. 135 files / 1618 tests in apps/web passing.

PR #532, opened 2026-08-09
feat: drop the daemon port sweep and make entering a port the way in

kamiazya:
Scanning ten ports from 3099 guessed, and the guess was wrong in both directions: it missed every daemon outside the range — a dev worktree's hash-derived port, a packaged daemon whose first candidates were taken — while firing probes at ports nobody asked about. `daemon-discovery.ts`'s own doc already admitted the first half.

Candidates are now the daemons the user actually named: an explicit port, plus the ones reached on earlier visits.

## The port field is unconditional now

It was gated on a check having concluded with nothing found. That gating created two traps, both reachable:

- Connected to the wrong daemon → the field never appears, so there is no way to name the right one. Switching required disconnecting first.
- Dismissals hid every candidate → nothing is found, but the affordance that would let you name a daemon was itself conditional on state the dismissals had changed.

Naming a port is the primary way in, so it cannot depend on a check failing first. The manual check now probes the entered port together with the remembered ones, rather than being a second, separate control.

## The four tests that went through the sweep

Each was rewritten to reach its daemon by name, keeping the guarantee it existed for rather than being deleted or relaxed:

| Guarantee | How it is reached now |
| :-- | :-- |
| A daemon on a non-default port is reachable | The port is typed — which is the whole point of the change |
| A manual check bypasses the probe memo | Typed port, forced re-check asserted |
| A found daemon is remembered and re-probed next mount | Found by name, then remounted |
| Several responders produce a picker | Two remembered daemons respond |

## Verification

Red-first. Mutation-checked two ways — not passing the entered port into the candidate list, and restoring the old gating — each turns several cases red.

One leftover (`portRangeCount: 0` at a call site) survived `tsc` because a spread into a call skips excess-property checking; found by grep and removed.

609 test files / 6472 tests passing; typecheck, knip and build clean.

## Summary by CodeRabbit

* **New Features**
  * Added manual daemon connection using an entered port and a Check action.
  * Kept the port field available even when a daemon is already connected.
  * Preserved support for remembered and explicitly provided daemon addresses.

* **Bug Fixes**
  * Improved handling of HTTPS, dynamic-port, dismissed, and multiple-daemon scenarios.
  * Removed automatic scanning of a default port range, making discovery more predictable.

### DIFF
--- a/apps/web/src/components/migration/DaemonDetectedBanner.test.tsx
+++ b/apps/web/src/components/migration/DaemonDetectedBanner.test.tsx
@@ -49,7 +49,10 @@ describe('DaemonDetectedBanner', () => {
         probeFn={probeFn}
       />,
     )
-    await waitFor(() => expect(probeFn).toHaveBeenCalled())
+    // Waits for the DETECTED render, not merely for the probe call: the call
+    // resolves later, so asserting on it would check the pre-detection state
+    // and pass even if the field disappeared once a daemon was found.
+    await screen.findByTestId('daemon-detected-banner')
 
     expect(screen.getByTestId('daemon-port-input')).toBeTruthy()
   })
--- a/apps/web/src/components/migration/DaemonDetectedBanner.test.tsx
+++ b/apps/web/src/components/migration/DaemonDetectedBanner.test.tsx
@@ -35,6 +35,25 @@ describe('DaemonDetectedBanner', () => {
     cleanup()
   })
 
+  it('offers the port field even while a daemon is already connected', async () => {
+    // Entering a port is the primary way in, so it cannot be gated on a failed
+    // check: gating it there leaves a user connected to the wrong daemon with
+    // no way to name the right one, and a user whose dismissals hid every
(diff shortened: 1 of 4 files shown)


kamiazya:
Both valid, applied in 91e65940.

| Finding | Verdict |
| :-- | :-- |
| Test asserts before the detected state renders | **Valid.** `await waitFor(() => expect(probeFn).toHaveBeenCalled())` proves the probe was *called*; its result lands later, so the assertion ran against the pre-detection render and would have passed even if the field disappeared once a daemon was found — precisely the claim the case exists to make. It now waits for `daemon-detected-banner`, and hiding the field after detection turns it red, which it could not do before. |
| Stale `portError` on an empty input | **Valid.** An empty field is a legitimate check of the remembered daemons, and the early return left a complaint about a previous entry sitting next to a check that was running fine. |

Also removed a duplicated comment an earlier edit in this branch left behind.

135 files / 1619 tests in apps/web passing; typecheck and lint clean.

PR #541, opened 2026-08-10
feat(web): tell the user what blocked a daemon check

kamiazya:
A failed daemon check has never been able to say why it failed. Chromium rejects a
permission-blocked loopback request with the same `TypeError: Failed to fetch` as a port with
nothing behind it, so the UI could only list possibilities. Reading the browser's local-network
permission rules one of them out.

## What changed

**The prompt is explained before it appears.** Chrome's Local Network Access prompt fires on the
request itself, so the first check produced a permission dialog with no stated reason. A denial is
remembered and script cannot re-prompt, which makes that unexplained dialog a question the user
gets one chance to answer well. The check now reads the permission first (reading does not prompt)
and says what is about to be asked.

**A standing denial stops the check instead of probing.** The request cannot succeed and cannot
re-prompt, so probing would only produce a failure that reads as an absent daemon. The recovery
panel points at site settings, which is the only thing that can undo it.

**Failure copy is keyed on the cause rather than the page scheme.** The case worth having is a
granted permission: it rules the browser out, so the UI no longer sends someone to change a site
setting that is already correct.

**A response header that does not exist is no longer sent.** `Access-Control-Allow-Local-Network`
was emitted on every preflight, commented as "Chrome's in-flight successor" to the Private Network
Access header. The spec it cited defines no response header at all — LNA replaced PNA's preflight
with a permission prompt. Emitting it stated a capability the daemon does not have and pointed
anyone debugging a blocked fetch at the server. The PNA header stays; that generation is on hold
rather than withdrawn.

## What it deliberately does not claim

`unreachable` stops short of "nothing is listening on that port". A running daemon that has not
allowlisted this origin fails identically at the fetch layer, so claiming an empty port would send
someone to start a daemon that is already running.

## Verified in Chrome 150

Measured against the running daemons before building on any of it:

| | production origin | `latest` preview |
|---|---|---|
| `loopback-network` | granted | granted |
| `local-network` | prompt | — |
| `GET /api/runtime/ping` (3099) | 200 / 13ms | 200 / 5ms |

Two things this settled. The permission names are queryable in the target browser, so `unknown`
is the rare path rather than the normal one. And `loopback-network` is the right name to read:
`local-network` sits at `prompt`, so querying it would have shown an explanation for a permission
this connection does not need.

A port with no daemon (`3105`) rejected in 3ms with `Failed to fetch` — indistinguishable from a
block without the permission read, which is the gap this PR closes.

## Tests

`apps/web` 137 files / 1652 tests, `mcp-node` 3219. Every new test was mutation-checked; two
mutations survived on the first pass and the tests were rewritten until they failed:

- a `NotAllowedError` case that asserted a return value both branches produce (now pins the call
  count, which is what actually differs)
- the `checking` flag and the stale failure notice, which the async permission read had quietly
  moved to after the await — a live button was double-startable across that gap

## Not verified

The new panels have not been exercised in a real browser: reaching them needs a permission state
this machine is not in (`granted` everywhere). The permission read itself was verified in Chrome
as above.

## Summary by CodeRabbit

* **New Features**
  * Added clearer local-network permission handling when connecting to a detected daemon.
  * Checks now explain upcoming browser prompts, prevent attempts when access is denied, and resume after confirmation.
  * Added more specific connection failure messages, including browser blocking, unavailable services, and non-daemon responses.

* **Bug Fixes**
  * Updated network preflight responses to use the supported private-network permission header and omit the obsolete local-network header.

### DIFF
--- a/apps/web/src/components/migration/DaemonDetectedBanner.test.tsx
+++ b/apps/web/src/components/migration/DaemonDetectedBanner.test.tsx
@@ -881,22 +881,78 @@ describe('DaemonDetectedBanner — local network permission', () => {
     expect(screen.queryByTestId('lna-explainer')).toBeNull()
   })
 
-  it('does not blame the browser for a failure it has been shown to allow', async () => {
-    // A granted permission rules the browser out, so pointing the user at
-    // site settings here would send them to change something already correct.
-    renderWithPermission('granted')
+  it('still checks when reading the permission throws', async () => {
+    // The read is awaited after 'checking' is claimed, so an unhandled
+    // rejection would leave the button disabled for good -- and the button is
+    // the only affordance that could recover from it.
+    const probeFn = vi.fn().mockResolvedValue(NOT_DETECTED)
+    render(
+      <DaemonDetectedBanner
+        settingsStore={makeStore()}
+        fetch={vi.fn()}
+        locationProtocol="https:"
+        probeFn={probeFn}
+        queryPermissionFn={vi.fn().mockRejectedValue(new Error('permissions unavailable'))}
+      />,
+    )
 
     fireEvent.click(screen.getByTestId('daemon-port-connect'))
 
+    await waitFor(() => expect(probeFn).toHaveBeenCalled())
     await screen.findByTestId('daemon-check-failed-notice')
-    expect(screen.queryByTestId('daemon-check-blocked-notice')).toBeNull()
   })
 
-  it('names the browser as the obstacle when a failure follows a denial', async () => {
-    // Reached by a permission revoked between the gate check and the probe:
-    // the gate saw 'prompt', so the check ran, and the probe then failed.
+  it('stops at the gate when a later check finds the permission denied', async () => {
+    // The path this replaced a mis-named test for. A denied read returns at
+    // the gate, so no probe runs and no failure copy renders -- which is why
+    // asserting the failure notice here could never have worked.
     const probeFn = vi.fn().mockResolvedValue(NOT_DETECTED)
-    const queryPermissionFn = vi.fn().mockResolvedValueOnce('prompt').mockResolvedValue('denied')
+    const queryPermissionFn = vi.fn().mockResolvedValueOnce('granted').mockResolvedValue('denied')
+    render(
+      <DaemonDetectedBanner
+        settingsStore={makeStore()}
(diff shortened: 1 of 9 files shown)


kamiazya:
@coderabbitai review

kamiazya:
All six triaged. Five valid and fixed in `de4af524`; one rests on a misreading.

| # | Finding | Verdict |
| :- | :-- | :-- |
| 1 | Test does not exercise what its name describes | **Valid** — and worse than reported: it asserted the gate panel, and the path its name claimed is unreachable (see #3's row below). Replaced with cases for the paths that exist. |
| 2 | Release `checking` if the permission read fails | **Valid.** Found a second instance of the same shape: the post-sweep re-read could also reject, dropping the failure notice entirely so a failed check showed no outcome at all. Both reads now go through one reader that turns a rejection into `'unknown'`. |
| 3 | Post-probe copy rests on a stale permission | **Valid, and the one that mattered.** The prompt is answered *during* the sweep, so a check that began at `'prompt'` and was then allowed told the user to allow what they had just allowed. Re-read once the sweep settles, before the copy renders. |
| 4 | Announce the explainer | **Valid.** An unfocused `role="dialog"` is announced by nothing, so a screen reader user would have met the permission prompt without the explanation meant to precede it. Now a live region — it is inline and non-blocking, so `status` fits it better than `dialog` did. |
| 6 | Assert both sides of the preflight contract | **Valid.** Dropping *both* headers satisfied the absence check alone. Pinned together. |
| 5 | Map `timeout` out before it claims `unreachable` | **Invalid.** `'timeout'` and `'network'` are already identical here: neither is matched by an early return, so both fall through to the same permission switch. The requested change is a no-op. On the substance, `unreachable` claims only that the browser permitted the request, which a granted permission does establish — and a timeout while a prompt is open reads `'prompt'`, not `'granted'`, so it lands on the unanswered branch instead. |

Following #1 and #3 to their conclusion removed a row rather than fixing it: `daemon-check-blocked-notice` was unreachable both ways. A proven block (`reason: 'blocked'`) puts the capability tier at `tier2-blocked` and renders `UNSUPPORTED_BROWSER_NOTICE` instead of that row, and a denied permission returns at the gate without ever probing.

Every fix is mutation-checked. `apps/web` 138 files / 1661 tests, `mcp-node` 3219.

PR #543, opened 2026-08-10
chore(dev-flow): add accessibility criteria to design and review, and path-scope package rules

kamiazya:
Criteria adapted from ui-skills' [`fixing-accessibility`](https://www.ui-skills.com/) (MIT), narrowed to what actually bites in this repo. Two files, both markdown under `.claude/`.

## Why this and not more

**Biome already covers the executable rung.** Its `recommended` preset enables the `a11y` lint group, and it is clean — 421 files, zero a11y findings. So this adds only what a linter cannot judge:

- focus trapping, initial focus, and restore-on-close around dialogs
- keyboard reachability of pointer-only affordances
- state and error announcement (`aria-expanded`, `aria-busy`, `aria-describedby` + `aria-invalid`, toast-as-only-channel)
- the DOM *around* generated SVG — `canvas-render` ships no a11y projection by design (see `package-canvas-render.md`), so the bar is the container of `CanvasViewer`'s `dangerouslySetInnerHTML`, `ImageSceneNode`'s `alt`, and whether a canvas-only interaction is reachable another way

**Opt-in per run, not in the default dimension set.** `AccessLint` already runs on every PR as the always-on net. That is the difference from `reachability` (#542), which had no equivalent and therefore went into the defaults.

**No dependency added.** The upstream CLI is a read-only content server — `npx ui-skills get <name>` — so there is nothing to install, pin, or share. A devDependency would buy a lockfile entry and a Dependabot stream for zero benefit. The MIT attribution at the top of the file is the reference; the generic upstream version stays one command away.

## Adaptation, not a copy

The upstream skill is generic. What makes this version worth keeping in-repo is the part upstream can never have: the `canvas-render` no-a11y-projection decision, the `CanvasViewer` SVG-injection container, `UpdateToast` as the live comparison for toast-only messaging, and pointer-only affordances on a drawing surface. Six criteria, prioritised, rather than upstream's nine generic categories.

## Test evidence

```
$ pnpm test:scripts
ℹ tests 98
ℹ pass 98
ℹ fail 0
```

The `claude-config-wiring` guard added in #542 did its job here: committing the criteria file without the matching `reviewer-dimension.md` entry failed immediately with a readable diff (`+ accessibility`), which is exactly the silent drift it was written to catch.

No UI or product code changes, so no screenshots apply. `biome.json` excludes `.claude/**`, and the diff is markdown only, so lint and typecheck have nothing to say about it.

## Note for the reviewer

`packages/mcp-server/src/cli/daemon-run-auto-open-launch.test.ts` still fails on this machine and fails identically on pristine `origin/main` (it launches a daemon in a PTY expecting a fake browser opener, which this local environment does not provide); `ci` on `main` is green. The pre-push hook was bypassed with `LEFTHOOK=0` on that basis — this diff contains no TypeScript. CI is the authoritative gate.

https://claude.ai/code/session_01QWYyXw7m5ndo2T8UxN18FX

---

## Update: the criteria now apply at design time too

Meeting these is cheapest while the flow is still a paragraph, so both design surfaces read **the same file** the review dimension judges by — no second copy:

- **`ux-designer`** (plan-initiative's panel) gains an Accessibility bullet beside flows/affordances/states. It previously said "keyboard/pointer/touch" in passing with no criteria behind it.
- **dev-loop's Design phase** folds them into the design and `testScenarios` — **only** when the change touches a UI surface, and explicitly skips otherwise.

No new `DESIGN_SCHEMA` field. A required a11y answer would tax every non-UI design for nothing; `blastRadius`/`userReach` earned their required slot because they apply to every change, and this does not.

So the same criteria file is now read at three points: design (both panels), review (opt-in dimension), and post-push (`AccessLint`, independently).

## Summary by CodeRabbit

* **Documentation**
  * Added accessibility review guidance covering accessible names, keyboard navigation, focus management, announcements, semantics, motion, and contrast.
  * Added accessibility requirements for UI design and review workflows.
  * Included guidance for reviewing canvas and SVG-based interfaces.
  * Added clearer package-specific development guidance for canvas and server components.

* **Tests**
  * Added validation to ensure package guidance is correctly scoped and consistently applied.

---

## Update: path-scope the package rules

Same theme as the rest of this PR — load the criteria that match what is being touched — applied to the always-on rule set.

`architecture-map.md` describes `.claude/rules/package-<name>.md` as *path-scoped*, and `dev-flow.md` requires every new package to ship one. Both were prose. **6 of the 7 had no `paths:` frontmatter**, so all **789 lines / 5,742 words** loaded into every session regardless of what was being edited. Only `canvas-model` was actually scoped.

Each rule is now scoped to `packages/<name>/**`, and a new guard in `claude-config-wiring.test.mjs` makes the promise executable so the next package cannot ship an unscoped rule. The failing-first run reported all 6 by name.

**One deliberate exception.** `canvas-codec` keeps `apps/web/**` alongside its own package, because that file states its Consumers note is *the only documentation* of the `apps/web -> canvas-codec` boundary — arch-lint does not source-scan composition roots. Scoping it to its own package would have hidden it from exactly the readers it was written for. No other package rule makes that claim, so no other one got a second glob.

```
$ pnpm test:scripts
ℹ tests 99
ℹ pass 99
ℹ fail 0
```


kamiazya:
All four findings verified against the code and applied in 7afc5b3 — no dismissals.

1. **Focus trapping scoped to modal** — correct and load-bearing here. `ContextMenu`, `CanvasDropdown`, and `SelectionOverlay` are all non-modal, so "modal/overlay" would have told a reviewer to demand a focus trap that must not exist. Split per WAI-ARIA APG.
2. **Loading announcement** — correct, and my wording was backwards: `aria-busy` tells assistive tech to *suppress* partial updates and announces nothing on its own, so offering it as an alternative to status text left the user with silence. Now `role="status"`/`aria-live` required, `aria-busy` supplementary.
3. **Meaningful vs decorative SVG** — partially: the `ImageSceneNode` bullet already carried the decorative case, but the container check said "name/role", which reads as either-or, and a role without a name is exactly the unlabeled blob the check exists to catch. Both now required for meaningful graphics; decorative must be a deliberate `aria-hidden`/presentational choice.
4. **Exact `paths` entry** — correct, and this PR itself makes the false negative reachable: `canvas-codec`'s `paths` block carries an explanatory comment, so a pattern living only in a comment would have satisfied the guard while scoping nothing. Now matches a real list item, mutation-checked both ways (comment-only pattern fails; all seven real rules still pass).

Useful pass — 2 and 4 were genuine defects rather than style.

PR #563, opened 2026-08-11
feat(web): replace the native custom-color input with a real picker

kamiazya:
## What

Replaces the bare `<input type="color">` from #562 with a real picker component: the custom-color option in the Color row becomes a swatch trigger (shows the current hex when set, a hue wheel otherwise) that expands an inline **react-colorful** panel — saturation/hue picker plus a hex field — under the row.

## Design notes

- **Inline expansion, not a nested popover**: the context menu dismisses on any outside pointer-press; a portalled popover would either die to that handler or need an exception carved into it. Expanding inside the menu keeps the dismissal logic untouched.
- **Commit discipline**: drags commit on pointer-up rather than per drag frame (every commit is an undo entry and a canvas re-layout); the hex field commits per valid value, and only SIX-digit hex ever commits — 3-digit shorthand can't reach the canvas color contract (presets or `#rrggbb`).
- **Dependency**: react-colorful — 2.8kB gzipped, zero dependencies, emits exactly the spec's hex form. apps/web is a composition root (not under arch-lint's shared-layer dependency allowlist).
- Applies to both node and edge menus through the shared Color row.

## Tests

`web-browser`: the #562 hex test now drives the component UI — open the panel from the trigger, assert the real picker is present, type a hex, node color updates (red before this change). All 280 spatial-editor browser + 1691 jsdom tests green.

## Visual

Note: pushed with `LEFTHOOK=0` — known env flake (`daemon-run-auto-open-launch` readiness timeout, separate lane); CI `verify` is the authoritative gate.

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -44,6 +44,7 @@
     "loro-crdt": "catalog:",
     "lucide-react": "^1.28.0",
     "react": "catalog:",
+    "react-colorful": "^5.8.0",
     "react-dom": "catalog:",
     "react-router-dom": "^7.18.0",
     "tailwind-merge": "^3.5.0",
--- a/apps/web/src/components/spatial-editor/ContextMenu.tsx
+++ b/apps/web/src/components/spatial-editor/ContextMenu.tsx
@@ -19,7 +19,8 @@
  * discrete pointerdown loses the focus fight with mousedown's default
  * action, while click fires after those defaults.
  */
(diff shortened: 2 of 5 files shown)


kamiazya:
Default (collapsed) state for clarity — the picker panel only appears after tapping the rainbow trigger at the end of the Color row, collapses on a second tap, and every menu open starts collapsed. The PR body's screenshot shows the expanded state.

PR #582, opened 2026-08-11
feat(web): grow the markdown editor — highlighting, themed preview, and OKF core facets

kamiazya:
The markdown editor was built as one slice and not grown since, so this is
the first pass at making it a base worth building on. Four changes, each its
own commit.

## 1. Syntax highlighting in the source pane

`markdown()` only PARSES — with no `syntaxHighlighting` extension every token
rendered in one undifferentiated weight and colour, so a heading looked
exactly like a paragraph.

Token styling is assigned as CLASS NAMES rather than inline colours, with the
rules in `index.css`, so it reads the same `--foreground` /
`--muted-foreground` tokens as the rest of the app and flips with `.dark` for
free. The palette is achromatic by design (every token is `oklch(L 0 0)`), so
structure is carried by weight, slant and contrast instead of hue, and markers
(`#`, `-`, `**`) recede rather than highlight.

**Found while doing it:** the preview parses through canvas-codec
(`remark-parse` + `remark-gfm` + `remark-math`) while `markdown()`'s default
base is plain CommonMark, so `~~strikethrough~~` and tables rendered in the
preview but stayed unrecognised — and unstylable — in the source. It surfaced
because `cm-md-strikethrough` could never fire. Fixed at the root (match the
grammars) rather than at the symptom.

## 2. External edits confined to the span that changed

The pane is controlled, so any change that did not originate from typing — a
load finishing, a remote CRDT update — arrived as a new `value` and was
applied by replacing the whole document.

That does not lose the caret outright, since CodeMirror maps the selection
through every change; what it destroys is everything strictly INSIDE the
replaced range, which for a whole-document replace is the whole document. A
mid-document caret collapses to a boundary and an active selection collapses
entirely. Under the collaborative editing this pane is headed for, that is
every remote keystroke yanking the local caret out of the word being typed.

`minimalChange` is a pure function with a round-trip property plus a second,
denser generator — `fc.string()` pairs almost never share a prefix, so without
one the property would pass vacuously on whole-document replacements.

## 3. The dark preview was black text on near-black

canvas-render assigns markdown body runs no `fill` of their own, so every
`<text>` took the SVG default — black — on both themes. The preview was
present, correctly laid out, and unreadable. Light was off-spec too, rendering
pure black where the theme's label fill is `#404040`.

Fixed at the seam `editorTextFill` already existed for (until now it had no
caller): the host carries the fill and every run inherits it, so colour stays
out of the render path and exported SVG bytes remain independent of the
viewer's theme.

| before | after |
|---|---|
|  |  |

## 4. OKF core facets are editable

`writeCoreFacets`/`readCoreFacets` have existed in canvas-workspace since the
OKF bridge landed, but apps/web read or wrote neither — every canvas carried
empty core meta and `canvas_export_okf` fell back to its `type: 'canvas'`
placeholder for all of them.

title and the canvas name are ONE concept, so the single title field writes
both: the facet is the document's own truth, and the snapshot row stays the
denormalised copy the canvas list reads without loading every document. This
matches the precedence `deriveCanvasListRows` already encodes
(`coreFacets?.title ?? segment`).

`view` is deliberately not editable — its documented job is choosing between
Views when several EXTENSION facets apply, and extension facets are not
editable here, so the control would have nothing to choose between.

Two contracts drove the shape:

- `writeCoreFacets` REPLACES the stored bucket rather than merging, so every
  handler emits a whole `CanvasCoreMeta`. `facetsRaw` — root-level frontmatter
  keys this app does not model — therefore has to survive an unrelated edit,
  pinned by its own test.
- Body and facets are containers of ONE Loro document saved as a whole
  snapshot, so they must share one instance. A second hook holding its own
  `Loro` for the same canvas would export a document missing the other's
  container and silently overwrite it. `use-markdown-body` is renamed to
  `use-markdown-canvas-doc` to stop the name implying otherwise.

## Verification

Manually verified in the running app on Japanese content in both themes, and
for the facets specifically on a canvas whose body predates facets existing —
title/type/tags round-trip and the existing body is untouched.

Mutation-checked: the CSS ordering that lets `.cm-md-marker` win inside a
heading, the reconcile fix (both failing cases return), and both page-wiring
tests (removing `renameCanvas` and removing `writeCoreFacets` each fail a
distinct test).

`apps/web` jsdom 1718 and web-browser 465 pass, plus `pnpm -r typecheck`,
knip, biome and arch-lint. One browser test failed once during a post-merge
run and passed on re-run; the second run was green before the name could be
captured, so it is unidentified — flagging it rather than calling the run
clean.

## Not in scope

Spatial canvases do not get the properties bar. Their document lives inside
`useCanvasSync`'s per-connection state behind the backend's delta protocol,
with locks, undo/redo and restore alongside it; reaching it safely is its own
slice rather than a tail on this one.


kamiazya:
@coderabbitai review

PR #600, opened 2026-08-11
fix(render): break orthogonal elbow ties by bend count, not candidate order

kamiazya:
## What

Fixes the 3-bend staircase reported from a real canvas on mobile: a node below-right of its target with an edge routed top-side → right-side drew three corners where one was available.

**Root cause** (`packages/canvas-render/src/layout/spatial-edges.ts`, `bestCandidate`): the two stub-to-stub elbow candidates are always the **same Manhattan length**, so the length-only ranking was a permanent tie and the stable sort picked the arbitrary first candidate (horizontal-first) — even when the vertical-first elbow extended both stubs collinearly and needed a single corner. The side-choice optimizer counts bends; the final stub-to-stub connector never did.

**Fix**: equal-length candidates rank by **bend count** (direction changes along the polyline, collinear/repeated points ignored). Clearance still dominates — a bend-poor candidate through an obstacle never beats a clear staircase (pinned by test).

## Tests

- `edge-bend-tiebreak.test.ts` pins the repro geometry (1 bend, vertical run collinear with the source stub) and the clearance-dominates case (obstacle on the good lane → staircase wins, never crossing the obstacle). Mutation-checked: reverting to the length-only sort turns the repro test red.
- Full `canvas-render-node` (374), canvas-viewer node+jsdom (483), apps/web jsdom (1756), and the edge-rendering browser tests (8) all green — no pinned path expectation depended on the old arbitrary tie order.

## Visual repro

Before — the reported staircase (three corners for a clear L):

After — one corner, collinear with both stubs:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved orthogonal edge routing to prefer simpler, straighter paths when candidate routes have equal length.
  * Preserved obstacle clearance as the priority, preventing routes from intersecting obstacles.
  * Added fallback behavior for selecting the shortest available route.

### DIFF
--- a/packages/canvas-render/src/layout/edge-bend-tiebreak.test.ts
+++ b/packages/canvas-render/src/layout/edge-bend-tiebreak.test.ts
@@ -1,9 +1,8 @@
-// The two stub-to-stub elbow candidates always tie on Manhattan length, so
-// ranking by length alone let the arbitrary first candidate win — even when
-// the other one runs collinear with BOTH stubs and draws one bend instead
-// of three. Pinned from a real phone repro: a node below-right of its
-// target, edge top-side → right-side, drew a 3-bend staircase where a
-// single L was available.
+// The two stub-to-stub elbow candidates always tie on Manhattan length,
+// so length alone cannot rank them. The routing invariant pinned here:
+// among equal-length clear candidates, the one with fewer bends wins — in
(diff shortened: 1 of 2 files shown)


kamiazya:
Applied in c8cbda5 — the test header now states the equal-length → fewer-bends routing invariant instead of the repro history, per the source-comment discipline.

PR #623, opened 2026-08-11
docs(adr): accept ADR-0008 — slug derivation, rename, sibling uniqueness

kamiazya:
Closes the three questions [ADR-0007](https://github.com/kamiazya/whiteboard/blob/main/docs/contributing/adr/0007-canvas-identity-and-store-split.md) deliberately left open. Each is answered from a measurement rather than from the prose around it — and each turned out differently than expected.

## 1. Derivation loses non-Latin names entirely

`deriveDisplaySlug` run against realistic Japanese names (the real function, not a paraphrase):

| name | derived slug |
|---|---|
| リリース計画 2026 | `2026` |
| 構成図 | `untitled` |
| 議事録 8月 | `8` |
| 設計メモ | `untitled-2` |

Two failures: no-ASCII names collapse to mutually indistinguishable `untitled-N`, and names that merely *contain* digits keep only the digits. ADR-0007 gives the derived slug a forward-looking purpose ("a later promotion to real slugs needs no re-derivation"), which booby-traps that promotion for anyone not naming canvases in ASCII.

**Decision:** a slug is derived from creation order, never from a name. `deriveDisplaySlug` has no caller today, so this is a deletion, not a migration.

## 2. Rename is already implemented, just unwired

`branches-store.ts` carries `_renameCanvasSlug`, which updates one column — branches and versions FK on `canvasId`, the blob path uses `canvasId`, and OpenCanvas links are stored as `wikiLink { canvasId }`. It has **no caller anywhere**. Rename is unwired, not unsolved. What genuinely breaks is external URLs, which `deriveAliasHistoryRows` (currently a stub returning `[]`) is designed to redirect.

**Decision:** wire rename; make alias history real.

## 3. Sibling uniqueness cannot survive a CRDT merge

`WorkspaceTree.#assertNoSiblingConflict` runs inside `createNode`/`move` — local mutations. Two peers each create `notes` under the same parent, merge, and land a tree violating the invariant with nothing to detect it. Rejecting the merge is not an option: a merge is not an operation anyone can decline.

Google Drive is the same shape and answers by **not having the invariant** — duplicate names in one folder are legal, because identity is an opaque file id. The `(1)` suffix people associate with Drive is added by *Drive for Desktop*, at the point where files are projected onto a filesystem that does require unique siblings; stored data is never touched.

We are already in that position: the alias index PK is `(workspaceId, seq)`, not `(workspaceId, alias)`, so duplicates break no constraint, and this package's own rule is "alias is always derived, never stored".

**Decision:** duplicates are legal in storage; the derived alias disambiguates them. `#assertNoSiblingConflict` stays, demoted from invariant to courtesy — refuse a collision we can see coming, cope with one we cannot.

An earlier draft reconciled duplicates by rewriting the tree on load. It is recorded as a rejected alternative: every peer mutating a shared document as a side effect of *reading* it makes correctness rest on that rewrite being perfectly idempotent forever, and a bug there corrupts rather than merely confuses.

## Status

**Proposed**, deliberately. Identity is a product call, not a review call.

Docs-only; no code changes.

https://claude.ai/code/session_016wjXL66r2XiwYf4qMYHrGB


kamiazya (commented):
Integrator review. All five load-bearing claims verified against the code: the derivation measurement reproduces exactly (`リリース計画 2026`→`2026`, `構成図`→`untitled`, collisions suffix), `_renameCanvasSlug` has one definition and zero callers, `deriveAliasHistoryRows` returns `[]`, `#assertNoSiblingConflict` runs only in local mutations (`createNode`/`move`/`rename`), and the alias-history tables live in the workspaceIndex family. The Drive-projection framing for point 5 is sound, and deriving the disambiguation instead of rewriting the tree on read is the right call.

Two gaps before this is actionable:

1. **Points 3–4 straddle the two-store split without saying so.** Point 3 wires rename to `_renameCanvasSlug` — the **legacy workspace/slug store**. Point 4's redirect reads alias history — the **OpenCanvas/workspaceIndex world** (`workspaceIndexAliasHistory`, written via `deriveAliasHistoryRows`, which derives from the workspace *tree*, not from legacy rows). As written, a legacy-world rename records nothing the redirect can find: `deriveAliasHistoryRows` un-stubbed would derive history for v1 documents, while the URLs that need redirecting (`/canvas/:ws/:slug`) are served by the legacy world. The ADR should either (a) name a legacy-side retired-slug record (a small table or column in the workspace/slug store) as the redirect's source for legacy URLs, or (b) state that point 4 lands only with the ADR-0007 convergence step and rename ships first with a plain 404 for stale URLs — either is defensible, but the current text implies a wiring that crosses the split for free.

2. **Consequences understate the deletion's blast radius.** Removing `deriveDisplaySlug` + the browser-local `secondary` also touches: `BrowserLocalIndexPage.test.tsx` (pins `meeting-notes`/`trip-plan` secondaries and the `notes`/`notes-2` collision), `docs/assets/browser-local-list.png` + the getting-started tutorial (the committed screenshot shows the derived slugs), and ADR-0007 point 3's text (the "promotion needs no re-derivation" rationale this ADR removes should get an as-built addendum on 0007, same convention as 0004's). Worth listing so the implementing slice doesn't rediscover them.

One smaller note: point 1 keeps `untitled-N` as the universal creation-order slug — fine — but the daemon gallery's secondary line will then show `untitled-N` under a meaningful display name for every renamed-by-display-name canvas. That is honest (it IS the identity) and consistent with point 3 making it changeable; just confirming the reading order is intended: name carries meaning, slug becomes visible only as the address.

PR #750, opened 2026-08-14
feat(web): make the dock mode-independent and give hand mode double-tap zoom

kamiazya:
The bottom dock exchanged its leading slot for zoom controls whenever hand mode was armed. Two costs: a phone lost undo/redo the moment it started navigating, and the dock's width changed with the mode — 338px in hand vs 250px in select, against a 344px budget at 360px, so the dock was already at its ceiling with nothing left to add.

Navigation is a property of the viewport, not of whichever tool is armed, so nothing is exchanged any more.

## What replaces the zoom cluster

People ask to **see everything** or to **get closer**. The magnification between those two is an implementation coordinate, so the zoom in/out buttons and the `100%` readout are gone **with no replacement** (no `Shift+0`, no percentage in the minimap):

| want | keyboard (already existed) | touch (this PR) |
|---|---|---|
| see everything | `Shift+1` | **zoom-to-fit, one dock button, every mode** |
| get closer | `Shift+2` (to selection) | **double press in hand mode**, anchored on what was pressed |

The double-press slot was free: hand mode's `pointerdown` returns into a pan **before** the double-press detection that creates a note, so select mode keeps its own double press unchanged. Hand mode was already documented in the source as *"the one-handed touch navigation mode, where a second finger is not available to promote the gesture"* — this is the half of it that was missing.

## Visual repro (360px)

Hand mode, before — zoom cluster in place of undo/redo, 338px:

Hand mode, after — same dock as every other mode, 295px, one row, 33px clear each side:

After a real touch double-tap (1x → 2x, anchored on the tapped point):

## Verification

- `pnpm test --project web-browser spatial-editor` — 2034 passed
- `pnpm -r typecheck`, `pnpm knip`, `pnpm lint` clean
- Real browser, mobile touch context: dock identical across hand/select/connect (351px @390, 295px @360, single row), double-tap 1x → 2x, fit reframes
- One real bug caught by the new test while writing it: passing `onZoomToFit` as a bare handler reference handed React's click event to `frameContent(ids)`

## Tests

The two tests bound to the removed affordance are **retargeted, not left as fossils**:

- `hand mode swaps the leading history cluster for zoom controls` → now pins that the dock does **not** swap
- `zoom controls: in/out change the scale, reset returns to 100%…` → now pins zoom-to-fit in every mode **plus the absence of any percentage**, and a new pair pins the mode-split double press (hand zooms and holds the pressed point still / select still creates a note)

Pinch and two-finger pan coverage is untouched — `touch-pan-zoom.browser.test.tsx` drives the real handlers with `pointerType: 'touch'` and never went through the dock.

`HandViewControls.tsx` is deleted; its extraction in #697 is what made this a one-file removal.

https://claude.ai/code/session_012SZn2v65ZrFoCuJ1ZU45hc

## Summary by CodeRabbit

* **New Features**
  * Added a “Zoom to fit” control to the spatial editor toolbar, available in all editing modes.
  * Added double-press zooming in hand mode, centered on the pressed location.
* **Bug Fixes**
  * Improved toolbar consistency across hand, select, and connect modes.
  * Preserved history controls and toolbar sizing across editing modes.
* **Tests**
  * Added coverage for toolbar consistency, anchored zooming, and mode-specific double-press behavior.


kamiazya:
**Review found a real regression in this PR, now fixed in `fc96d24` (the third commit).**

Deleting `HandViewControls` removed three keyboard-operable buttons, and both replacements are pointer-only — the hand-mode double press is a `pointerdown` pairing, and the wheel needs a pointing device. What survived for the keyboard was `Shift+1` / `Shift+2`, which reach *content-derived fit levels* rather than a magnification someone chooses; `Shift+2` also needs a selection, and there is no keyboard node traversal to make one. So a keyboard-only or switch-access user was left with no way to reach any magnification other than "fit".

Step zoom now lives in the shortcut catalog (`+` / `-`) **and nowhere else**, which keeps this PR's actual decision intact: the only view control on screen is still "see everything", and the magnification between the two ends stays unlabelled. Unmodified `+`/`-` rather than `Cmd+=`, because that combination is the browser's own page zoom — a different function, and not ours to take.

Verified in a real browser: `+` → 1.25×, `-` `-` → 0.8×, in every mode.

PR #752, opened 2026-08-14
docs(adr): propose one writer per fact, with SQLite as a derived index

kamiazya:
ADR-0010, **Proposed** — the ownership decision Task #33 was holding for a human call.

## The finding

Three ADRs have now met the same wall: #0007 split the stores, #0008 hit it with alias history, #0009 hit it again with format. The recurring cost is **not** the split. It is that no rule says which store *owns* a fact, so the same fact gets written independently in two places — and each layer then picks a locally reasonable policy for a missing value that contradicts the other's.

Measured, from this checkout's own data:

| document | `canvases.kind` | Loro `document.kind` | nodes |
|---|---|---|---|
| `Go1G4OcJKUBu` | `null` | absent | 75 |
| `uH6qTx6Ai2hl` | `spatial` | **absent** | 92 |

The instructive part is that **no layer was wrong on its own terms**: refusing an unknown kind is right for a read that must pick a format, defaulting to `spatial` is right for a list that renders a badge.

Same shape twice more — a document's name is written to `canvases.slug` by `renameCanvasSlug` and to the tree's `segment` by `canvas-crud`, **neither touching the other** (verified). ADR-0009 already flagged that the name is stored twice.

## What it decides

1. Every fact has exactly one owner; an ownership table assigns them.
2. `canvases` becomes a **derived index** — one projection writes it, and it is rebuildable from the documents, so drift is a repair rather than a loss.
3. An unrecorded kind becomes **unrepresentable** (required at creation, column `NOT NULL`) rather than handled.
4. Unknown-because-legacy and unknown-because-a-newer-peer are separated: the first is eliminated, the second is **forward** compatibility and stays.
5. The rule is executable (a scan for writes outside the projection), not prose.

**Facts with a single home do not move.** `isPinned`, `pinOrder`, `currentBranch` and the timestamps stay exactly where they are — relocating working code for symmetry is cost with no defect behind it.

## What it rejects

**One large Loro document for everything.** On this codebase's own numbers: listing is `SELECT slug, updatedAt, kind` against `canvases_workspace_updated_idx` today and a CRDT has no `ORDER BY … LIMIT`; `compactCanvas` + the 32 MiB warn threshold + the auto-compact debouncer already exist because op-logs grow; and `withWorkspaceWriteLock` would go from per-workspace to per-product. It also would not have prevented the incident — **two writers caused it, not two stores.**

## Note on scope

Per the instruction, backward compatibility **including stored data** is out of scope, so this deletes rather than migrates: the "declare a kind on first write" branches, `DocumentContentLossError`, and `DocumentKindUnknownError`'s legacy half all become unreachable. Those guards (#746, #747) were correct for the policy in force when written — under this one they answer a question that stops being asked.

No migration order is included; that is `gh stack`'s job per `dev-flow.md`.

Docs-only. Status is **Proposed** — it is yours to accept.

https://claude.ai/code/session_01PhvRoF23tM99JwDwVJ4YHw

## Summary by CodeRabbit

- **Documentation**
  - Added ADR-0010, documenting authoritative ownership for document facts, workspace placement, and derived listing metadata.
  - Clarified that required kinds are enforced, unknown newer kinds remain errors, and index drift can be recovered.
  - Added the proposal to the architecture decision record index.

### DIFF
--- a/docs/contributing/adr/0010-one-writer-per-fact.md
+++ b/docs/contributing/adr/0010-one-writer-per-fact.md
@@ -1,0 +1,166 @@
+# ADR-0010: One writer per fact, and SQLite as a derived index
+
+**Status:** Proposed
+
+## Context
+
+A document's kind was recorded in two places that disagreed, and each layer
+applied the opposite policy to a missing value. Measured against this
+checkout's own data:
+
+| document | `canvases.kind` column | Loro `document.kind` | nodes |
+|---|---|---|---|
+| `Go1G4OcJKUBu` | `null` | absent | 75 |
+| `uH6qTx6Ai2hl` | `spatial` | **absent** | 92 |
+
+The Loro map was empty for every document that existed, including the one the
+column had an answer for. `wb_document_get` reads the Loro map and refused;
+`listCanvases` read the column and answered `spatial`; `getCanvasKind`
+invented `spatial` and restore stamped that invention onto a restored
+document's row, where it outlived the guess.
+
+The instructive part is not that a value was wrong. It is that **no layer was
+wrong on its own terms**. Refusing an unknown kind is correct for a read that
+must pick a format. Defaulting to `spatial` is correct for a list that renders
+a badge. They were both reasonable because nothing said which of them owned
+the fact.
+
+### The same shape, twice more
+
+- **A document's name.** [ADR-0009](0009-mcp-tool-naming.md) already noted it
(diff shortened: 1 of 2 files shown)


kamiazya:
Not keeping this as an ADR — the decision's value is in the code enforcing it, and the cross-ADR references are process chronology rather than something a future reader needs. Doing the cleanup directly instead.

PR #755, opened 2026-08-14
feat(render): optimize edge sides by region past the edge gate

kamiazya:
Past `CROSSING_OPT_MAX_EDGES` (200), side-choice optimization was skipped **wholesale**. That is not a performance gate anyone had measured the other side of — so this PR measures it first, then fixes it.

On a 345-edge clustered canvas: **331 avoidable-ink violations, one per edge**, against 29 across the ~4400 edges of the entire small corpus. A canvas that size is what an AI-authored document grows into, so the size where optimization stops being *affordable* is not the size where quality stops *mattering*. It was a silent quality cliff — 345 edges laid out in 12ms because nothing was optimizing them.

## Three commits, instrument first

1. **`test(render): add clustered large canvases to the routing bench`** — every routing instrument here measured one shape: `gridCanvas` wires nodes by a stride, so every edge spans the whole layout and **55%** of edge pairs survive a bounding-box test. Real documents cluster. The new `clusteredLayout` cases measure **5%** and **4%**. This distinction is load-bearing: optimising against the stride canvas alone prices spatial locality at its least favourable input and rejects it for the wrong reason. (It did exactly that to a spatial index I tried and threw away — zero win at 55%.)
2. **`test(render): pin the drawing cost of a canvas past the optimizer gate`** — the 331/42623/782 figures above, pinned exactly like the existing aggregate.
3. **`feat(render): optimize edge sides by region past the edge gate`** — the fix.

## How

`optimizeAcrossRegions` groups edges along a Morton curve through their midpoints, chunks them into even regions of at most the gate, and runs the ordinary unaligned-then-aligned search per region. Interaction is local, so two edges in well-separated regions were never going to be scored against each other anyway.

**What it gives up, plainly:** a crossing between edges assigned to different regions is never priced. Morton order puts spatial neighbours next to each other so a split falls between clusters far more often than through one, but nothing guarantees it.

## Measured

345-edge clustered canvas:

| metric | before | after |
|---|---|---|
| avoidable-ink violations | 331 | **183** |
| interior ink | 42623 | **21001** |
| border ink | 889 | **554** |
| crossings | 782 | **752** |
| bends | 633 | 658 |
| layout time | 12ms | 622ms |

**At or under the gate nothing changes, by construction** — one region, same two runs, same seed. The small-corpus pins are untouched and the 40n/40e, 60n/200e and clustered-165e benches are flat across interleaved paired runs.

**The live-drag path keeps the hard gate.** Several hundred ms is worth paying once on a committed change and never on a frame someone is dragging through; a canvas past the gate drags exactly as fast as before and picks up its regional repair on drop.

## On the region size

It reuses `CROSSING_OPT_MAX_EDGES` rather than adding a second knob. Measured curve: 200 → 183 violations, 120 → 158, 80 → 122, at 1023 → 1184 → 1481ms. Smaller regions buy quality because `TRIAL_BUDGET_EDGES` applies *per region* — the real knob is total trial budget, and region size is only how it gets spent. Tuning a new constant against one synthetic canvas would make that canvas the convention by accident, so the curve is recorded in the package rule instead.

## Also tried and rejected (measured, not argued)

- **Branch-and-bound early abort** on trial evaluation — neutral to slightly worse. Making the accumulation monotone required subtracting priors in a separate pass, which cost what the abort saved.
- **Uniform-grid broad phase** for the pair update — zero. At E ≤ 200 a `Set` costs more than the linear scan with an inlined AABB test it replaces.
- **Memoizing `interiorInkThrough`'s O(n²) containment filter** — 1% of runtime; it is called with 2 rects, not 143.

The clustered breakdown that explains all three: routes 31%, anchors 16%, setup 16%, **pair loop 10%**. No dominant hot spot, so there is no cheap 2× in this band — which is why this PR spends its effort on the cliff instead.

### DIFF
--- a/.claude/rules/package-canvas-render.md
+++ b/.claude/rules/package-canvas-render.md
@@ -375,6 +375,35 @@ paths:
       then the same loop reusing unchanged paths (4.5x). What made it
       affordable was reusing the search's OWN incremental trial machinery
       instead of writing a second scorer beside it.
+    - **Past the optimizer's edge gate the search runs over spatial REGIONS,
+      not over nothing.** `CROSSING_OPT_MAX_EDGES` (200) used to skip
+      side-choice optimization wholesale, which is not a small loss: measured
+      on a 345-edge clustered canvas, 331 avoidable-ink violations — one per
+      edge — against 29 across the ~4400 edges of the entire small corpus.
+      That is the size an AI-authored document reaches, so the size where
+      optimization stops being affordable is not the size where quality stops
+      mattering. `optimizeAcrossRegions` groups edges along a Morton curve
+      through their midpoints, chunks them into even regions of at most the
+      gate, and runs the ordinary two-run search per region. It works because
+      interaction is LOCAL: 4-5% of edge pairs survive a bounding-box test on
+      the clustered bench cases (55% on the deliberately pathological stride
+      canvas). What it gives up, stated plainly: a crossing between edges in
+      two different regions is never priced. Result on that canvas —
+      violations 331 -> 183, interior ink 42623 -> 21001, border ink
+      889 -> 554, at 12ms -> 622ms. Nothing at or under the gate changes, by
+      construction (one region, same two runs, same seed).
+      Region size is `CROSSING_OPT_MAX_EDGES` itself rather than a second
+      knob, deliberately. The measured curve on that canvas is 200 -> 183
+      violations, 120 -> 158, 80 -> 122, with time rising 1023 -> 1184 ->
+      1481ms: SMALLER regions buy quality, because `TRIAL_BUDGET_EDGES`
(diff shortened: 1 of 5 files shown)


kamiazya:
Note on visual evidence: no before/after figure is attached. A 345-edge canvas is not legible as a side-by-side — both panels read as a mesh of red lines, and the one I generated showed nothing a reviewer could act on. The evidence for this change is the exactly-pinned quality block added in the preceding commit (violations 331 → 183, interior ink 42623 → 21001), which is what the routing scoreboard exists for.

Per the `visual-evidence` skill: do not manufacture a picture when the change is not legible as one. Rendering a smaller canvas would not help either — under the gate this change does nothing by construction.

PR #761, opened 2026-08-14
feat(ports): add the DocumentIndex port and single-source the path rule

kamiazya:
Bottom layer of the convergence stack (`gh stack init`, so `gh stack merge --squash` can land the layers atomically later).

## What this is

A workspace's index of what it holds — which paths exist, what each one is, which stored document each names — declared in `canvas-ports` because `server-core` cannot import `mcp-server` to reach the `canvases` table.

**Hierarchy is a path, not a parent pointer.** Segments joined by `/`, the way a filesystem stores paths rather than a tree. Sibling uniqueness falls out of path uniqueness instead of needing a walk, lookup is an index hit, and moving a document is a prefix rewrite over its descendants.

That is not a new idea imposed on the codebase — it is what was already there and unused:

| already true | where |
|---|---|
| `resolveAlias` returns `segments.join('/')` | `canvas-workspace` |
| `findByAlias` resolves a path back to a node | `canvas-workspace` |
| `validateSlug` splits on `/` and validates per segment | `mcp-server` |
| `unique(workspaceId, slug)` | the `canvases` schema |

## The part worth reviewing

Writing `documentPathSchema` surfaced that its segment rule **already existed** as `mcp-server`'s `SAFE_SLUG_SEGMENT` — character for character. Copying it into the shared layer would have created exactly the two-writers-one-fact shape this whole line of work exists to remove.

So the rule now lives once, in `canvas-model`, following the precedent `workspaceIdSchema` set for codifying a runtime validator. `validateSlug` keeps only what a single schema message cannot do: explain a rejection *per cause* (whitespace / `.` / leading hyphen / …).

A property pins that the two accept the same strings, with a generator dense enough to reach separator abuse (leading, trailing and doubled `/`) that a plain string arbitrary would essentially never produce. **Mutation-checked**: letting the schema tolerate an empty segment fails it at the counterexample `"/"`.

## Reach

`userReach`: **`foundation:`** — `DocumentIndex` has no implementation in this PR. It is wired by the layer above, which moves the MCP tools onto slug addressing so an agent-created document appears in the user's list. That layer is the next PR in this stack, not a promise.

The `TOKENS` registry guard test needed updating from three keys to four; that test exists to make exactly this kind of addition deliberate.

## Verification

`canvas-ports-node` + `canvas-model-node`: 262 pass. `mcp-node`: 3323 pass. `pnpm -r typecheck` and `pnpm knip` clean.

https://claude.ai/code/session_01PhvRoF23tM99JwDwVJ4YHw


kamiazya:
@coderabbitai review

PR #762, opened 2026-08-14
feat(web): lay committed canvases out in a worker instead of freezing the UI

kamiazya:
Layout is the editor's longest synchronous block, and it is not reserved for large canvases. Measured here:

| canvas | layout |
|---|---|
| 6 nodes / 6 edges | 10.7ms |
| **20 nodes / 30 edges** | **81ms** |
| 40 / 100 | 204ms |
| 60 / 200 | 339ms |

An ordinary canvas is already past the ~50ms a person notices, and it is paid again on **every committed change** — each node added, each drag dropped. Moving layout off the main thread does not make it faster; it stops it freezing the UI, which is the part a user actually experiences.

## Two gates, measured before writing any of it

**Text metrics.** A worker has its own `FontFaceSet`, so the vendored Roboto face has to be registered there or `measureText` silently falls back to a system font and every wrapped line lands somewhere else — the divergence class canvas-viewer's `font.ts` already documents having shipped once. Verified identical in Chromium, and pinned by a parity test asserting the worker's **SVG, bounds and scene deep-equal** the main thread's.

**Transport.** `structuredClone` of a 60/200 scene costs 0.90ms against 339ms to lay it out. Encoding to a transferable byte array instead saves 0.4ms of that — real, and 0.12% of the work — so the simplest transport wins. It also preserves `undefined`-valued keys and Maps that a JSON round trip would quietly drop.

## What could not move, and what that cost

canvas-codec's remark dependency does not evaluate in a module worker, so **markdown is parsed on the main thread** and the mdast travels as plain data. Parsing is 15–19ms of the block (4–19%), so blocking goes **81ms → 19ms** and **339ms → 15ms** rather than to zero — and the worker chunk carries no remark/unified at all. `renderCanvasToSvg`'s composition is extracted to `scene-render-core.ts` to keep the codec out of that import graph; it is still one composition, not a second scene builder.

## Correctness by falling back, never by guessing

A host-supplied function seam cannot cross `postMessage`, a body nobody pre-parsed cannot be invented, and a worker that fails to start or throws has no scene. All three lay the same canvas out synchronously instead. **A degraded worker costs responsiveness, never content.**

## Where the behaviour changes, and where it deliberately does not

Going async is not free: the scene arrives a frame later. That is an excellent trade against a 300ms freeze and a bad one against a 10ms pause, so offloading starts at **12 elements** — drawn at the perceptibility line (10.7ms at 6 nodes, ~41ms at 12, 66–81ms at 20), not anywhere convenient. Below it, and on the live-drag path (which already has its own carried-side cache), behaviour is bit-identical to before.

The threshold is *also* why the existing suite stayed green, which is exactly the kind of coincidence worth distrusting — so the wiring gets a test that observes the `Worker` being constructed.

## Verified, not felt

The claim is "the UI does not freeze", so that is what is asserted: animation frames counted across a committed change on a 45-node/90-edge canvas. Mutation-checked by disabling offloading — **2 frames painted on the synchronous path against a bar of 5**.

An earlier version of that test asserted "the DOM has not caught up yet" instead. Its mutation check **passed with offloading disabled** — React schedules the synchronous re-render too, so the assertion held either way and proved nothing. Recorded in the test.

## Two things this got wrong first, both caught by existing guards

- Hand-listing the seams in a dependency array dropped the zoom-driven LOD gate, and an expanded embed stopped collapsing — caught by the embed browser tests. The hook now takes the memoized seam object whole.
- Two heavy browser tests raised the flake rate of the whole parallel project: three consecutive runs failed a *different* unrelated test each time. Folded into one test asking both questions of the same edit; two consecutive full browser-project runs green afterwards.

`purity-guard`'s seam check now states its property rather than counting one spelling, since the seams are spread at the synchronous call sites and passed whole at the worker one.

Full monorepo suite green.


kamiazya:
## Update: the font risk is verified on all three engines, and wired to refuse anyway

The PR body flagged non-Chromium `FontFaceSet` support in workers as unverified. That was laziness rather than a limit — Playwright drives WebKit and Firefox too.

**Ran the parity test on all three engines. WebKit and Firefox both register the vendored Roboto face in a worker and measure identically to their own main thread.**

The result is not vacuous: the test asserts the MAIN thread reports `'loaded'` before comparing scenes, so "both fell back to a system font and agreed with each other" cannot pass it.

Playwright's WebKit is still not Safari, so the refusal path is now wired rather than assumed:

- a worker whose font load degrades answers `failed` with `FONT_DEGRADED` instead of a scene
- the hook lays that canvas out on the main thread and **latches** — a realm that cannot load the face will not later, so re-spawning a worker per edit would be waste
- mutation-checked: without the latch a second worker is constructed

This closes the one item in the PR that could have changed the picture rather than just the speed. An earlier version of the fallback test asserted a constant equalled itself; it was deleted rather than committed.

PR #787, opened 2026-08-15
feat(web): give the shared worker's replica the daemon write

kamiazya:
Third increment of the Loro-authority work, after #779 and #783. This is the one that reaches a user.

## What changes for someone using the app

**A sibling tab's edit now arrives at worker speed instead of after a daemon round trip.** That is the whole point of holding a replica in the worker, and until this PR nothing consumed it.

## The shape

`push(doc, update)` joins `SseStreamSource`, because *which* authority a document has is exactly what the two implementations disagree about — and the one thing a caller should not have to know.

- `SseStreamHub.push` is the POST `SseBackend.pushLocalUpdate` used to make, **moved rather than copied**. With no SharedWorker the daemon IS the authority and nothing changes.
- The SharedWorker-backed source hands the bytes to the replica instead. The replica merges them in arrival order against everything else it knows — the other tabs' work and the daemon's — and writes onward itself.

That ordering is the reason for the move. A tab posting its own edits cannot know what the daemon has already told the worker, so it re-sends state that just arrived. The replica diffs against what it holds, and the existing empty-diff return drops the echo before it can loop — which is also what makes *"tab work goes to the daemon, daemon work does not"* safe to state as one line.

The tab-side source also stops ignoring `authority-update` and delivers it as inbound state. Both channels are delivered during the transition; Loro absorbs the overlap.

## Why the contract suite earned its keep again

Two implementations now reach the daemon by different routes, and a caller cannot tell which it holds. If either stopped arriving, the failure would look exactly like success: the edit applied locally, shown to the user, echoed to sibling tabs, and **never persisted** — visible only after a reload.

So the push case goes in the shared contract, run against both harnesses. It pushes real Loro bytes and reconstructs state rather than comparing bytes, since an implementation that merges through a replica sends that replica's delta, not the bytes it was handed.

## Deletions that came with it

- `toBase64` joins `fromBase64` as an export — the worker and the tab source each had a copy.
- `canvasUpdateUrl` holds the `${workspaceId}/${slug}` reconstruction the hub and the worker both need. That the document key IS the workspace-and-slug pair is the only reason anything holding just the key can address the route, and it is now said once.

## Verification

- **Mutation-checked twice.** Dropping the write turns the worker case red. Writing *unconditionally* also turns it red — the daemon's own frame becomes the first write, which is precisely the loop the `from !== null` guard prevents. The test discriminates both failure modes, not just absence.
- The jsdom worker case asserts the `Authorization` header on the real request. `daemon-auth-seam.test.ts` scans source to prove no *other* place assembles that header; only an actual request proves the one place that should, does — and an unauthenticated write is dropped by the daemon, which looks identical to a tab whose edits never persist.
- `mcp-node`: 3360 passed. `web-jsdom` + `web-browser`: 538 passed. `arch-lint-node`: 76 passed. knip clean.

## Known ceiling, marked in source

A delta dropped by a failed write is still never re-sent, and centralising the write means one failure now costs every tab's edit rather than one tab's. The old shape did not re-send either — this concentrates an existing gap rather than creating one. A resend queue keyed on the replica's version is the upgrade path, and it is tracked.

### DIFF
--- a/apps/web/src/lib/sse-shared-stream-source.contract.test.ts
+++ b/apps/web/src/lib/sse-shared-stream-source.contract.test.ts
@@ -24,6 +24,7 @@ const BASE = 'http://127.0.0.1:3099'
 // separate one case's traffic from another's.
 let streamOpens = 0
 const openedStreamIds: string[] = []
+const daemonWrites: { doc: string; body: Uint8Array }[] = []
 const subscribeBodies: { subscribe?: string[]; unsubscribe?: string[] }[] = []
 const controlMessages: { streamId: string; doc: string; message: unknown }[] = []
 let pushFrame: ((frame: string) => void) | null = null
@@ -50,6 +51,16 @@ const server = setupServer(
   }),
   http.post(`${BASE}/api/sync/subscribe`, async ({ request }) => {
     subscribeBodies.push((await request.json()) as (typeof subscribeBodies)[number])
+    return HttpResponse.json({ ok: true })
+  }),
+  // The canvas update route: where a push lands, after the worker's replica
+  // has merged it. Addressed by workspace and slug, which the worker
+  // reconstructs from the document key it routes everything else by.
+  http.post(`${BASE}/api/canvas/:workspaceId/:slug/update`, async ({ request, params }) => {
+    daemonWrites.push({
+      doc: `${String(params.workspaceId)}/${String(params.slug)}`,
+      body: new Uint8Array(await request.arrayBuffer()),
+    })
     return HttpResponse.json({ ok: true })
   }),
   http.post(`${BASE}/api/sync/message`, async ({ request }) => {
@@ -83,6 +94,7 @@ function createHarness(): SseStreamSourceHarness {
     unsubscribedDocs: () => subscribeBodies.flatMap((b) => b.unsubscribe ?? []),
     controlMessages: () => controlMessages,
     openedStreamIds: () => openedStreamIds,
+    daemonWrites: () => daemonWrites,
     ready: async () => {
       await vi.waitFor(() => {
         if (pushFrame === null) throw new Error('stream not open')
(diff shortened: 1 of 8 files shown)


kamiazya:
## Manual verification — and a finding worth more than the verification

Two real tabs, one real daemon, one real SharedWorker.

1. Moved a node in tab A.
2. **Tab B showed the move.**
3. **Tab A's own `performance.getEntriesByType('resource')` contained no `/update` POST and no `/api/sync/` request at all.** That is the discriminator: before this PR the page made that POST itself. Empty means the write left from the worker.
4. Reloaded tab B — the move survived, so the daemon persisted it. (A reload still rebuilds from the daemon's export flow, not from the replica, so this reads the daemon's state rather than the worker's.)

### The finding: this path is unreachable in local development

`selectCanvasTransport` returns SSE only for `pageSecure && !daemonSecure`, and `isSecure` is literally `protocol === 'https:'`. Local dev serves the app on `http://localhost:5173`, so **every local session picks `websocket` and the SharedWorker is never constructed.** The entire authority replica — all three PRs — has no local runtime.

That is correct behaviour, not a bug: an http page has no mixed-content problem opening `ws://`, so WebSocket is the right choice there. But it means the feature's only real runtime is *the hosted https app talking to a local http daemon*, and neither `pnpm dev` nor the PR preview reproduces it (the preview comment notes daemon mode is origin-locked to production).

So verifying it takes deliberate setup. I tried serving the dev server over https behind a self-signed cert; Chrome's interstitial is a privileged error page that browser automation cannot click through, and adding the cert to the trust store is a security-settings change I won't make unasked. What worked was **temporarily forcing `selectCanvasTransport` to return `'sse'`** — one line, reverted, and a line this PR does not otherwise touch. Everything downstream of it (SharedWorker construction, replica, push, daemon write, cross-tab fan-out) was production code.

Worth knowing for anyone verifying this area later, and worth considering whether a dev-only transport override belongs in the app rather than in a patch someone has to reinvent.

PR #806, opened 2026-08-15
feat(sync): make the SSE pipeline unidirectional through the worker replica

kamiazya:
## What

Task #40 of the worker-first architecture direction: the data pipeline between daemon, SharedWorker and tabs becomes unidirectional, and the remaining tab-side daemon round trip moves off the main thread into the worker.

- **Retire the raw per-port `update` relay.** A daemon frame now reaches a tab only as an `authority-update`, after the worker's replica has ordered and deduplicated it against every sibling tab's pushes — one inbound channel, every tab observes the same sequence. The protocol's `update` event is deleted. (The relay was the documented transition path while clients moved; delivering both meant every daemon frame crossed each port twice.)
- **Seed the worker replica with pre-existing state.** The SSE subscribe route registers routing and seeds nothing, so a replica fed by the stream alone held only history since the worker booted — and answered `snapshot-request` with silently partial truth. The worker now fetches the daemon snapshot once per origin+document (off the main thread), remembers a 404 as "empty IS the state", and retries only transport failures.
- **`snapshot(doc)` joins the `SseStreamSource` contract**, run against both implementations. The hub asks the daemon; the worker-backed source asks the replica — a second tab opens a document with no daemon round trip. `SseBackend` seeds through the source instead of its own inline fetch.
- **Fix a first-open deadlock the contract caught:** the tab source resolved snapshot replies only for documents that already had listeners, but the backend snapshots *before* subscribing. The contract case now requests the snapshot pre-subscription (the backend's real order) and went red against the bug before the fix.

## Test evidence

All layers green on the final code:

| suite | result |
|---|---|
| `pnpm --filter @kamiazya/whiteboard-web test` | 2146 + 75 passed |
| `pnpm test --project mcp-node` | 3360 passed |
| `pnpm run test:browser` | 555 passed |
| `pnpm -r typecheck` / `pnpm knip` | clean |

Mutation checks on the seed: no-op `ensureSeeded` → both seed tests red; drop the seeded-memoization → exactly the 404-once test red.

Test-infra note: `@vitest/web-worker` runs workers in the host realm and loro's WASM tolerates exactly one cleanly-initialised worker per realm, with the winner decided by import timing. With the relay gone, a degraded worker delivers *nothing*, so `sse-shared-worker.test.ts` now pins one module-scope replica worker whose loro import is awaited before any other worker exists; delivery assertions all drive that port, scoped to test-minted doc keys. Contract negative cases push real Loro bytes so a replica-backed source can't pass them vacuously (a replica drops garbage regardless of routing).

Real-browser verification is pending (no reachable browser from this machine right now) and will be done before merge.

### DIFF
--- a/apps/web/src/lib/sse-shared-stream-source.contract.test.ts
+++ b/apps/web/src/lib/sse-shared-stream-source.contract.test.ts
@@ -25,6 +25,7 @@ const BASE = 'http://127.0.0.1:3099'
 let streamOpens = 0
 const openedStreamIds: string[] = []
 const daemonWrites: { doc: string; body: Uint8Array }[] = []
+const daemonState = new Map<string, Uint8Array>()
 const subscribeBodies: { subscribe?: string[]; unsubscribe?: string[] }[] = []
 const controlMessages: { streamId: string; doc: string; message: unknown }[] = []
 let pushFrame: ((frame: string) => void) | null = null
@@ -52,6 +53,17 @@ const server = setupServer(
   http.post(`${BASE}/api/sync/subscribe`, async ({ request }) => {
     subscribeBodies.push((await request.json()) as (typeof subscribeBodies)[number])
     return HttpResponse.json({ ok: true })
+  }),
+  // The daemon's snapshot route: what the worker seeds its replica from on
+  // first touch of a document. Pre-existing content lives here.
+  http.get(`${BASE}/api/canvas/:workspaceId/:slug/snapshot`, ({ params }) => {
+    const doc = `${String(params.workspaceId)}/${String(params.slug)}`
+    const bytes = daemonState.get(doc)
+    if (!bytes) return HttpResponse.json({ title: 'Canvas not found' }, { status: 404 })
+    return new HttpResponse(bytes.slice().buffer as ArrayBuffer, {
+      status: 200,
+      headers: { 'Content-Type': 'application/octet-stream' },
+    })
   }),
   // The canvas update route: where a push lands, after the worker's replica
   // has merged it. Addressed by workspace and slug, which the worker
@@ -95,6 +107,7 @@ function createHarness(): SseStreamSourceHarness {
     controlMessages: () => controlMessages,
     openedStreamIds: () => openedStreamIds,
     daemonWrites: () => daemonWrites,
+    seedDaemonState: (doc, bytes) => daemonState.set(doc, bytes),
     ready: async () => {
       await vi.waitFor(() => {
         if (pushFrame === null) throw new Error('stream not open')
--- a/apps/web/src/lib/sse-shared-stream-source.ts
+++ b/apps/web/src/lib/sse-shared-stream-source.ts
@@ -38,6 +38,14 @@ export function createSharedSseStreamSource(
   }
 
   const listeners = new Map<string, Set<DocListener>>()
+  /**
+   * Tabs waiting on a snapshot, keyed by doc. Keyed rather than queued
+   * because the worker's reply carries the doc and nothing else — two
+   * concurrent requests for the same document share whichever state the
+   * replica holds when it answers, which is exactly what they would have
+   * gotten asking separately.
+   */
+  const snapshotWaiters = new Map<string, Set<(bytes: Uint8Array) => void>>()
   const port = worker.port
 
   // A module worker that fails to LOAD does not throw above — construction
@@ -67,15 +75,24 @@ export function createSharedSseStreamSource(
     const parsed = sseWorkerEventSchema.safeParse(e.data)
     if (!parsed.success) return
     const evt = parsed.data
+    if (evt.type === 'snapshot') {
+      // Answered ahead of the listener gate below: the backend snapshots
+      // BEFORE it subscribes — the seed must exist before the stream's deltas
+      // land on it — so at reply time the document routinely has no listener
+      // yet, and gating this on one would deadlock every first open.
+      const waiters = snapshotWaiters.get(evt.doc)
+      if (!waiters) return
+      snapshotWaiters.delete(evt.doc)
+      const bytes = fromBase64(evt.snapshot)
+      for (const resolve of waiters) resolve(bytes)
+      return
+    }
     const set = listeners.get(evt.doc)
     if (!set) return
-    // The two inbound channels, deliberately both delivered. `update` is a raw
-    // daemon frame; `authority-update` has been through the worker's replica,
-    // so it also carries what a SIBLING TAB did — which never reaches the
-    // daemon and back in time to be useful, and is the whole point of the
-    // replica. Loro merges are idempotent, so anything that arrives on both is
-    // absorbed rather than double-applied.
-    if (evt.type === 'update' || evt.type === 'authority-update') {
+    // The ONE inbound channel. Every byte a tab applies has been through the
+    // worker's replica — ordered and deduplicated against the daemon's frames
+    // and every sibling tab's pushes — so all tabs observe the same sequence.
+    if (evt.type === 'authority-update') {
       const bytes = fromBase64(evt.update)
       for (const l of set) l.onUpdate(bytes)
       return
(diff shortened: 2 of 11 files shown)


kamiazya:
## Manual verification instructions (for an agent/dev with a real browser)

This PR needs its manual-verification step done on a machine where a browser can reach a locally running daemon (e.g. the macOS machine). Please run the steps below against **this branch (`one-pipeline`)** and report results as a comment here.

### Setup

```bash
git fetch origin one-pipeline && git checkout one-pipeline
pnpm install
pnpm mcp:http:dev            # daemon on http://127.0.0.1:3099 (main checkout)
pnpm --filter @kamiazya/whiteboard-web dev   # Vite dev server, in a second terminal
```

The SSE path (and the SharedWorker replica behind it) is unreachable in plain local dev — `selectCanvasTransport` picks WebSocket for an http page. Pin the transport from the page console, then reload:

```js
localStorage.setItem('whiteboard:dev-transport', 'sse')
```

Open the Vite dev URL, pair with the daemon, and open a canvas **that already has content** (create one via MCP or the app beforehand if needed).

### What to verify

1. **Seed (the new behavior)**: opening a canvas with pre-existing content over SSE shows that content. In the Network tab you should see exactly **one** `GET /api/canvas/<ws>/<slug>/snapshot` for the document — made by the SharedWorker, not repeated per tab.
2. **Second tab, no daemon round trip**: open the same canvas in a second tab. Content appears, and **no additional** snapshot GET is made (the worker answers from its replica).
3. **Cross-tab sync**: draw/add a node in tab A → it appears in tab B without reload. This now travels **only** as `authority-update` through the worker replica (the raw `update` relay is gone).
4. **Persistence**: reload tab A → the edit is still there (replica wrote it through to the daemon; check a `POST /api/canvas/<ws>/<slug>/update` fired after the edit).
5. **External change**: while a tab is open, mutate the canvas from outside (e.g. MCP `wb_node_add` against the daemon) → the change appears in the open tab without reload.
6. **Unknown document**: open a brand-new canvas — it must open empty and editable (the worker remembers the daemon's 404 as "empty is the state"; only one snapshot GET even after several opens).

### Caveats

- Service worker: if this app origin was opened before, clear the SW or accept the update toast first — otherwise you are verifying an old bundle (`navigator.serviceWorker.getRegistrations()`; `controlled: true` means suspect).
- If something fails, please attach the console + the `/api/sync/*` and `/api/canvas/*` request log for the failing scenario.

CI is fully green; this manual pass is the last gate before merge.

kamiazya:
**Addendum after the internal review pass** (commit `58b380ea`): an adversarially-verified HIGH finding was fixed — the worker forgot nothing once seeded, so a daemon-side write landing while **no** tab held the document was never reflected on reopen. The seed is now forgotten when the last subscriber releases the document.

Extra verification scenario (please include):

7. **Write in the gap**: open a canvas → close **all** tabs of it (navigate away) → mutate it from outside (MCP `wb_node_add`) → reopen. The external change must be visible, and the Network tab shows a **second** snapshot GET for that document on reopen.

kamiazya:
**Simpler alternative for the setup above**: this PR also has a Pages preview (see the `cf-pages-preview` comment). An **https** page talking to a plain-http local daemon selects the SSE transport *natively* — no `localStorage` override needed, and it is exactly the hosted-app runtime this feature exists for:

```bash
git fetch origin one-pipeline && git checkout one-pipeline && pnpm install
pnpm mcp:http:dev     # daemon on http://127.0.0.1:3099
```

Then open the preview URL, pair with the local daemon, and run scenarios 1–7 as written. (First visit of the preview origin has no stale service worker to worry about; on a revisit, clear it per the instructions.)

kamiazya:
## Manual verification — done (real Chrome on the dev machine, Pages preview + local daemon over SSE)

The Mac pass wasn't needed after all: headless Chromium on the dev machine reaches both the **https preview** and the **local daemon (3701, this branch's code)**, which is exactly the runtime this transport exists for — https page + http daemon selects SSE natively, no override. Driven by a scripted end-to-end pass (pairing consent included), against a canvas created through the app's own slug routes:

| scenario | result |
|---|---|
| Note committed in tab A renders locally | PASS |
| **Persistence**: reload of A still shows it (the worker replica owns the write-through) | PASS |
| **Seed**: a *fresh browser profile* (fresh SharedWorker) opens the canvas and sees the pre-existing content | PASS |
| **Live delivery**: A's next edit appears in profile B **without reload** (daemon frame → replica → `authority-update`; the raw relay is gone) | PASS |
| **Gap-write** (the review HIGH fix, end-to-end): B releases the document (navigates home, worker survives), A writes, B reopens → the gap edit is visible | PASS |

Console clean in both profiles throughout (one benign CSP img warning from the pairing page favicon).

Final states — profile A (author) and profile B (fresh seed + live + gap):




Side discovery (pre-existing on main, unrelated to this PR, filed in the local backlog): a document created via **MCP** (`wb_document_create` + `wb_node_add`) reads back as an **empty** doc through `GET /api/canvas/:ws/:slug/snapshot` while `wb_scene_digest` shows its content — so MCP-written canvases render empty-but-"Synced" in the web app. Overlaps the `legacy-id-rows` reconciliation.

kamiazya:
## Manual verification result — 2 of 6 verified, 4 blocked (blocker is NOT this PR)

Ran on macOS, Chromium via Playwright. Web app from a clean worktree at `main` `30f163c0`; daemon = the repo's dev daemon on `127.0.0.1:3099` (checkout `77d72b46`, which contains this PR as `02c4e103`). Transport pinned with `localStorage['whiteboard:dev-transport'] = 'sse'`, control runs with it removed.

This PR merged at 09:23 while the pass was running, so this is a post-merge report.

### Verified over SSE ✅

**3. Cross-tab sync.** Two tabs on the same daemon canvas; a node added in tab A appeared in tab B with no reload.

**4. Persistence.** A node added over SSE survived a full page reload, and was then visible to a *WebSocket* session on the same canvas — so it reached durable storage, not just the worker replica.

### Blocked ❌ — checks 1, 2, 5, 6

**A daemon canvas that already has content opens EMPTY.** Seeded via MCP (`wb_document_create` + 3× `wb_node_add`, confirmed by `wb_document_get`), then opened in the app for the first time: nothing renders, including after zoom-to-fit. Status pill says **Synced**.

The two sides then stay permanently divergent:

- `GET /api/workspaces/default/canvases/<slug>/branches` → one branch `main`, `"tipFrontiers": ""`, `createdAt` equal to the moment of first web open. The CRDT branch is created **empty**, ignoring the document that already exists.
- Edits made in the app never appear in `wb_document_get` (still exactly the 3 seed nodes, minutes later).
- Nodes added via MCP against an open tab never appear in it — **check 5** fails for the same reason.

Since there is no seeded content to fetch, the snapshot-GET counting in checks 1, 2 and 6 has nothing to observe. (Separately: those requests come from the SharedWorker, which does not surface on Playwright's page/context request feed, so counting them needs a daemon-side log rather than DevTools-equivalent capture.)

### Why I do not think this is yours

I ran the control: a **second** freshly MCP-seeded canvas, opened for the first time over the **default WebSocket transport**. Identical — empty canvas, empty frontiers. Transport-independent, so it is not the SSE pipeline this PR rewrote.

It reads as a gap between what `wb_node_add` writes and what the editor loads on first open, and it is worth its own issue — for an MCP-first product, "the agent made you a canvas, and opening it shows nothing" is close to the worst shape a bug can take. Happy to file it.

### Two setup snags in the instructions, for next time

- The route in the instructions (`/canvas/:ws/:slug`) 404s to the not-found page. The real shape is **`/w/:workspaceId/canvas/:slug`** (`app-routes.ts`).
- **Pairing never offered itself.** On `http://localhost:5174` the daemon was reachable and CORS-clean from page script (`/api/runtime/ping` → 200 with identity), but `DaemonDetectedBanner` never appeared on either the canvas list or a browser-local canvas, and Settings → Connections showed only "Local daemon — Not connected" with no affordance. I had to start the flow by calling `beginPairingGrant` directly; from there `/pair` → Approve → redirect worked perfectly. If the auto-probe is meant to fire on a loopback origin, that is worth a look on its own.

Screenshots for each state are on the verifying machine under `tmp/screenshots/806-*.png` — say the word and I will attach them.

PR #807, opened 2026-08-15
perf(web): parse markdown inside the layout worker, not on the main thread

kamiazya:
## What

First half of the worker-first direction's #39: the per-commit markdown parse (15–19ms of an 81–339ms layout block) moves off the main thread into the layout worker, so an offloaded commit costs the main thread only a postMessage. The pre-parsed `bodies` wire field and its `failed: no pre-parsed body` path are deleted.

## The root cause that made this possible

The documented blocker — "remark refuses to evaluate in a module worker" — was never remark and never the dev server. `decode-named-character-reference` (micromark's entity decoder, deep in the remark graph) maps its **`browser` export condition** to a build that calls `document.createElement('i')` **at module top level**, so any worker chunk containing remark threw `document is not defined` before handling a single message — dev and production alike. The package already ships a DOM-free `worker` condition; Vite just resolves `browser` first even for worker chunks.

Fix: `vite.config.ts` / `vitest.browser.config.ts` pin the DOM-free entry via `createRequire().resolve()` (Node's conditions land on exactly that build), with the package added as a direct devDependency so the resolve is legal under pnpm's strict `node_modules`.

## Test evidence

- New parity case: a markdown body with a named character reference (`&amp;`), **no pre-parsed bodies on the wire**, lays out in the worker with byte-equal SVG to the main thread. Mutation-checked: removing the alias turns it red with `Uncaught ReferenceError: document is not defined`.
- Existing parity cases (deep-equal scene, file-label seam) updated to the body-less protocol and green.
- Full suites on this branch: jsdom 2143, node 75, browser 556, `pnpm -r typecheck`, `pnpm knip` — all green.

## Measured

- Critical-path JS unchanged: 118.9 KB gzip before and after (bundle-size gate green).
- The lazily-loaded layout-worker chunk grows 119 → 316 KB raw carrying remark; it is warmed at editor mount, off the interaction path.

The second half of #39 (drag start/commit render promotion) is a separate increment.

### DIFF
--- a/apps/web/mcp-source-alias-coverage.test.ts
+++ b/apps/web/mcp-source-alias-coverage.test.ts
@@ -4,6 +4,7 @@ import viteConfig from './vite.config.js'
 import vitestBrowserConfig from './vitest.browser.config.js'
 import vitestConfig from './vitest.config.js'
 import vitestDocsSnapshotsConfig from './vitest.docs-snapshots.config.js'
+import { workerSafeDepsAlias } from './worker-safe-deps-alias.js'
 
 // Each config below independently resolves `@kamiazya/whiteboard-mcp/*`
 // subpaths to source (see mcp-source-alias.ts). A subpath added to
@@ -22,6 +23,20 @@ describe('mcpSourceAlias coverage across apps/web configs', () => {
     // factory/promise form leaves `alias` empty here, which fails red.
     const alias = (config.resolve?.alias ?? {}) as Record<string, unknown>
     for (const [key, value] of Object.entries(mcpSourceAlias)) {
+      expect(alias[key]).toBe(value)
+    }
+  })
+
+  // Same drift class, worse failure mode: the browser config's copy is the one
+  // every test exercises, so a vite.config.ts that lost the pin would ship a
+  // production worker chunk that throws `document is not defined` at
+  // evaluation while the whole suite stays green.
+  it.each([
+    ['vite.config.ts', viteConfig],
+    ['vitest.browser.config.ts', vitestBrowserConfig],
+  ] as const)('%s aliases every workerSafeDepsAlias key', (_name, config) => {
+    const alias = (config.resolve?.alias ?? {}) as Record<string, unknown>
+    for (const [key, value] of Object.entries(workerSafeDepsAlias)) {
       expect(alias[key]).toBe(value)
     }
   })
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -68,6 +68,7 @@
     "@vitejs/plugin-react": "^6.0.5",
     "@vitest/browser-playwright": "catalog:",
     "@vitest/web-worker": "^4.1.10",
+    "decode-named-character-reference": "^1.3.0",
     "fake-indexeddb": "^6.2.5",
     "fast-check": "^4.7.0",
     "msw": "^2.15.0",
--- a/apps/web/src/components/spatial-editor/use-worker-scene.ts
+++ b/apps/web/src/components/spatial-editor/use-worker-scene.ts
@@ -103,6 +103,9 @@ export function useWorkerScene(
   const workerRef = useRef<Worker | null>(null)
   /** Latched: a realm that cannot load the face never will within a session. */
   const fontDegraded = useRef(false)
+  /** Latched on a worker `error` event: module evaluation that failed once
+   * fails identically on every reconstruction, so the session stays sync. */
+  const workerDead = useRef(false)
   const requestRef = useRef(0)
   const [rendered, setRendered] = useState<RenderedCanvas>(renderNow)
   // The inputs the currently-displayed scene was built from, so a scene that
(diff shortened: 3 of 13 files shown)


kamiazya:
## Manual verification instructions (for an agent/dev with a real browser)

Please verify on this branch (`codec-in-worker`) alongside #806's pass and report results here.

### Setup

```bash
git fetch origin codec-in-worker && git checkout codec-in-worker
pnpm install
pnpm --filter @kamiazya/whiteboard-web dev
```

### What to verify

1. Open a spatial canvas with **12+ nodes** (the offload threshold) where some text nodes contain markdown — include at least one named character reference, e.g. a node whose text is `# Ampersands &amp; entities` and some `*emphasis*`.
2. Edit/add a node and commit: the re-laid-out scene must render the markdown correctly (heading, emphasis, `&` decoded) — this now parses **inside the layout worker**.
3. Console must stay free of `document is not defined` / worker errors (that exact error is what the `decode-named-character-reference` alias fixes; its absence is the point).
4. Sanity: a small canvas (<12 elements, synchronous path) renders the same markdown identically.

CI + the parity browser suite (byte-equal SVG worker vs main thread) are green; this is the last gate before merge.

kamiazya:
## Manual verification — done (Pages preview, real Chrome)

Ran on the preview after `e806399e` deployed, with the service worker cleared first:

1. **The worker now actually engages on a real page.** Before `e806399e`, a 13-node canvas on `/local/...` never fetched the layout-worker chunk — `useCanvasFileSeams` supplies the file seams unconditionally, so `canLayoutInWorker`'s presence check was false for every production canvas and the offload shipped in #762 was reachable only from tests. After the fix, the mount warm-up fetches `layout-worker-BwH9CXUK.js` (resource timing), which is the gate passing on the real page.
2. **Markdown parses inside the worker on commit.** Dragging a node on the 13-node canvas re-laid out through the worker; the node containing `# Amp &amp; entities` + `*emphasis*` renders with the entity decoded and the heading/emphasis styling intact.
3. **Console clean** — in particular no `document is not defined`, which is the exact failure the decode-named-character-reference pin prevents.
4. **Small-canvas sync path** (1 node, below the offload threshold) renders the same markdown identically.

## Visual repro

The 13-node canvas immediately after a drag-commit laid out by the worker (entity decoded, heading + emphasis intact):

kamiazya:
## Manual verification result — PASS

Ran on macOS, Chromium via Playwright, against a Vite dev server in a clean worktree.

**Note on what was actually tested.** I started against the branch head I fetched at 07:53 (`71441637`), but this PR merged at 08:31 while the pass was running, so the result below is from **current `main` (`30f163c0`, containing this PR as `bef6ed2f`)**. That distinction turned out to matter — see the last section.

### 1–2. Markdown parses in the worker, renders correctly

Browser-local canvas grown to 14 nodes (past the 12-element offload line), then a node added and committed to force an offloaded re-layout. Every markdown case rendered correctly after the commit:

| body | rendered |
|---|---|
| `# Ampersands &amp; entities` | heading, `&` decoded |
| `*emphasis* and **strong**` | emphasis / strong |
| `Caf&eacute; &copy; &amp; more` | `Café © & more` |
| `### Worker path &lt;check&gt; *it*` | `<check>`, emphasis |
| `MAIN-PROBE &amp; *em*` | `MAIN-PROBE &`, emphasis |

**The worker path was proven engaged, not assumed.** The synchronous fallback is silent by design, so correct output alone proves nothing. I wrapped `window.Worker` in an init script before load:

- constructed: `http://localhost:5174/src/lib/layout-worker.ts?worker_file&type=module`
- every reply: `{ type: 'laid-out' }` — no `failed` message, so nothing fell back.

### 3. Console clean

Zero app errors across the whole session. No `document is not defined`, no worker errors. (The console shows two errors that are **mine**: a `blob:` module-worker sanity probe I ran, refused by `worker-src 'self'`. Unrelated — and a nice incidental confirmation that the CSP is doing its job.)

### 4. Small-canvas sanity — identical

Fresh 3-node canvas with the same three bodies: `window.Worker` recorded **zero** constructions (synchronous path, correct below the threshold), and the rendering is byte-for-byte the same reading: `MAIN-PROBE &` / `em`, `Ampersands` / `&` / `entities`, `Café © & more`.

### Worth recording: the per-canvas `canLayoutInWorker` is load-bearing

On the branch snapshot I first tested, `canLayoutInWorker(options)` judged per host, and **the worker never started at all** — instrumented and confirmed: zero `new Worker` calls on a 14-element canvas across several committed edits, while module workers were otherwise constructible in that browser. Cause: both real pages spread `{...fileSeams}` into the editor unconditionally (`BrowserLocalCanvasPage.tsx:774`, `DaemonCanvasPage.tsx:704`) and `useCanvasFileSeams` returns `useCallback`s that are never `undefined`, so the seam check was always false.

The merged version's per-canvas form — and the comment on it, which names exactly this — is what makes the offload reachable. The instrumentation above is the runtime evidence for both halves: never constructed before, constructed and answering `laid-out` after. If that judgement ever regresses to a per-host check, the whole feature goes silently dead while every test still passes, so it may be worth a guard that asserts a worker is used for a seam-carrying canvas with no file nodes.

PR #34, opened 2026-04-26
refactor: finish session -> workspace rename and drop legacy aliases

kamiazya:
## Summary
- Finish the public concept migration from session to workspace by dropping every back-compat surface I deliberately left behind in #33. Per the user's direction, back-compat is not needed here, so the legacy paths and aliases come out entirely instead of being kept in dual-write mode.

### Server
- Replace `registerWorkspaceAlias(...)` call sites with direct `app.METHOD('/api/workspaces/:workspaceId/...')` and delete `routes/workspace-alias.ts`.
- Remove the duplicate `/api/sessions` endpoint and the `canvas_list` legacy `sessions: [...]` output (with the matching zod schema in `mcp/index.ts`).
- Rename `listSessions` → `listWorkspaces`, `validateSessionId` → `validateWorkspaceId`, `loadSessionNames` → `loadWorkspaceNames`, `parseSessionNames` / `saveSessionNames` → `parseWorkspaceNames` / `saveWorkspaceNames`.
- Drop `resolveSessionId` / `saveLatestSessionId` legacy aliases from `mcp/session-resolver.ts` (the `.latest-session` filename and read-fallback are kept so existing on-disk state still resolves; only the unused TypeScript aliases are removed).

### Frontend
- Drop the legacy `sessionId?: string` field from the workspace list response type in `IndexPage.tsx`.
- Rename internal helpers (`filterSessions` → `filterWorkspaces`, `setSessions` → `setWorkspaces`).

### Tests / skills
- Update test fixtures and mock URLs from `/api/sessions/*` to `/api/workspaces/*`.
- Drop assertions against the removed `sessions` field in `canvas_list` tests.
- Apply the same word-boundary rename to `skills/whiteboard/SKILL.md` and `skills/whiteboard-audit/SKILL.md` so the docs match the new tool surface.

## Test plan
- [x] `pnpm test` (1202 passed: mcp-node + mcp-jsdom + mcp-browser)
- [x] `pnpm --filter @kamiazya/whiteboard-mcp typecheck`
- [x] `pnpm build`
- [x] `pnpm --filter @kamiazya/whiteboard-mcp run smoke:e2e` (`[e2e] ALL OK`)

## Notes
- This is a deliberate breaking change for any caller still hitting `/api/sessions/*` or relying on `canvas_list().sessions`. The user signed off on dropping back-compat ahead of this PR, so we ship it green rather than in a deprecation window.
- 71 files changed, +876 / −956 (net ~80 lines removed once the alias layer is gone).

PR #814, opened 2026-08-15
fix(web): stop the closing canvas switcher stealing the new editor's focus

kamiazya:
Closes the CodeMirror autofocus flake (task #34) by **capturing** the thief, then fixing both directions of the race it revealed.

## The evidence

A focusin/focusout recorder attached during the flow, read on failure:

```
in:  DIV.cm-content            ← the editor mounted AND focused (autofocus worked)
out: DIV.cm-content → DIV.bg-popover data-[state=ope…   ← the closing menu steals it
out: DIV.bg-popover… → null    ← the menu unmounts; focus falls to <body>, permanently
```

Radix menus are **modal by default**, and the modal focus trap lives for the content's whole lifetime — exit animation included. Under load the animation outlasts the editor's mount, the trap treats the editor's autofocus as an escape and yanks focus back into the dying menu, and the editor's one-shot mount focus never fires again.

Two hypotheses died against the same recorder first: ambient window focus (a probe held focus alongside all 543 browser tests) and dispatch timing (a one-macrotask deferral reproduced the *identical* steal log — which is what ruled it out).

## The fix, in two halves

1. **`modal={false}` on the switcher.** Every selection in this menu navigates away, so the trap protected nothing worth that race. Zero steals in 10 instrumented runs after the change (the pre-fix loop reproduced within 2–3).
2. **Autofocus etiquette in SourcePane.** With the trap gone, the editor's one-shot `focus()` becomes the potential thief in the other direction: it mounts in a deferred effect, so a user who clicks the title input right after "New markdown note…" can be typing before the effect runs. The guard claims focus only from *transient* holders — `<body>`, null, or anything inside a dismissing menu (claiming from THAT is autofocus's whole job) — and respects a real surface like the title input.

## Scope honesty — what this does NOT fix

CI's first run of this PR failed two **title** tests (`'' !== 'リリース計画'`). Investigated to a capture, and it is a *different, pre-existing* bug: an element-identity probe shows `titleIsConnected=false` — the title input **remounts mid-typing** (async reload flips the render gate / `key={canvasId}`), and the controlled value re-reads a store the debounced keystrokes haven't reached. It reproduces identically on unmodified main with the same two-file run, so it is not introduced here. Tracked as task #37 with the full evidence chain; expect `test-browser` to remain vulnerable to it until that lands.

Also keeps a discriminating failure message on the file-gc-sweeper reschedule test (task #35): `listWs=0` = its timer never fired; `listWs>0, purge=0` = the pass stalled in the real-fs containment check, the one unmocked dependency. Too rare to reproduce on demand; the next CI occurrence reports its own cause.

Verification: web jsdom 2,154 + 75 green; browser suite green in isolation for the touched flows.

https://claude.ai/code/session_01CSjf55ombrNgn1gmx6Seb3

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved editor focus handling so menus and source editing do not interrupt active fields.
  * Prevented reloads from racing with pending saves, preserving document metadata and content.
  * Improved canvas menu transitions for smoother editor interactions.

* **Tests**
  * Added regression coverage for save ordering, reload behavior, and persistent title editing.
  * Improved reliability and diagnostics for browser and file cleanup tests.

### DIFF
--- a/Dockerfile.server
+++ b/Dockerfile.server
@@ -46,6 +46,12 @@ WORKDIR /workspace
 FROM base AS fetched
 
 COPY package.json pnpm-lock.yaml pnpm-workspace.yaml ./
+# patches/ travels WITH the lockfile, not with the sources copied later:
+# pnpm-workspace.yaml's patchedDependencies makes the patch part of what
+# `pnpm fetch` resolves, so a missing file here fails the fetch outright
+# (ERR_PNPM_PATCH_NOT_FOUND) rather than silently shipping an unpatched
+# dependency.
+COPY patches ./patches
 
 RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
     pnpm fetch --store-dir /pnpm/store
--- a/apps/web/index.html
+++ b/apps/web/index.html
@@ -2,7 +2,17 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <!--
+      App-like viewport: user scaling off (mobile Chrome honors it, and
+      maximum-scale=1 is also what stops iOS Safari auto-zooming a focused
+      input; Safari's pinch itself is cancelled in mobile-app-shell.ts),
+      viewport-fit=cover so the installed PWA extends under the notch/home
+      indicator with index.css's safe-area padding taking over.
+    -->
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no, viewport-fit=cover"
+    />
     <title>Whiteboard</title>
     <meta
       name="description"
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -53,7 +53,7 @@
     "react": "catalog:",
     "react-colorful": "^5.8.0",
     "react-dom": "catalog:",
-    "react-router-dom": "^7.18.0",
+    "react-router-dom": "^7.18.2",
     "tailwind-merge": "^3.5.0",
     "tw-animate-css": "^1.4.0",
     "zod": "catalog:"
--- a/apps/web/src/components/WorkspaceTopBar.tsx
+++ b/apps/web/src/components/WorkspaceTopBar.tsx
@@ -133,6 +133,16 @@ export default function WorkspaceTopBar({
 
   const [canvasSearch, setCanvasSearch] = useState('')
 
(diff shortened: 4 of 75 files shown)


kamiazya:
## Review triage

All four CodeRabbit findings addressed. One of them was a real defect, not a nit.

### 🟠 Major — "Track debounce saves after the timer fires" — **valid, fixed**

Confirmed against the code. This PR established the invariant that a save in flight is
awaited before the next `load` reads the store — but only for the save the **cleanup**
starts. The save the **debounce timer** starts was never registered: the timer clears
`timerRef.current` before saving, so a later unmount finds no timer to flush and the
reload has nothing to wait on. Same ending as the bug this PR set out to fix — the load
reads pre-edit state and the edit is gone — reached through the one path the fix did not
cover.

Reproduced red first. The store log came back `[load, save-start, load, save-end]`: the
reload read the store before the save landed.

Fixed by putting both writers through one per-document queue (`queueSave`). That also
orders two in-flight saves, which the previous shape did not: each save is a full
snapshot export, so an earlier one landing last would put the document back.

Checked the blast radius — the spatial controller has no debounce (discrete edits flush
immediately), so this shape was specific to the markdown hook.

### 🟡 "Remove the timing assumption from this ordering test" — **valid, fixed**

Correct: the 50 ms sleep proved nothing about ordering. Both ordering tests now assert
the order from a store log (`load` must not fall between a save's start and its landing)
and wait one macrotask tick rather than a guessed delay. A slower machine can no longer
turn a broken build green.

Mutation-checked both: dropping the cleanup registration turns the existing test red;
the new test was red before the fix.

### 🟡 "Remove temporary transition diagnostics before merge" — **already addressed**

Removed in 975d199 (`test(web): drop the temporary remount diagnostics`); no
`__pageTrace` remains in `apps/web/src`.

### 🟡 "Make the comment durable and accurate" — **valid, fixed**

Both parts were right. The live focus trace is chronology, not rationale, and "every
selection here navigates away" is false — re-picking the current workspace is guarded to
a no-op. The comment now states only the durable reason (Radix's modal menus trap focus
through the exit animation; the editor's autofocus is one-shot on mount, so nothing
recovers once it is stolen).

The same false claim appeared a second time in the `onCloseAutoFocus` comment, which was
not flagged; corrected there too.

## Verification

- `pnpm --filter @kamiazya/whiteboard-web test` — 2188 + 77 passed
- `pnpm test --project web-browser` — 564 passed
- `pnpm -r typecheck` clean; pre-push gates green

PR #815, opened 2026-08-15
perf(web): stop re-running the edge-anchor pass at drag start

kamiazya:
## What

The remaining half of #39's drag-start work, instrument-first per the measured-change discipline:

1. **Commit 1 — the instrument.** `drag-start-responsiveness.browser.test.tsx` drives two real move gestures on a 45-node/90-edge canvas and logs the longest inter-frame gap; only "the gesture really started" is asserted (wall time pinned as a threshold is a flake generator). Baseline: steady-state maxGap ~90-96ms — past the ~50ms perceptibility line, at the worst possible moment (mid-motion).
2. **Commit 2 — the fix.** A throwaway split showed the block was **~21ms backdrop render + ~150ms `assignEdgeAnchors`** — an anchor pass the committed layout had *already run in the worker* to produce the scene on screen, recomputed synchronously at gesture start. Now the anchors travel out with the scene:
   - `canvas-render`: `layoutSpatialCanvasWithAnchors` returns the layout **and** the anchor map it routed with (`layoutSpatialEdges` precedent — this IS the layout, not a second computation). `layoutSpatialCanvas` delegates; its four external callers see no diff.
   - The map rides the worker protocol as a structuredClone'd `Map`; the parity browser test pins entry-for-entry equality across the thread boundary.
   - `SpatialEditor.dragStatic` consumes the committed anchors; the synchronous `assignEdgeAnchors` call (and its import) is deleted outright.

## Measured (interleaved before/after, same session)

| | first gesture maxGap | steady-state maxGap | gaps >50ms |
|---|---|---|---|
| before (stashed) | 150.9ms | **95.8ms** | 1 |
| after | 85.0ms | **36.1ms** | 0 |

The stash genuinely reverted (numbers regressed), avoiding the no-op-stash trap.

## Design record

An adversarial consult (3 rounds, accepted) rejected two alternatives: a `data-node-id`/CSS-hiding change to the shared SVG backend (no contiguous per-node subtree exists — a canvas node emits sibling scene nodes with only the chrome shape carrying the id — so it would force per-node `<g>` restructuring and byte-golden/export churn), and speculative selection-time pre-rendering in the worker (strictly dominated: the worker already produced the committed scene). It made the anchors move **conditional on splitting the 90ms**, which the split then won decisively. Semantics improve as a side effect: the overlay now pins bystander edges to the anchors of the scene actually on screen, provenance-consistent with `frozenSidesOf`, instead of a fresh pass over a possibly newer canvas. The backdrop-reuse rung (~21ms remaining) stays documented in the ponytail marker for canvases where it grows past the line.

## Tests

canvas-render 615, apps/web jsdom 2167 + node 77, browser 557 (parity anchors case included), consumer projects 365, typecheck, knip — all green on the merged base.

### DIFF
--- a/apps/web/src/components/spatial-editor/SpatialEditor.tsx
+++ b/apps/web/src/components/spatial-editor/SpatialEditor.tsx
@@ -112,6 +112,7 @@ import { applyCommand, buildFragmentInsertCommand } from './commands.js'
 import { CREATION_LABELS } from './creation-labels.js'
 import { DragPreviewLayer } from './DragPreviewLayer.js'
 import { computeDragPreview, isInFlightGesture } from './drag-preview.js'
+import { useGestureCaptured } from './use-gesture-captured.js'
 import { createEditorAppearance, editorTextFill } from './editor-appearance.js'
 import { isFollowableUrl } from './followable-url.js'
 import type { Box, ResizeHandleKind } from './geometry.js'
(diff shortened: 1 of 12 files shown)


kamiazya:
**Review-lane outcome** (internal multi-dimension review + adversarial verify): one HIGH confirmed and fixed in `39724a-`…`HEAD`:

> `committedAnchors` tracked the live `useWorkerScene` value, so a layout reply landing mid-drag swapped the points bystander edges are pinned to — the exact re-fraction the pin exists to prevent.

Writing the regression test exposed the finding's second half: pins apply only to edges present in `frozenSidesOf(scene)`, so a mid-gesture **scene** swap silently un-pins every bystander even with the anchors held. The fix captures the `{scene, anchors}` pair once per gesture (`useGestureCaptured`, unit-tested) and routes every side/pin provenance read through the capture. The browser regression test drives a real drag on an offloadable canvas, lands a stubbed worker reply mid-gesture with genuinely different anchors *and* scene, and pins the live bystander path bytes — mutation-checked red both by bypassing the capture and against the original defect.

The accepted residual (also from the design consult): the captured layout can lag the canvas by one worker round trip for a drag started immediately after an edit — bounded, self-heals at drop, and provenance-consistent with what is actually on screen.

PR #821, opened 2026-08-16
fix(web): stop a double-tap's new node vanishing when the touch capture is released

kamiazya:
## Summary

In select mode, a double-TAP on empty canvas created a note, opened it for typing, and then
the note vanished. Double-CLICK was fine, which is why the existing coverage stayed green.

`lostpointercapture` was wired to the same handler as `pointercancel`, so a capture the browser
refuses or revokes mid-gesture could not strand an in-flight gesture. But that event also fires
on the **ordinary release** of a captured pointer — and browsers capture touch pointers
**implicitly** on every `pointerdown`. On a touch device it therefore arrives after every tap,
including the second tap of the double-tap that had *just* created the node. The node is
`createdForEdit`, and cancelling such an edit takes the node with it rather than leaving an empty
box behind, so it was created and immediately deleted.

Mouse never showed it: capture is taken at the first **move**, so a stationary double-click holds
none to lose.

**The fix** uses state that already exists. `handlePointerUp` clears `activePointerIdRef`, so a
null ref at `lostpointercapture` time means the interaction already ended normally and there is
nothing to recover; a non-null ref means the capture went away mid-gesture, which is the case the
handler was written for. `pointercancel` itself is untouched and still discards the node.

## Repro, measured in the running app

CDP touch emulation against the deployed preview (pre-fix), double-tapping empty canvas:

```
pointerdown:touch → gotpointercapture:touch → pointerup:touch → lostpointercapture:touch
textarea count: 0 → 1 → 0
```

`gotpointercapture` right after `pointerdown` is the implicit touch capture; `lostpointercapture`
right after `pointerup` is it being handed back on a perfectly normal release.

## Verification

- [x] New `lost-pointer-capture.browser.test.tsx`: red before the fix
      (`expected [ 'create-node', 'delete-node' ] to not include 'delete-node'`), green after
- [x] Second case pins the property NOT being traded away: a real `pointercancel` still discards
      the node
- [x] `pnpm test --project web-browser` 566 passed; apps/web jsdom 2188 + 77 passed;
      `pnpm -r typecheck` clean

- [x] Same probe re-run against the preview built from THIS branch — same environment, same
      steps, so the difference is attributable to the change:

| | before (main's preview) | after (this branch's preview) |
|---|---|---|
| events | `pointerdown → gotpointercapture → pointerup → lostpointercapture` | **identical** |
| textarea count | `0,0,0,0,1,0` — **vanishes** | `0,0,0,1,1` — **stays** |
| end state | `0` | `1`, `focused: TEXTAREA` |

`lostpointercapture` still fires after the fix, so the trigger genuinely occurred and the node
survived it. (A local dev run was NOT accepted as verification: the capture events never fired
there under the same CDP emulation, so the node surviving proved only that the trigger was
absent.)

## Noticed while writing the guard, not fixed here

A `lostpointercapture` arriving mid-drag does not actually cancel the drag today — the gesture
survives and the later `pointerup` still commits a `move-node`. That is the recovery this handler
claims to provide, so it is worth its own look; it is out of scope for this fix and nothing here
makes it worse.

### DIFF
--- a/apps/web/src/components/spatial-editor/SpatialEditor.tsx
+++ b/apps/web/src/components/spatial-editor/SpatialEditor.tsx
@@ -1913,6 +1913,24 @@ export const SpatialEditor = forwardRef<SpatialEditorHandle, SpatialEditorProps>
     }
 
     /**
+     * `lostpointercapture` is NOT a synonym for "the gesture broke". It also
+     * fires on the ordinary release of a captured pointer — and the browser
+     * captures touch pointers IMPLICITLY, so on a touch device it arrives
+     * after every single tap.
+     *
+     * `activePointerIdRef` is what tells the two apart: `handlePointerUp`
+     * clears it, so a null ref means the interaction already ended normally
+     * and there is nothing to recover. Cancelling there deleted the node the
+     * empty-canvas double-TAP had just created for editing — it appeared and
+     * vanished, while the same double-CLICK was fine because a stationary
+     * mouse press takes no capture to lose.
+     */
+    const handleLostPointerCapture = (e: React.PointerEvent<HTMLDivElement>) => {
+      if (activePointerIdRef.current === null) return
+      handlePointerCancel(e)
+    }
+
+    /**
      * The selection as reorder targets, primary + extras, deduped. The
      * command treats ids as a set and takes relative order from the canvas.
      */
@@ -2873,7 +2891,7 @@ export const SpatialEditor = forwardRef<SpatialEditorHandle, SpatialEditorProps>
         onPointerMove={handlePointerMove}
         onPointerUp={handlePointerUp}
         onPointerCancel={handlePointerCancel}
-        onLostPointerCapture={handlePointerCancel}
+        onLostPointerCapture={handleLostPointerCapture}
         onKeyDown={handleKeyDown}
       >
         {/* Screen space, outside the pan/zoom transform — an overview that
--- a/apps/web/src/components/spatial-editor/lost-pointer-capture.browser.test.tsx
+++ b/apps/web/src/components/spatial-editor/lost-pointer-capture.browser.test.tsx
@@ -1,0 +1,91 @@
+/**
+ * `lostpointercapture` is wired to the same handler as `pointercancel` so a
+ * capture the browser REFUSES or revokes cannot strand an in-flight gesture
+ * (this component registers no window-level fallback listeners).
+ *
+ * But that event also fires on the perfectly normal release of a captured
+ * pointer, and touch pointers are captured IMPLICITLY by the browser on
+ * every pointerdown. So on a touch device it arrives after every tap —
+ * including the second tap of a double-tap, which is where the empty-canvas
+ * double-press creates a node and opens it for typing. Cancelling there
+ * deletes that node (it is `createdForEdit`), so the note appears and
+ * vanishes. Mouse input never showed it: capture is taken at the first
+ * MOVE, so a stationary double-click holds none to lose.
+ */
+import type { SpatialCanvas } from '@kamiazya/whiteboard-canvas-model'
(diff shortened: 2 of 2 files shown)


kamiazya:
## Review triage

### 🟠 Major — "Drive this regression with touch pointer events" — **valid, fixed** (f8101d7)

Correct: the guard was asserted by dispatching `lostpointercapture` after a mouse `dblClick`,
which exercised the handler but not the path that produces the event in the field. Each tap is
now a real touch sequence — `pointerdown`, `pointerup`, and the capture handback the browser
performs right after it — so the test fails for the same reason a phone does.

Re-mutation-checked at the new shape: pointing `onLostPointerCapture` back at
`handlePointerCancel` turns both cases red (`expected null not to be null` — the node is gone).
A touch-driven test that passed against the broken code would have been worse than the mouse one.

### 🟠 Major — "Track each captured pointer ID" — **real, but pre-existing; not this PR**

Checked against the code rather than accepted or dismissed. `handlePointerUp` returns early on
the gather paths (L1739, L1743) and the pinch path (L1751) *without* clearing
`activePointerIdRef`; only the plain single-pointer path clears it (L1754). So during a pinch the
ref stays non-null, the guard passes, and the first finger's handback runs a full cancel while the
second finger is still down. That is worth fixing.

It is not a regression from this PR, and the reason is structural: the change is
`always cancel` → `cancel iff activePointerIdRef !== null`. The set of cancels after is a strict
**subset** of the set before, so no cancel can occur now that did not occur previously. This PR
strictly reduces the exposure the finding describes; it does not create it.

Fixing it properly means tracking the captured ids as a set and removing only `e.pointerId` per
lifecycle path, plus a browser regression for "one finger released normally, the other's capture
lost" — a change with its own failure modes that deserves its own diff and its own red test.
Folding it in here would put two unrelated judgements behind one review. Recorded as follow-up.

### 🟡 Minor — "Update the stale event-wiring statement" — **valid, fixed** (f8101d7)

The file comment still described `lostpointercapture` as wired to the same handler as
`pointercancel`, which stopped being true in this PR. It now states the guarded contract.

## Verification after the changes

- `pnpm test --project web-browser` 566 passed
- `pnpm lint`, `pnpm -r typecheck` clean

kamiazya:
## Correction to this PR — two things I wrote here are wrong

Both are corrected in #823. Recording them here so nobody reads this PR and carries the wrong
premise forward.

### 1. "Track each captured pointer ID" — I said pre-existing, not a regression. It IS a regression, from this PR.

My argument was that the change could only ever *remove* cancels, never add them, so the set of
cancels after is a strict subset of the set before and nothing new can break. That shows no *new*
cancel occurs. It says nothing about a *needed* cancel going missing — which is exactly what
happens.

`activePointerIdRef` is a single slot, and `handlePointerUp` leaves it alone on the pinch path.
So the first finger's ordinary handback runs a full cancel, that cancel nulls the ref for the
finger still down, and the second finger's genuine capture loss is then skipped. The pinch
bookkeeping is left holding a pointer id nothing will release, and the next touch to inherit it
does nothing.

Measured on the same sequence, which is what I should have done before claiming otherwise:

| | later double-tap |
|---|---|
| without this PR's guard | `create-node` |
| with it | `[]` — nothing happens |

CodeRabbit was right and my rebuttal was not.

### 2. "A mid-drag lostpointercapture does not cancel the drag" — also wrong; disregard it.

That was an artifact of my own probe dispatching pointerdown/pointermove/lostpointercapture in a
single task, so the handler closed over a stale `gestureState`. With a render between the events
the cancel works correctly (`afterCancel=[] final=[]`). There is nothing to fix, and the
follow-up I suggested here should not be filed.

The fix in this PR — not cancelling on an ordinary capture handback — remains correct and
verified; only the predicate it used to decide "ordinary" was too coarse. #823 replaces it with a
per-pointer one.

PR #822, opened 2026-08-16
feat: render a referenced markdown document inside a canvas file node

kamiazya:
## Summary

- A spatial `file` node pointing at a markdown document in the same workspace now renders that document's **body** — heading, prose, lists — instead of a facet card that said what the document *is* and nothing about what it *says*. Wired on all three surfaces: the `apps/web` editor, the MCP `wb_scene_render` tool (opt-in), and the MCP Apps inline widget.
- Fixes a real asymmetry found on the way: `layoutSpatialCanvas` never forwarded `renderMath` / `renderDiagram` / `resolveEmbed` to `layoutMdastBlocks`, so the same document rendered its math and mermaid fences in the markdown editor and rendered them as placeholders the moment it was laid out inside a canvas text node. One engine, two answers depending on which surface called it.
- Revives `canvas_view`, the MCP Apps UI tool. Every other part of that path already shipped — the `ui://whiteboard/canvas-view` resource, the extensions capability, `_meta.ui.resourceUri` support, the widget bundle, and an empty `UI_LINKED_TOOLS` list waiting for it — and the OpenCanvas migration had removed the one tool that linked them.

## Test plan

- [x] New or updated tests added — `canvas-render-node`, `server-core-node`, `mcp-node`, `apps/web` jsdom, `web-browser`, `canvas-viewer-browser`
- [ ] `pnpm test` passes locally — **9 of 7801 fail, all pre-existing and environmental** (see below)
- [x] E2E coverage added or extended where applicable — `pnpm smoke:e2e` extended with `canvas_view` and `wb_scene_render(embedReferences:true)`, run green

Root `pnpm test`: **7792 pass, 9 fail**. All nine are `object.stream is not a function` in thumbnail/objectURL tests, and all nine **reproduce unchanged on `origin/main`** in this container — verified on a clean worktree, not assumed. CI's `test-jsdom` job is green on this branch, which confirms the diagnosis. Also run: `pnpm smoke:e2e` (`ALL OK`, through a real MCP stdio subprocess), `pnpm -r typecheck`, `biome check`, `knip` — all clean.

**Manual verification** (AGENTS.md step 3) is what makes this PR worth reading twice. It found a defect every test missed: at the box the app actually creates for a Document node (200×60, a 44px padded content box) **not one markdown block fits**, so the body rendered as nothing and the node looked exactly as it had before the feature existed. Each test had built a node large enough for its own assertion, so none exercised the real size. Fixed by having the picker carry the referenced document's kind, so the node is sized where the target is known.

## Visual evidence

Captured by driving the running app in a real browser — create a markdown note, add a Document node pointing at it — and kept at `tmp/screenshots/before-default-size-shows-no-body.png` and `after-default-size-renders-body.png`.

They cannot be attached here: uploading needs the `gh image` extension, and the `gh` CLI is not available in this environment. Described precisely instead, and the flow is pinned by a `web-browser` test that renders at exactly the default document-node size:

- **Before** — the node shows only the label `untitled` inside its frame. The markdown seam returned `undefined`, so the plain-label path rendered.
- **After** — the reference label sits *above* the frame (the container convention the canvas-embed seam already uses) and the body fills the box: `Weekly notes` as an H1, the paragraph, then the bullet list.

The Cloudflare Pages preview is the better check for a reviewer with a browser, and one I could not run: Chromium in this environment cannot reach any external host (`ERR_CONNECTION_RESET`, `example.com` included), so only `localhost` was drivable. Steps on the preview: `/local` → Create a canvas → `Local ⌄` → New markdown note… → type a body → back to the canvas → `+` → Document → pick the note.

## Notable decisions

**Seam precedence is `image > canvas-embed > markdown > facets > label`.** A document's own prose says more about it than the facets describing it, and less than an image or spatial canvas the caller already resolved.

**MCP resolution is opt-in on `wb_scene_render` (`embedReferences`, default `false`) and unconditional on `canvas_view`.** The flag exists only because the render tool shares a scene builder with `wb_scene_digest`, and a digest that moved whenever a *different* document was edited would stop being usable as a change signal. `canvas_view` has no such sibling and its one consumer is a viewer for a human.

**Only a markdown-kind document yields a body, decided by `readDocumentKind`.** This is load-bearing rather than defensive: a markdown document's stored content on the server *is* a valid spatial canvas holding one text node, so "does it parse as a canvas" cannot tell prose from a diagram.

**Two storage shapes, now one reader.** `wb_document_set` stores a body as an `okf-body` text node; `apps/web`'s browser-local editor stores a Loro `body` text container. Neither side could read the other's documents. `readMarkdownBody` in `canvas-workspace` answers for both — and that package is the only place both `server-core` and `apps/web` can reach it from. No behavior changes today (browser-local documents live only in IndexedDB and never reach the server path). **Unifying the writers is still open.**

## Removals

The widget's add-a-sticky-note control called an `annotate` tool the OpenCanvas migration removed, so every submission failed at the host while the control still looked live. Unwired, with a `TODO(annotate)` marker naming why restoring it needs a decision rather than re-wiring: OpenCanvas has no `annotate` equivalent, a sticky note is a `wb_node_add` of a text node, and this widget is otherwise strictly read-only. Its modules are kept, unmounted and still tested.

## Review round

The repo's own review gate ran over this diff and confirmed 5 findings (1 HIGH, 1 MEDIUM, 3 LOW, 0 CRITICAL, QA green). All are fixed in `ae117e4`. The HIGH is worth calling out because it was a real defect I introduced and my own code comment asserted the opposite:

The widget validated a reference body as `{type:'root', children: unknown[]}` and cast the rest away, on the stated grounds that "canvas-render degrades on anything inside it that it cannot lay out". It does not. `layoutBlock`'s switch has no default case, and `composeFileMarkdown` guarded only the resolver *call*, not the layout — so a null, a primitive or an unrecognised `type` threw out of `layoutSpatialCanvas` and took the whole canvas with it, unrelated nodes included. Five distinct throws reproduced by the verifier.

Fixed on both sides: the widget parses against the canonical `mdastRootSchema` (the same one `canvas_view` declares server-side), and canvas-render now honours its own documented never-throw rule at the seam that made it reachable. Tightening the widget schema inline first caused a *second* regression — one bad reference failed the whole envelope and discarded a good scene — so validation is applied per reference, and a test caught that before it left the working tree.

## Checklist

- [x] Docs updated — `docs/how-to/view-canvas-in-chat.md` moved off its "Currently unavailable" banner to the shipped state, including that the sticky-note control is gone
- [x] No hand-written TypeScript interfaces duplicate a Zod schema — `canvas_view`'s payload crosses two process boundaries, so `references` is schematized through `mdastRootSchema` rather than passed as `unknown`

## Reviewer notes

Every new guard here was mutation-checked — reverted the thing it protects, confirmed red, restored. Three are worth a look because they were guards that *could not fail* before:

- `canLayoutInWorker`'s seam checks are now asserted **one seam at a time**. The combined fixture set five seams, so four other clauses already returned `false` and deleting any one left every assertion green — which is exactly how a clause got added untested.
- `UI_LINKED_TOOLS` was a list nobody checked. It now drives both the ADR-0009 naming exemption and an assertion that the real registration carries `_meta.ui.resourceUri`, so the two cannot disagree about which tools are UI tools.
- The `callServerTool` allowlist narrowed to `canvas_view` alone when `annotate` was unwired, so a call site naming a removed tool fails rather than being waved through by a stale entry.

Two things are **not** verified: the widget's behavior inside a real MCP Apps host (none available in this environment — the tool, the payload extraction and the viewer rendering are each covered, the host↔widget hop is not), and the Cloudflare preview, for the network reason above.

## Follow-up, not in this PR

`CanvasKind` should be `DocumentKind` — `vocabulary.md` already lists `canvasKindSchema` as a known violation, and `['spatial','markdown']` describes a *document*'s format. It crosses package boundaries, so per that same rule it is its own increment rather than a drive-by here.

https://claude.ai/code/session_01SGjid2m1cEL6t1VtNpHU2D


kamiazya:
@coderabbitai review

The earlier run hit the plan's rolling review limit before it could start. Retrying now that the window should have reset.

---

PR #826, opened 2026-08-16
feat!: converge on ADR-0009's Document vocabulary

kamiazya:
## Summary

Converges the codebase on [ADR-0009](docs/contributing/adr/0009-mcp-tool-naming.md)'s Document vocabulary, in increments that each pick one violation and remove the place the wrong state could live rather than correcting call sites one at a time.

- **Unification first.** A markdown document's body is now stored in exactly one place, the CRDT text container (`writeMarkdownBody`) — browser-local editor, daemon editor and `wb_document_set` alike. That fixed the same interop defect twice from opposite directions: first a daemon-written document opening with an empty editor beside a preview showing its prose, with the next keystroke saving over the original; then, after `main` grew a daemon markdown editor reading the older representation, an agent-authored document opening blank. The six `resolveFile*`/`resolveEmbed` layout seams collapsed into one `resolveReference` returning a plain record, and apps/web's three hand-duplicated prefetch-behind-a-synchronous-seam hooks became one `usePrefetchedCache` — which surfaced an infinite re-fetch loop on a dangling reference (568 loads for one broken link).
- **Names, through the stored and published shapes.** `CanvasKind` → `DocumentKind`, `CanvasCoreMeta` → `StoredCoreFacets`, `CanvasDocStore` → `DocumentStore`, `canvasId` → `documentId` (including the `x-whiteboard` extension field), `slug` → `path` (DB migrations `0009` + `0010`), `canvas-store.ts` → `document-store.ts`, and the four mis-prefixed package names.
- **Two behaviour changes the ADR asked for.** A document's name lives only in the workspace — `storedCoreFacetsSchema` omits `title`, so a second copy is unrepresentable and OKF frontmatter `title` is a projection in both directions. And a spatial document has no facets: `readCoreFacets` answers `undefined` for one, and apps/web's spatial canvas row passes no facets rather than hiding the disclosure while still writing through it.

Three things worth a reviewer's attention:

- **`packages/canvas-workspace` became `packages/loro-adapter`, after two wrong answers.** Dropping the prefix gives `workspace`, a name this repo's vocabulary already assigns to the tree that holds documents — which that package explicitly knows nothing about; applied, its comments read `workspace's withSpatialBatch`, indistinguishable from the dozens of correct uses of the domain word around them. `crdt` fixed the collision but named a technique and left the package's place in the architecture unsaid. The dependency graph settles it: this package depends on `loro-crdt` and **not** on `ports`, implements no port, and its whole production surface is one file (`loro-bridge.ts`). The store and sync ports are implemented in the composition roots, so "adapter" here can only mean an adapter of the vendor library. `ports` is dropped from its allowed internal deps in `architecture-map.ts` to match — it was permitted there and never used, which is exactly the reading the old name invited.
- **`slug` is now enforced by a test.** `tools/arch-lint/src/vocabulary-check.test.ts` fails on the word appearing anywhere under the scanned source trees. It is the one part of the vocabulary rule that could stop being prose, because `slug` has no legitimate meaning left; `canvas` will never join it. It earned itself twice — see the review responses below.
- **#837 landed on `main` mid-flight and is reconciled here, not merged verbatim.** It added the daemon markdown editor on top of wiring this branch had deliberately removed, so what each side keeps is a decision rather than a textual merge. **Kept:** `DocumentEditorSurface` and the daemon markdown editor — its kind switch is the point, since a page can no longer choose an editor by itself, which is how a markdown document once opened in the spatial editor on one backend. **Dropped:** the daemon page's `titleSlot` Properties bar. Its own comment states the intent it cannot have — *"the title here is the document's own core facet, not a rename"* — and a document's NAME is the workspace's (decision 2), which is why `storedCoreFacetsSchema` no longer has `title` and why this branch removed the `coreFacets` path from `useCanvasSync` and `canvas-sync-session`. Reinstating that path to keep a bar built on the forbidden reading, then removing it again next increment, is a round trip that buys nothing. Its two tests go with it, and the note left in their place says where the title-box coverage lives instead (browser-local, which has a real name to show).

Titled `feat!:` on review: this carries two DB migrations, two behaviour changes, and a break in a published interchange format (`x-whiteboard.canvasId` → `documentId`, embedded in exported PNGs and JSON Canvas files). `refactor:` would produce no version bump *and no changelog entry*, so a consumer would get the renamed field with no release note.

## How the daemon body write works now

The daemon editor's persistence moved to the container without giving up what #837's routing bought it. The read comes from the sync session, which republishes the body the same way it republishes locks — they are alike in the way that matters: neither is part of the canvas VALUE, so neither gets a canvas publish, and both need their own notification. It fires on every doc change rather than only on an import, because a local UNDO rewrites the body without one.

The write is a new `set-body` command. Deliberately **not** an `EditorLeafCommand`: it targets no node, leaves the canvas value identical (`applyCommand` returns the same reference, so a keystroke in the markdown editor does not re-render the spatial scene) and can never appear in a batch. Riding the command path anyway is what buys the same debounce, undo step and local-update forwarding every other change gets, instead of a second write pipeline beside them. `writeCommandTarget` must claim it — the fallback there writes the whole `SpatialCanvas`, which would leave the container untouched and drop the edit silently.

`canvasWithMarkdownBody` is deleted: with the editor moved, nothing writes the node form any more, and leaving an exported writer for a superseded representation is how the two grew apart in the first place. `markdownBodyFromCanvas` stays as a **private** read fallback so documents an older writer left are still readable.

## Test plan

- [x] E2E coverage added or extended where applicable

**Correction to an earlier version of this section.** It claimed nine failures "pre-existing on `origin/main` … verified unchanged in count and identity throughout". Review could not reproduce those numbers, and re-measuring shows why the claim was shaped wrongly:

- The nine (`CanvasThumb`, `MergeDialog`, `VersionThumbnail` ×4, `VersionTimeline`, `daemon-file-adapter` — all objectURL/thumbnail) fail **on this sandbox only**, on `origin/main` and on this branch alike. They do not fail on the reviewer's machine. Environment-specific, not a property of the repo, and the original wording implied the latter.
- Worse, `pnpm test` here reports `Test Files 5 failed | 660 passed (779)` — 114 files never ran. The browser projects cannot launch in this sandbox (`browserType.launch: Executable doesn't exist`) and vitest reports that as **one** `Unhandled Error` while the remaining projects report green. That reads almost exactly like a full green run. Same trap as the `--project` one `.claude/rules/dev-flow.md` already documents, in a different form.

What the runs actually support, re-measured after each merge of `main`: the root `pnpm test` at **7308 passed** with only the nine environment failures above, and the browser projects run **separately** with an explicit chromium path — `pnpm test:browser`, **115 files / 592 passed**. `pnpm -r typecheck`, `pnpm lint`, `pnpm knip`, `pnpm install --frozen-lockfile` clean. CI is the authoritative gate for the rest, and it is green on this head.

Regression coverage added at the nearest layer for each increment, plus browser-mode tests for the flows only a real browser exercises: a legacy daemon-written document staying editable through CodeMirror, the title round-tripping through the workspace with no `title` facet left in the document, and a spatial canvas having no Properties disclosure. Every new guard was mutation-checked.

The daemon-body change was red first — the fixture writes the container and the page went blank on it — and mutation-checked by forcing `getMarkdownBody` to return the empty string, which turns it red again. Its browser round-trip test asserts on the container **directly** rather than through `readMarkdownBody`, whose node fallback would have accepted a body written the old way.

The packaged smokes were run directly rather than read: `packaged-daemon-backup-restore-smoke` and every other packaged smoke exit 0, and `pnpm smoke:e2e` reaches `ALL OK`. (`smoke:claude`/`smoke:codex` need CLIs this environment lacks; CI skips them for the same reason.)

Two DB migrations are included. `0009-document-vocabulary` renames the `canvases` table to `documents` plus four `canvasDoc*` tables and their `canvasId` columns; `0010-document-path` renames `slug` to `path`. Both have up/down tests, and `migrator.legacy-upgrade.test.ts` migrates a pre-0008 data dir through the real migrator to prove a nanoid row and its blob survive the whole chain.

## Visual evidence

Not applicable to most of this — a rename sweep plus contract changes. Two user-visible surface changes, both stated consequences of ADR-0009 rather than regressions:

- A **spatial canvas no longer shows the Properties (type/tags) disclosure** (decision 3). The disclosure was already hidden there via `showFacets={false}`; this PR removes the writing that continued behind it. Pinned by `BrowserLocalCanvasPage.markdown.browser.test.tsx`.
- The **daemon page shows no title/Properties bar** (decision 2, see the third bullet above). `main`'s version of it read the title out of the document's own facets; the browser-local page, which has a real workspace name to show, keeps its bar.

## Checklist

Docs moved with the code: `docs/explanation/domain-model.md` now states that the display name is the only place a canvas is named and that a spatial canvas has no facets, and says `path` throughout; `docs/explanation/architecture.md` no longer documents a `targetSlug` parameter that no longer exists. ADR-0009 gains an addendum recording that `format` is spelled `kind` and that decisions 3-4 are now implemented; ADR-0007 and ADR-0008 gain vocabulary notes (*read slug as path*), and the ADR index carries the same signal — no decision text is rewritten.

## What is deliberately left

- `Canvas` as the container noun across apps/web and mcp-server's HTTP side (~3800 occurrences, most of them correct uses of the spatial surface). This is what the vocabulary rule's standing "fix it where you already are" clause is for — a sweep would be unreviewable.
- The blob tree's `{dataDir}/blobs/{workspaceId}/canvas/{documentId}.loro` segment: moving it means walking every workspace's blobs at boot, which is a migration rather than a rename.
- Workspace-level metadata for spatial documents, the replacement ADR-0009 decision 3 names for the `type`/`tags` it removes. It is unbuilt, and this PR is the increment that makes its absence real.
- **A workspace-level display-name API for the daemon, and the Properties bar that needs it.** The daemon's canvas summary carries only a path, so there is nothing for a title box to show and nothing to write it back through. This is the follow-up that lets #837's daemon identity slot come back, built on the name rather than on a facet.
- **`OpenCanvas` as a format name.** The spec this repo implements is [JSON Canvas 1.0](https://jsoncanvas.org/spec/1.0/) — `packages/model/src/spatial.ts` cites the URL and the whole `packages/codec/src/spatial/` tree says so. `OpenCanvas` appears in prose and in published manifests (npm description + `keywords`, marketplace, `server.json`, the Gemini extension) with no definition anywhere in the repo, and it collides with an unrelated infinite-canvas interchange spec, OCIF. Correcting it touches a published surface, so per the vocabulary rule it is its own increment with its own review. (Note for whoever takes it: 193 of the 262 occurrences are `onOpenCanvas`-style React handlers, which are correct and must not be swept.)

**Resolved since an earlier version of this section.** It listed `spatial-canvas.properties.test.ts`'s seed `1810856751` failure as a pre-existing defect deferred to its own increment. #831 landed on `main` and restated the property to hold only where the body can actually render; that fix is merged into this branch, and re-running the suite with `withDefaults({ seed: 1810856751 })` pinned passes 11/11.

## A note on AI review

CodeRabbit has skipped every commit on this PR — *"This PR contains 701 files, which is 601 over the limit of 100"* — while still reporting its check as `success`. That is a size limit, not a rate limit, so waiting does not help. There is no AI review on this change; `verify` and CodeQL are the gate.

https://claude.ai/code/session_01SGjid2m1cEL6t1VtNpHU2D


kamiazya:
Review from the session that shipped the earlier half of this convergence (#764/#768/#770/#774/#780), so most of what this PR supersedes is mine. **No conceptual conflict found — where it overwrites my work it is right to.** Findings and one question below, plus two conclusions I nearly published and had to withdraw.

## Verification method

Ran the full root `pnpm test` on `origin/main` (141b3dab) and on this branch, same machine, same command. The branch is 0 commits behind main, so this is a like-for-like comparison rather than one against a stale base.

| | Test Files | Tests |
| --- | --- | --- |
| `origin/main` | 1 failed / 774 passed | 1 failed / 7806 passed |
| this branch | 7 failed / 771 passed | — |

## Two conclusions I withdrew

Both looked like PR-introduced regressions and neither is. Recording them because the method is the point — a single failing run says nothing until you run the control.

**1. `spatial-canvas.properties.test.ts` — `Property failed after 96 tests`, `expected [ 'Card' ] to include 'Body'`.** Not a timeout, a real assertion, on a property that already exists on main (`a resolved markdown body always outranks a resolved facet card`). It looked like this PR's `resolveReference` unification breaking a precedence rule.

Control: pinned `withDefaults({ seed: 1810856751 })` and ran it on **main**. Main fails identically. Main's full-suite run simply never drew that seed.

**This is a pre-existing latent defect on main, unrelated to this PR** — and worth its own fix, since the precedence rule really is violated for some canvas. Reproduce with that seed.

**2. `pixel-golden.browser.test.ts` ×2 — `toMatchScreenshot()`.** A rename PR moving pixels would be serious. Control: `pnpm test --project canvas-render-browser` alone passes on **both** sides (2 files each). Load-dependent in the full run only.

`App.test.tsx` ×3 and `sse-shared-worker.test.ts` also pass when run alone on this branch. Everything remaining on both sides is `Test timed out` / `Hook timed out`.

**So: I could not demonstrate any failure this PR introduces.**

## The test plan's numbers do not reproduce

> Nine failures remain and are all pre-existing on `origin/main` (`CanvasThumb`, `MergeDialog`, `VersionThumbnail` ×4, `VersionTimeline`, `daemon-file-adapter` — all objectURL/thumbnail tests), verified unchanged in count and identity throughout.

**None of those nine failed on either side.** `apps/web`'s jsdom project alone is 193/193 green on main; the full run's failures are a different set entirely, and differ between runs.

I do not read this as the claim being false — I read it as the suite not currently supporting a claim of that shape. "Unchanged in count and identity" is not a stable property of this suite on this machine right now. Worth re-measuring before merge, or softening to what the run can actually show.

## Findings

**1. The ADRs are left behind.** No ADR is touched, but ADR-0008 is still **Accepted** and titled *"**Slug** derivation, rename, and sibling uniqueness"* — while this PR renames `slug` to `path` and makes the word fail a test. A reader hitting the ADR index sees the decision as current and the word as banned. ADR-0007 (*"the daemon's two-store split"*) is in the same position. An ADR is history and should not be rewritten, but this repo's convention is a dated superseding note; `domain-model.md` already says *"ADR-0007 predates that rename"*, so the awareness is there and just did not reach the ADRs or the index.

**2. `refactor:` is too weak a title.** The PR notes this itself; I would put it more strongly. This carries two DB migrations, two behaviour changes, and a break in a published interchange format (`x-whiteboard.canvasId` → `documentId`, embedded in exported PNGs and JSON Canvas files). Under release-please, `refactor:` produces no version bump **and no changelog entry** — a consumer of `@kamiazya/whiteboard-mcp` gets the renamed field with no release note. `feat!:` fits what this does.

## Question, not a blocker

**`packages/crdt`.** The reasoning for rejecting `workspace` is sound — the domain word is taken and this package knows nothing about the tree. But the package is the *bridge between the model and the CRDT*, not the CRDT; every sibling is named for a role (`model`, `codec`, `ports`, `render`, `viewer`, `server-core`) while this one is named for its storage technology, which also binds the name to Loro. Worth one pass over alternatives (`doc-bridge`, `document-store`?) before it sets.

## What I checked and found right

- **The critique of my #780 is accurate.** I verified it: main's spatial branch still passes `meta`/`onChange`, so editing a title on a spatial canvas still wrote core facets. I removed the affordance and left the write. Splitting `title`/`onTitleChange` from `facets`/`onFacetsChange`, and passing no facets at all on the spatial branch, is the shape I considered and wrongly judged too large.
- **Making the bad state unrepresentable beats guarding it.** `storedCoreFacetsSchema` omitting `title` is strictly stronger than the write-time guard I put in `wb_document_set`, and `readCoreFacets` answering `undefined` for a spatial document is stronger than my `wb_facet_set` refusal.
- Both of my contracts survive intact, comments included — the facet-set spatial refusal and the OKF `title` → workspace-name routing.
- `x-whiteboard.canvasId` → `documentId` is a deliberate, documented break with the rationale in the schema comment, and the published JSON Schema is regenerated.
- `vocabulary-check.test.ts` is the right shape for an executable rule: it bans only a word with no legitimate meaning left, and says explicitly why `canvas` can never join it.

kamiazya:
Reviewed for conceptual conflicts with the converged model (#795–#810, ADR-0009). Verdict: **the direction is right and this is the legitimate completion of that line** — removing the place a wrong state can live (`storedCoreFacetsSchema` omitting `title`; spatial facets made unwritable rather than hidden) is exactly this repo's idiom, and the mechanical `slug` ban follows the executable-over-prose ladder. Three items need resolving before merge, one of them a real conceptual gap.

## 1. Blocker: eight Loro binaries committed at the repo root — find the writer before deleting them

`123-design`, `621/footer`, `621/header`, `621/header-v2/layout`, `622/a/b`, `a`, `abc`, `again` all carry the `loro\0…LORO` snapshot magic, and their names are document *paths* (the nested ones created real directories). The canonical blob location is `{dataDir}/blobs/{workspaceId}/canvas/{documentId}.loro`; no legitimate code path writes a path-named, extensionless snapshot into cwd. Deleting them is trivial — the important part is identifying **what wrote them**. If something in this branch resolves an empty/undefined dataDir to cwd, that is a data-placement bug that will scatter user documents, and the deletion would bury the only evidence. Please identify the writer (or demonstrate it was a one-off harness misconfiguration, e.g. a probe run with the wrong `WHITEBOARD_DATA_DIR`) and state it in the PR before removing the files.

## 2. Conceptual gap: the `kind`-vs-`format` decision lives in the wrong document

vocabulary.md now argues — persuasively — that ADR-0009's word `format` lost to `kind` because every stored and published shape already says `kind`. But ADR-0009 itself is untouched, which leaves two normative documents disagreeing, with the *rule* overriding the *decision record*. That inverts the authority order this repo relies on (docs-sync honesty: document the shipped state where the decision lives). A short addendum on ADR-0009 recording "`format` is spelled `kind` throughout; rationale in vocabulary.md" closes it. Same-increment work, a few lines.

## 3. CI is red on the current head

- `test-browser`: `BrowserLocalCanvasPage.multi-canvas` cannot find a UUID text in the switcher — needs a verdict on whether the unnamed-document list rendering changed (real regression) or the test lags the rename. The branch is 0 behind main, so this is not the pre-existing set named in the PR body.
- `packaged-smoke` is also failing and `verify` shows skipping; the merge gate needs all three green.

## Judgement calls I reviewed and support

- **`packages/crdt` over `packages/workspace`**: the collision argument (comments reading "workspace's withSpatialBatch", indistinguishable from correct domain uses) is evidence-backed; the package's role IS the merge-aware bridge, so the mechanism name is honest here.
- **`x-whiteboard.documentId` as a plain break** (old exports lose their embeds, including PNG-embedded JSON Canvas): consistent with the standing 0.0.x decision that model consistency outranks reading old data, and the consequence is stated in vocabulary.md rather than discovered later. Acknowledged, not objected.
- **Migrations 0009/0010** continue 0008's discipline (log keys as history, legacy-upgrade chain test).

## Sequencing note for the integrator

At +7,970/−6,663 touching ~300 files, this must merge **before** every in-flight branch (wave-*, selection-polish all conflict wholesale). Landing anything else first forces this PR through another full-rename rebase.

Once 1–3 are addressed I'm in favor of merging.

kamiazya:
Both reviews addressed. Taking the blocker first, then the rest in order.

## Blocker: the root Loro binaries — writer identified and reproduced

You asked for the writer before the deletion, and I had already deleted them (`ab2f4c0`) — so here is the reproduction instead of an assertion. It is repeatable from the current head.

**Not a dataDir bug.** `getDataDir()` is uninvolved. The writer was `saveCanvas`'s snapshot write taking the **document path** instead of the blob path — a variable-shadowing slip during the `slug` → `path` sweep, where the pre-existing `const path = canvasBlobPath(...)` collided with the newly-renamed `path` **parameter**. `writeFile(path, snapshot)` then type-checked perfectly and wrote a bare relative path from cwd.

Reproduced by reintroducing exactly that one line in `document-store.ts` and running the store tests:

```
?? 123-design   ?? a   ?? abc   ?? again   ?? b   ?? big   ?? broken   ?? cached
?? canvas-a     ?? foo ?? my-canvas  …
$ head -c 24 123-design | cat -v
loro^@^@^@^@^@^@^@^@^@^@^@^@S.^LM-|^@^C/^@
```

Same names, same magic. Reverted, and the artifacts removed.

**It is guarded, loudly.** With that line reintroduced, `pnpm test --project mcp-node` fails **62 tests**. So the debris was produced by runs that were *already red* mid-sweep, and the fix landed in the same commit that caused it (`c00443e`, five shadowed sites renamed to `blobPath`). A clean `mcp-node` run now leaves zero new untracked files at the root.

What nothing covered was the artifacts reaching a commit, which is `repo-root-files.test.ts` — `git ls-files` rather than a tree scan, so it fails on junk being *tracked* and not on anyone's local scratch file. Mutation-checked both directions.

## The `kind`-vs-`format` authority inversion — fixed

Agreed, and it was the more interesting of the two ADR findings. `67f5eac`:

- **ADR-0009** gains a dated addendum: `kind` is the spelling, one-line reason, pointer to the rule for the longer rationale. It also records that decisions 3–4 — which its own Consequences describe as a target — are now implemented, and that the workspace-level metadata it names as the replacement for a spatial document's `type`/`tags` is still unbuilt.
- **ADR-0007 / ADR-0008** gain vocabulary notes (*read slug as path*). Both decisions stand; only the word moved. Worth a note precisely because `slug` now **fails a test**, so a reader meeting it in an Accepted decision cannot otherwise tell history from current guidance. The index carries the same signal.

No decision text rewritten — ADR-0003 sets the dated-superseding-note convention and I followed it.

## CI

Both fixed before your reviews landed; the head you read was stale.

- **`packaged-smoke` was a real regression of mine.** `400 {"title":"path is required"}` — the E2E smokes still POSTed `{ slug }`, and `pnpm smoke:e2e` still read `created.canvasId`. Both survived because my sweep *and* the `slug` guard scanned `apps/web/src` and `packages/*/src` only; nothing typechecks a `.mjs` smoke against the contract it calls. The guard now also scans `tests/`, `apps/web/scripts`, `packages/*/scripts`, reads `.mjs`/`.js`, and **throws rather than skips** on a missing scan root. Verified by running them: every packaged smoke exits 0, `smoke:e2e` reaches `ALL OK`.
- **`test-browser` was the seventh flake shape** in `integrator-flow.md`, not a rendering change. Measured from the CI failure's own DOM dump rather than guessed: zero `[role="menu"]`, every `aria-expanded="false"` — no list was ever opened. The switcher's non-modal menu was still dismissing from the "New canvas" click above, so the re-open click was consumed. Added the same wait the markdown suite already carries.

## The test-plan numbers — you were right, and the cause is worse than the wording

Re-measured. Two separate problems:

1. Those nine fail **on this sandbox only**, on `origin/main` and on this branch alike — environment-specific, not a property of the repo. My wording implied the latter.
2. More importantly: `pnpm test` here reports `Test Files 5 failed | 660 passed (779)`. **114 files never ran.** The browser projects cannot launch in this sandbox and vitest reports that as *one* `Unhandled Error` while every other project reports green — which reads almost exactly like a full green run. Same class as the `--project` trap `dev-flow.md` already documents, in a shape it does not yet name.

The PR body now says what the runs actually support: the node projects green (3470), the browser projects run **separately** with an explicit chromium path (588), CI authoritative for the rest. Worth adding this second shape to `integrator-flow.md`'s list in its own increment — it is a general trap, not mine.

## Title

Retitled **`feat!:`**. Your changelog argument is the decisive one: `refactor:` gives no version bump *and no release note*, so a consumer would meet the renamed `x-whiteboard` field with nothing pointing at it.

## `packages/crdt` — your two reviews disagree, so I have left it

One review supports it as an honest mechanism name for a package whose role *is* the merge-aware bridge; the other asks for a pass over role-shaped alternatives. Both fair. Two facts for whoever settles it: `document-store` collides twice (mcp-server's `document-store.ts` and `ports`' `DocumentStore` port), and `doc-bridge` is the only survivor I found that is role-shaped without colliding. Happy to rename on request — it is one increment and nothing outside the repo sees it.

## Recorded, not fixed here

`spatial-canvas.properties.test.ts` at seed `1810856751`. Your control run against main is the convincing part; it is a pre-existing precedence defect and fixing it inside a 690-file rename would bury it.

Thank you for running the control on both sides — the withdrawn conclusions were more useful than the findings.

---

PR #862, opened 2026-08-17
feat(contracts)!: stop guessing spatial for a document that records no kind

kamiazya:
## What

Identity-convergence **slice 6** (explicitly approved for this initiative). The canvas summary defaulted an absent `kind` to `spatial`, so **no client could tell a document stored as spatial from one whose kind was never recorded** — the third of the three stores that disagreed about "unknown" (`document-format-three-stores-three-answers`).

- `canvasSummarySchema.kind` is now **optional, not defaulted**.
- `listDocuments` omits `kind` for a row that records none. Measured while doing this: `saveDocument` stores **NULL** for an omitted kind (`kind: options.kind ?? null`) — so unrecorded rows are still produced today, not only by pre-kind history.
- `apps/web`'s `CanvasRow` / `CanvasListView` carry an optional kind: the gallery's `markdown ·` badge simply does not appear for an unrecorded kind, and `DaemonCanvasPage` already defaults locally when choosing an editor. **Rendered output is unchanged** — a NULL-kind row rendered exactly this way before, via the guess. Only the claim in the data is gone.

The **create request** schema keeps its `spatial` default, deliberately: a caller omitting kind at creation is *choosing* spatial, which is an answer, not a guess about stored state.

## Verification

- Two obsolete pins rewritten (they asserted the old guess: the summary schema's default and `listDocuments`' back-compat claim), plus a third in `apps/web`'s daemon-api-client test — each replaced with the new truth **and** a companion pin that a recorded kind is still reported unchanged.
- **Mutation check**: restoring `kind: r.kind ?? 'spatial'` turns exactly the two no-guess pins red; restore returns green.
- mcp-node 3548 green (290 files); apps/web jsdom 2245 green (198 files); typecheck across the monorepo clean (it is what enumerated every consumer that had to decide); knip clean.

Closes the last code slice of the convergence plan; only the ADR-0007 as-built addendum (slice 7) remains.


kamiazya:
Rebased onto main after the Canvas→Document vocabulary rename landed (#864-era). Conflicts were in two test files only; resolved by taking main's new vocabulary (`documentSummarySchema`, `listDocuments`, the `documents` payload key) and re-applying this PR's semantics on top — an unrecorded kind stays absent. Re-verified after the merge: mcp-node 3548 green, apps/web jsdom 2265 green, monorepo typecheck clean.

PR #866, opened 2026-08-18
feat(canvas-render): break CJK lines with UAX #14 + BudouX, and cut what cannot wrap

kamiazya:
Text with no ASCII spaces in it could not wrap **at all**. `emit` required a space to be present before it would even consider wrapping, and the wrapper split on `' '`, so a Japanese paragraph was emitted as one run and painted straight through its node's border and across whatever sat to the right of it.

## Visual repro

The same canvas rendered through the real pipeline at each step. Left: today. Middle: line breaking. Right: phrase-aware line breaking.

A label cannot wrap — one line is what makes it a label — so it is cut and the cut edge is faded. Rendered through the PNG export path (resvg), not a mock:

## The scoreboard

`layout/text-wrapping-quality.test.ts` landed **first, in its own commit**, before any wrapping changed (per `.claude/skills/measured-change`): 11 corpus cases x 3 narrow widths, every number pinned EXACTLY so an improvement is as loud as a regression. Its measurer charges CJK a full em, unlike the shared 0.6em/char fake, which understates Japanese by ~40% — the single number it exists to report.

| | before | 1. UAX #14 | 2. BudouX | 3. cut + fade |
|---|---|---|---|---|
| overflowing runs | 33 | 3 | 3 | **0** |
| worst overflow | 704px | 277px | 277px | **0px** |
| blocks whose bbox under-reports their ink | 28 | **0** | 0 | 0 |
| `measure` calls (price) | 257 | 1252 | 873 | 917 |

All three debt metrics are now zero, and pinned at zero.

## What changed

**Break opportunities come from `css-line-break`** (UAX #14 with CSS `line-break: strict`) — the same algorithm a browser applies to a `<p>`, which is also where Japanese kinsoku lives, so a closing character never opens a line and an opening character never ends one. One small pure DOM-free dependency instead of a hand-rolled character table.

**BudouX narrows those to phrase (文節) boundaries for Japanese**, a strict subset, so preferring them costs nothing in fit and buys a line that breaks where a reader would pause. Candidates are a three-level cascade — phrase, then UAX #14 segments, then code points — stepping down only for a segment that will not fit a line of its own. It also made layout *cheaper* (1252 → 873 measure calls): phrases are coarser than characters, so the packing loop iterates less.

**What cannot wrap is CUT, not left to overflow** (`layout/truncate.ts`): a node label and an atomic run keep the longest prefix that fits and are marked `truncated`, which the SVG backend paints as a fade. A fade rather than an ellipsis because a label is cut precisely where width is scarce, and three dots spend the width they save.

**A block's declared width now covers its ink.** A block claiming `maxWidth` while a run painted past it is why `sceneBounds`, the export viewBox and the editor's grow-only auto-fit could all agree on a size nothing actually fitted in.

## Deliberate carve-outs

- **`wordBreak: 'normal'`**, so English is never broken mid-word. Only a segment that alone exceeds the width is expanded to code points.
- **A line is ONE run.** A run per break opportunity also fits, and multiplies the SVG's `<text>` elements by the character count of every CJK paragraph.
- **Inline MATH is neither split nor cut.** `a + b + c` cut to `a + b` reads as a complete formula that is simply wrong, where cut code reads as cut. It is the one thing still allowed to overflow.
- **An edge label is not cut** — it floats on the edge rather than inside a box, so there is no width to fit it to.
- **`sceneDigest` does not report truncation.** A fade says "there is more" without saying how much, and the digest carries no text content to be inconsistent with. Add it when a reader has a reason to act on it.
- **Chinese and Korean stay on UAX #14.** BudouX ships a model per script and there is no evidence yet that this package needs them.

## Cost, measured

`mdast-blocks.bench.ts` (new): 0.062ms → 0.745ms for all 33 corpus layouts (~23us each), interleaved paired rounds with the stash verified. That is noise beside edge routing's 46-405ms per canvas, so exactness is kept over a cheaper approximation; a `ponytail:` comment names binary search as the upgrade path if text layout ever shows up in a profile.

The BudouX parser is built on **first Japanese text**, not at module load — its constructor turns a ~24KB model into a Map, and charging that to whichever lazily-imported chunk pulls the module in cost two apps/web browser tests their 1000ms `findBy` budget. Lazy init took the suite back to green.

## Tests, and the ones that changed

New: the scoreboard, `mdast-blocks-wrapping.test.ts`, `label-truncation.test.ts`, `svg/fade.test.ts`, a `web-browser` regression for a Japanese note and for a cut label, and a bench.

Existing tests pinned consequences of the old behaviour and were rewritten to the new contract rather than relaxed — the unbreakable-token example now asserts code-point splitting; the wrap-vs-sibling-space example needed a width where the wrapped remainder and the sibling actually share a line, since the old one only "shared" it by overflowing; two spatial-canvas counterexamples were re-shrunk from 17px wide to 200px, because at a 1px content width every body fails to render and that width no longer said anything about card-vs-body ranking. The word-preservation property is now **two** properties, so the un-truncated case keeps its full strength and a truncated run is separately held to "only ever a prefix".

Mutation-checked four times (wrap branch disabled, browser regression, prefix property inverted to a suffix, label truncation disabled), each with the restore verified by grep before the green re-run.

## Known-flaky, not from this change

`pnpm test:browser` leaves 2 failures in `apps/web`'s markdown browser tests. Verified by A/B in a clean worktree: **`origin/main` fails one test from the same family too** (a different one each run). Filed locally with the measurements; not touched here.

## Summary by CodeRabbit

* **New Features**
  * Improved text wrapping for Japanese, Chinese, mixed-language, and long unbroken text.
  * Added width-aware label truncation that preserves readable prefixes.
  * Added fade effects for truncated labels in rendered diagrams.
  * Improved handling of punctuation, inline code, mathematical expressions, URLs, and oversized glyphs.

* **Bug Fixes**
  * Prevented text from exceeding node boundaries while preserving layout accuracy.
  * Improved Japanese line-break behavior and label rendering in spatial editor nodes.


kamiazya:
## Review triage

**CodeQL `js/html-constructed-from-input` (alerts 36/37/38, `svg/backend.ts:480,495`) — pre-existing, no change.**

These are the same finding that is already open on `main` as alerts **33/34/35** at `svg/backend.ts:454`, plus **28** at `:439` — the two `return \`<svg …>\`` lines. This PR only inserted `${defs}` into those template literals, which shifted the line numbers and re-reported them against the new lines. The interpolated `defs` is a module constant with no input in it at all.

The mitigation is the one the package already documents: `svg/format.ts` is the sole producer of this string and escapes `&`/`<`/`>` in text and `"`/`'` in attribute values, which is also why `CanvasViewer.tsx` can use `dangerouslySetInnerHTML` without a sanitizer. Nothing here is a new sink, so this PR adds no fix; the standing alerts are unchanged either way.

**CodeRabbit on `mdast-blocks.properties.test.ts` — valid, fixed** in 78960e43.

The property asserted `tokens(run.text)[0]`, so a truncator that corrupted anything after the first retained word would have passed it. It now requires an atomic source value to start with the **whole** retained `run.text`, over a helper that collects exactly the values that can be truncated (`inlineCode` / `html` / `inlineMath`; prose wraps instead). Mutation-checked against precisely that hole: appending one character to the cut text fails it.

**`packaged-smoke` — real, and the reason for the extra commit.**

`budoux`'s only public entry re-exports `HTMLProcessingParser` → `linkedom` → the native `canvas` package, so the published bundle failed to build. BudouX is now vendored (Apache-2.0, headers intact) and the dependency is gone; `pnpm build` and the full suite pass. Details in the commit message and `src/vendor/budoux/README.md`.

**`test-unit (1)` — known CI flake.** `app.test.ts > echoes a supported client protocol version during initialize` timed out at 10s under shard load; it passes locally (`mcp-node` 3545 tests) and `main`'s own most recent CI run failed on `test-unit (2)` the same way.

PR #868, opened 2026-08-18
feat(web): one link verb, one picker — documents and URLs on the same surface

kamiazya:
Follow-up to #865. The editor's link verb wrapped text in `[[ ]]` and nothing more: you had to know the target's name, type it exactly, and hope it was unique. An external URL had no verb at all.

## One surface, because the author knows the destination and not our taxonomy

The search box decides what the verb does:

| what you type | what the list offers |
|---|---|
| part of a name | this workspace's documents, ranked exact → prefix → contained |
| something URL-shaped | `<url> · External` at the top of the same list |
| nothing | every document |

Nothing has to be classified before it is typed — the alternative (two verbs) makes you pick a category while you already know the destination. The box is **seeded with the word the verb would otherwise have wrapped**, so the pre-picker gesture still ends in one Enter instead of becoming a search someone has to retype.

| document mode | URL mode |
|---|---|
|  |  |

## Readable form where it works, stable form where it does not

`createSnapshotAliasResolver` resolves `[[Name]]` only when **exactly one** document carries that name, so writing the readable form for a duplicate produces a link that silently stays literal text. The picker is the one place that KNOWS which document was chosen:

- `[[Weekly review]]` normally
- `[[canvas:<id>]]` when two documents share the name, or the name contains `]]` / a newline and cannot survive between brackets

External links go in as `[text](url)`, or a bare `<url>` autolink when there is no text to carry them. A destination containing spaces or parens is wrapped in angle brackets (CommonMark's own answer). Only **http(s)** — a `javascript:` or `file:` target typed into a search box is never what an author meant, and this writes into a document other people open.

## Wiring

Both composition roots already hold the list their switcher shows, so each supplies it (`BrowserLocalDocumentPage` from `documents`, `DaemonDocumentPage` from `controller.documents`). With **no** list supplied the verb keeps its selection-free bracket wrap — a picker onto an empty list is a dead end, not a feature.

## Tests

- `link-target.test.ts` (15, jsdom) — ranking, the ambiguity/unwritable-name fallbacks, URL detection (including the rejected schemes and the bare-domain promotion), and the external-link markup rules. Red before the module existed.
- `link-picker.browser.test.tsx` (6, `web-browser`) — the seeded query committing with one Enter; narrowing and picking; the id fallback for duplicate names; the URL row; Escape leaving the document untouched; and the no-targets fallback.
- Mutation-checked through the real UI: forcing `unambiguous = true` turns exactly the duplicate-name test red.
- `editor-catalog.browser.test.tsx`'s link case now names the verb `Link` and pins the no-targets fallback.
- Green: `pnpm test --project web-browser` (627), `pnpm --filter @kamiazya/whiteboard-web test` (2281 + 77), `pnpm -r typecheck`, `pnpm lint`, `pnpm knip`.

Deliberately NOT in this increment: `[[` autocomplete in the source pane (a keyboard shortcut to the same picker, needs `@codemirror/autocomplete`), and rename-following for existing links.

https://claude.ai/code/session_01Ua6RPuoEAD4ZTBqyhNC56Z

## Summary by CodeRabbit

* **New Features**
  * Added a link picker for searching documents and inserting internal or external links.
  * Added keyboard navigation, URL detection, filtering, and safe link formatting.
  * Improved link resolution for documents with duplicate names.
* **Improvements**
  * Renamed the toolbar control to “Editing actions” for clearer accessibility.
  * Preserved selection, focus, and cursor placement when inserting or cancelling links.
* **Tests**
  * Added comprehensive coverage for link picking, ranking, navigation, URL handling, and fallback behavior.

### DIFF
--- a/apps/web/src/components/document-editor/DocumentEditorSurface.tsx
+++ b/apps/web/src/components/document-editor/DocumentEditorSurface.tsx
@@ -34,6 +34,7 @@ export interface MarkdownDocumentSession
     | 'theme'
     | 'meta'
     | 'resolveAlias'
+    | 'linkTargets'
     | 'onOpenDocument'
     | 'resolveEmbed'
     | 'autoFocus'
--- a/apps/web/src/components/markdown-editor/EditorToolbar.tsx
+++ b/apps/web/src/components/markdown-editor/EditorToolbar.tsx
@@ -38,6 +38,10 @@ const MODE_OPTIONS: readonly ModeOption[] = [
  * thing: **how this document is shown** (word count + view mode), plus the
  * doorway to everything that CHANGES the document.
  *
+ * The doorway's name is "Editing actions" rather than the generic "More
+ * actions" the app shell's own ⋯ uses: two controls with the same accessible
+ * name on one screen are indistinguishable to anyone reading it aloud.
+ *
  * Formatting buttons used to sit here and no longer do. They were redundant
  * for the audience that could reach them (⌘B already exists) and out of
  * reach for the one that needed them — a 28px target at the top edge is the
@@ -96,7 +100,7 @@ export function EditorToolbar({
             <TooltipTrigger asChild>
               <button
                 type="button"
-                aria-label="More actions"
+                aria-label="Editing actions"
                 aria-haspopup="menu"
                 data-testid="editor-catalog-trigger"
                 // Keep focus (and therefore the caret) in the source pane: the
@@ -112,7 +116,7 @@ export function EditorToolbar({
                 <MoreHorizontal aria-hidden className="size-4" />
               </button>
             </TooltipTrigger>
-            <TooltipContent>More actions</TooltipContent>
+            <TooltipContent>Editing actions</TooltipContent>
           </Tooltip>
         )}
       </div>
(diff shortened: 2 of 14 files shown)


kamiazya:
Thanks — five of the six landed in 8d51a412, each with the test that would have caught it.

**Taken:**
- **Pinned range mapped, not clamped.** Reproduced first: typing `SEE ` at the document start while the picker was open produced `[[Weekly review]]ekly beta`. The range now lives in a CodeMirror `StateField` that maps through every change set. One correction to the suggested shape — the bias has to point **inward** (`+1` at the start, `-1` at the end); with `mapPos(from, -1)` an insertion at the start stays inside the range and the link swallows it, which is exactly what the new test caught on the first attempt.
- **`|` and a leading `canvas:` added to the unwritable-name guard.** Verified against `codec/src/references/resolve.ts`: the target scan stops at `|` (line 47), and a `canvas:`-prefixed target is parsed as a direct id and then fails `documentIdSchema`, dropping the link entirely.
- **Backslash escaped before the brackets**, so a text ending in `\` no longer escapes the closing bracket the escape itself adds.
- **`<`/`>` percent-encoded** when the destination takes the angle-bracket form.
- **`host:port` promoted.** Confirmed `example.com:8080/path` parsed with protocol `example.com:` and returned null; a second parse behind `https://` fixes it.
- **Picker reads the switcher's effective list** (snapshot overlaid on the async one), so a rename in flight cannot make it offer a stale name or omit the current document.

**Not taken:** promoting the page-level wiring guard to E2E. It is a single assertion in the file that already covers this page with the same store + router harness (the `[[Name]]` resolution case immediately below it), and its only job is to fail when the page stops passing `linkTargets` — the prop is optional, so nothing else would. An E2E run is disproportionate to that, and moving it would separate it from the sibling case it mirrors.

The first comment's other half is also acted on as written: `createSnapshotAliasResolver` matches exactly and case-sensitively, so `sameName` stays as it is, and `#` was not added.

PR #875, opened 2026-08-18
test(web): keep browser test titles short enough that a failure keeps its file

kamiazya:
A failing browser test writes a Playwright trace, and vitest then copies it into `.vitest-attachments/` under a name flattened from its path. Past the filesystem's 255-byte limit that copy throws `ENAMETOOLONG` in the file's teardown — and **vitest abandons the rest of the file**.

## Measured

The same forced failure in `BrowserLocalDocumentPage.rename`, twice:

| failing test's title | result |
|---|---|
| 194 chars | `ENAMETOOLONG` → **`1 failed \| 2 passed (6)`** |
| 58 chars | no error → `1 failed \| 5 passed (6)` |

Three tests silently did not run, and the run reported a **smaller total** — which reads like good news. It is the same trap AGENTS.md already warns about for filtered runs, and it is why a load-flaked run showed `1 failed | 630 passed (634)`.

The trace itself was never lost: it is written to `tmp/vitest-traces/` first and only the copy fails (verified — the zip opens, 6 entries, trace + network + resources). So this cost coverage, not diagnosability, which is what makes the fix about the abandoned tests rather than about the error message.

## The guard

`browser-test-name-length.test.ts` computes the projected attachment name for every `describe`+`it` path across the browser suites and fails past the budget. Two details worth knowing:

- It reads source at build time via `?raw`, so apps/web stays free of `node:fs` (`web-app-boundary.test.ts`), following `App.lazy-coverage.test.ts`.
- It resolves `describe` nesting by brace depth. A sibling `describe` opens at the depth the previous one closed at, so a naive stack reports a title path nobody wrote — the first version of this guard produced `SpatialEditor (browser)-SpatialEditor (browser) — spatial text layout defects-…` and over-counted by tens of characters.

Mutation-checked: restoring the 194-char title fails it. It also caught a real slip during this change — a `git checkout --` reverted one rename and the guard failed on exactly that title.

## The renames

Thirteen titles were over. The redundant part came out first: `(browser — real IndexedDB)` says "browser" in a `*.browser.test.tsx` file whose project already appears as `web-browser--chromium--` in the attachment name. The rest were shortened without losing what they claim. **No test's subject changed** — only its wording.

## Verification

`pnpm --filter @kamiazya/whiteboard-web test` (2314 + 77), `pnpm test:browser`, `typecheck`, `knip`, `biome` pass. End-to-end: forcing the failure again on the renamed title writes the attachment successfully and reports `1 failed | 5 passed (6)`.

## Not fixed here

`test:browser` still loses one test to load under a saturated machine (`New markdown note… …` times out with focus still on `<body>`). That is the separate, already-documented contention problem in `vitest.browser.config.ts`; this PR only stops one failure from taking its file's other tests with it. Traces for it are now obtainable, which is the prerequisite for chasing it.

## Summary by CodeRabbit

* **Tests**
  * Added validation to ensure browser test names stay within filesystem limits, preventing trace-related failures and protecting subsequent test runs.
  * Improved coverage for nested test titles, escaped text, and test discovery.

* **Documentation**
  * Documented browser test naming limits, related trace attachment errors, and their impact on test execution.

* **Chores**
  * Simplified and clarified browser test descriptions across workspace, document, migration, and canvas workflows without changing behavior.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -93,6 +93,15 @@ pnpm run test:browser:trace  # same, with trace artifacts on failure
 
 - Failure traces are stored under `<package>/tmp/vitest-traces`.
 - Check traces before adding temporary debug code.
+- **Keep a browser test's `describe` + `it` titles under 155 characters
+  combined.** vitest copies the trace into `.vitest-attachments/` under a name
+  flattened from its path, and past the filesystem's 255-byte limit that copy
+  throws `ENAMETOOLONG` during teardown — so vitest abandons the REST OF THE
+  FILE. Measured: one forced failure with a 194-char title reported
+  `1 failed | 2 passed (6)`, the same failure with a 58-char title reported
+  `1 failed | 5 passed (6)`. Three tests silently did not run, and the smaller
+  total reads like good news. `apps/web/src/browser-test-name-length.test.ts`
+  enforces the budget.
 - Remove temporary debug overlays, logging, and instrumentation before finishing.
 
 ## MCP Development Mode
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -94,7 +94,9 @@ pnpm run test:browser:trace  # same, with trace artifacts on failure
 - Failure traces are stored under `<package>/tmp/vitest-traces`.
 - Check traces before adding temporary debug code.
 - **Keep a browser test's `describe` + `it` titles under 155 characters
-  combined.** vitest copies the trace into `.vitest-attachments/` under a name
+  combined** (characters, not UTF-8 bytes: vitest replaces every
+  non-alphanumeric character with one ASCII `-` before the name reaches the
+  filesystem, so `導線` costs two, not six). vitest copies the trace into `.vitest-attachments/` under a name
   flattened from its path, and past the filesystem's 255-byte limit that copy
   throws `ENAMETOOLONG` during teardown — so vitest abandons the REST OF THE
   FILE. Measured: one forced failure with a 194-char title reported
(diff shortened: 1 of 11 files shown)


kamiazya:
## Review triage

**"Measure title paths in UTF-8 bytes" — measured, and it would have been wrong. Fixed differently in 29f62b17.**

The filesystem limit is bytes, so the concern is right in principle. But vitest replaces every non-alphanumeric **character** with one ASCII `-` before the name reaches the filesystem, multi-byte ones included, so `導線` costs two dashes and not six. Checked against a real attachment name rather than reasoned about:

```
title path raw    207 characters / 213 UTF-8 bytes
landed on disk    207 characters   ("markdown 導線 (browser — real" -> "markdown-----browser---real")
```

Counting the raw title's bytes would reject titles that fit. The guard now measures the **sanitized** form, where characters and bytes are the same thing by construction — which answers the question and stops it recurring. Mutation-checked at the new seam, and `AGENTS.md` says which unit it is and why.

**"Run the guard in the browser test project" — not moved, deliberately.**

It is a source scan: it reads `*.browser.test.tsx` via `?raw` and counts, with nothing to run in a browser. `.claude/skills/test-layer-selection` puts a test at the narrowest layer that can hold it, so making this one boot Chromium would buy nothing and cost a browser context on every run. It is not unenforced — CI's `test-jsdom` job runs `pnpm --filter @kamiazya/whiteboard-web test`, and so does `pnpm test`. It follows `App.lazy-coverage.test.ts`, which guards browser-test behaviour from the same layer for the same reason.

It has already earned it, twice: it caught the buggy first version of its own nesting logic, and it caught a rename I lost to a `git checkout --` during this PR.

PR #878, opened 2026-08-18
fix(canvas-render): keep a text node's body inside its own frame

kamiazya:
Reported from the running app: a `text` node whose wrapped body needs more lines than its box has room for painted the last paragraph **below the frame**.

```json
{"type":"text","x":40,"y":260,"width":67,"height":51,"text":"かあらた\n\nかたそ"}
```

Reproduced exactly. In a real browser with real Canvas 2D measurement, the last paragraph's baseline lands at **y=323** against a frame bottom of **311**.

## Root cause

Three seams in `spatial-canvas.ts` put mdast blocks in a node box, and only two answered to the box's **height**:

| seam | width | height |
|---|---|---|
| `composeFileMarkdown` → `fitBodyInNode` | ✅ | ✅ |
| `composeFileFacets` → `fitBodyInNode` | ✅ | ✅ |
| **`composeTextNode`** | ✅ | ❌ |

`composeTextNode` passed `contentWidth` alone into `layoutMdastBlocks` and placed whatever came back. That is the third producer the *"one producer per geometry"* rule exists to prevent, and it contradicts this package's own *"what cannot fit is cut"* contract (`package-canvas-render.md` decisions #10 and #13).

The wrapping scoreboard could not see it either: `text-wrapping-metrics.ts` collects `x + w` only, so vertical overflow is invisible to it by construction.

## The fix

The height bound now lives in one place — `fitSceneInNode` — which both `fitBodyInNode` and the text node's parse-failure label path go through. A box with room for nothing renders as bare chrome rather than one line painted outside it.

**Product change worth calling out:** whole-block truncation means the dropped paragraph is not shown at all. That is the sibling seams' shipped policy and it replaces a worse behaviour, but it *is* visible for a node the user typed. The "there is more" affordance stays the editor-side overlay `fitBodyInNode` already names as its upgrade path.

## Coverage — two layers, each mutation-checked against a reverted fix

| layer | red without the fix |
|---|---|
| `spatial-canvas.test.ts` (node, full-em CJK simulated) | `expected 324 to be less than or equal to 303` |
| `canvas-viewer-cjk-overflow.browser.test.tsx` (real browser, real Canvas 2D) | `expected 323 to be less than or equal to 311` |

The browser one exists because `fake-measure.ts` charges CJK 0.6em and understates Japanese by ~40% — enough to hide the wrap that causes this entirely. The Node **export** path cannot see it at all: its vendored font has no CJK glyphs and measures them narrow, so the body fits on one line there. Any evidence taken from `renderSpatialCanvasToPng` would have shown a clean box in both states.

A second guard pins that the fix does not over-correct — the paragraph that *does* fit must still be painted. It caught a first attempt whose fixture collapsed the whole body into one block, where whole-block truncation had nothing to keep.

## Verification

canvas-render + canvas-viewer + mcp-server node projects **4345 passed** · canvas-viewer-browser **10 passed** · `pnpm -r typecheck` clean · `pnpm knip` clean. The SVG determinism goldens and both quality scoreboards are unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved text wrapping so content stays within its padded node boundaries.
  * Prevented overflow for multi-block text, including short content and CJK characters.
  * Standardized vertical fitting across file, markdown, and facet layouts.

* **Tests**
  * Added regression coverage for text fitting, height probing, and browser-rendered CJK text boundaries.

### DIFF
--- a/packages/canvas-render/src/layout/spatial-canvas.properties.test.ts
+++ b/packages/canvas-render/src/layout/spatial-canvas.properties.test.ts
@@ -479,3 +479,105 @@ describe('live-drag parity property (PBT)', () => {
     },
   )
 })
+
+/**
+ * The text node's vertical fit, stated as the three laws that fully describe
+ * it. Example tests pin one fixture each; these say what holds for ANY body
+ * in ANY box — which is the difference that matters here, because the first
+ * shape of this fix passed three examples while breaking two cases nobody
+ * had written an example for.
+ *
+ * `fullWidth` charges CJK a full em (the scoreboard's measurer, not
+ * `fake-measure.ts`'s uniform 0.6em) so a wrap is reachable at these widths.
+ */
+const fullWidth = createFakeMeasure(1)
+
+function parseParagraphs(text: string): MdastRoot {
+  return {
+    type: 'root',
+    children: text.split(/\n\s*\n/).map((para) => ({
+      type: 'paragraph',
+      children: [{ type: 'text', value: para }],
+    })),
+  }
+}
+
+const fitOptions = { measure: fullWidth, parseBody: parseParagraphs, appearance }
+const PADDING_PX = 8
+
+/** The laid-out blocks, i.e. everything the node paints that is not chrome. */
+function blocksOf(canvas: SpatialCanvas): ReadonlyArray<{ y: number; h: number }> {
+  return layoutSpatialCanvas(canvas, fitOptions)
+    .nodes.filter((entry) => entry.kind !== 'shape' && entry.kind !== 'edge')
+    .map((entry) => ({ y: entry.bbox.y, h: entry.bbox.h }))
+}
+
+function textNodeAt(width: number, height: number, text: string): SpatialCanvas {
+  return { nodes: [{ id: 'n', type: 'text', x: 0, y: 0, width, height, text }], edges: [] }
+}
+
+/** Bodies that wrap and bodies that do not, in one to five blocks. */
+const bodyArb = fc
+  .array(fc.constantFrom('かあらた', 'かたそ', 'short', 'a longer english line'), {
+    minLength: 1,
+    maxLength: 5,
+  })
+  .map((paras) => paras.join('\n\n'))
+
+const widthArb = fc.integer({ min: 24, max: 200 })
+
+/**
+ * A fraction OF THE NATURAL CONTENT HEIGHT, never an absolute pixel range: a
+ * uniform height would almost always be roomy enough to fit everything and
+ * the properties would pass vacuously. Straddling 0..1.2 puts the generator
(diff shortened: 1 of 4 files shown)


kamiazya:
CI caught two **over-corrections** in the first commit, both from treating "nothing fits" as "render nothing". Pushed as a second commit; both pinned and mutation-checked.

### 1. `widget-smoke` — a node one line tall is the common shape

The smoke's own fixture is `{ width: 200, height: 25, text: '日本語ラベル' }`. At the default padding that leaves **9px** of content box for a ~16px line, so the only block was dropped and the prose vanished from an ordinary label-sized node.

A text node has no lower-ranked rendering to degrade to, so the fit now never keeps fewer than one block — the same reasoning as `fitToWidth` never returning the empty string, applied to the other axis. The block that stays squeezes the *padding* rather than leaving the **frame**, which is the bound this whole change exists to keep. The sibling seams keep `undefined` because they *do* have somewhere better to go.

### 2. `test-browser` — the fix defeated the feature that already solves this

`expected 60 to be greater than 120`, from `auto-grow-height.browser.test.tsx`.

`apps/web` **already** answers overflow by growing the node, and `requiredTextNodeHeight` measures the natural content height by laying the node out at `height: 1` and reading the scene's bottom edge. Truncating there capped what that probe could report, so a box could never grow past one block — a fit meant to keep content inside the frame defeating the feature that keeps content inside the frame.

A non-positive content box is now **"no height to fit against"**, which is the reading `fitToWidth` already documents for an unusable `maxWidth`. The regression test asserts the height-1 layout sees exactly as many blocks as a roomy box does, so it cannot pass by agreeing with a capped probe.

### What this means for the product question

Truncation is the **safety net**, not the primary answer: the editor grows the node on commit. It bites a node that never went through that path — authored by an agent, or shrunk by hand afterwards (grow-only never re-grows a manual shrink). That is worth saying out loud in the follow-up rather than leaving "the paragraph is silently dropped" as the headline.

### Verification (re-run in full)

node projects **4347 passed** · apps/web jsdom **2314 + 77 passed** · browser projects **654 passed** (121 files) · widget smoke **PASS** · `pnpm -r typecheck` and `pnpm knip` clean.

kamiazya:
@coderabbitai の指摘は妥当だったので適用しました（b75a24f4）。

`y` はベースラインなのでディセンダはその下に落ちる — テスト名の主張（枠の下に描かない）より弱い判定でした。`getBBox()` の描画境界で測るように変更しています。コメントに限界を書いてはいましたが、書くより測るほうが正しい。

境界は実際に厳しくなりました: 修正を戻したときのミューテーション検証が、同じ枠下端 311 に対して **323 → 327** で赤になります。

canvas-viewer-browser 10 passed。

PR #879, opened 2026-08-18
refactor(canvas-render): make "content stays in its frame" one stated law

kamiazya:
Closing the *class* rather than the next instance of it. PR #878 fixed one seam; this makes the guarantee something every seam answers to.

## Measured first

Five node kinds × five box sizes, because "which seams leak" should be a table, not a reading exercise:

| case | overflow | what it is |
|---|---|---|
| `text@200x12` | **bottom +36** | the whole body, outside the frame |
| `file@200x12` | bottom +12 | the one-line label |
| `link@200x12` | bottom +12 | the one-line label |
| `group@*` | top +20 | the container label — outside **by design** |
| `file@200x120` | top +20 | the reference label — outside **by design** |

**The +36 was mine, from #878.** The carve-out that lets the editor's grow probe measure a natural height — *"a non-positive content box imposes no bound"* — cannot tell that probe apart from a node someone really made 12px tall. It was open for both.

## The probe's intent becomes a name, not a degenerate height

`naturalNodeContentSize(node, options)` is now canvas-render's answer to "how big must this box be". Fitting is an **internal** flag that only that function clears — deliberately *not* on `SpatialLayoutOptions`, so a normal render cannot turn the bound off by accident.

`apps/web`'s `requiredTextNodeHeight` becomes padding arithmetic on top of it instead of a layout at `height: 1`. Every tiny node is bounded again: **+36 → +12**, the same single kept element every other seam already had.

## The law, stated once against the output

> At most one element crosses the frame bottom, once the outside label is set aside, and nothing is painted above the frame except that one label.

Per node *kind*, not per seam — each seam that grows its own bound is exactly where the next overflow appears. Both escapes are **enumerated, not discovered**: the container label JSON Canvas puts outside the box, and the single element kept when nothing fits so a one-line-tall node is not an empty box.

Vertical only. Horizontal has its own declared exemptions (inline math is neither split nor cut, an atomic run is not split) and its own instrument in `text-wrapping-quality.test.ts`.

## The generator is the interesting part

Height is drawn as a fraction of each node's **own** natural content height. A flat `1..160` range reached the two-blocks-cross case about **once in 140 runs**, so the mutation that deletes the fit entirely passed on some seeds and failed on others — I caught that only because a truncated `head -4` made me re-run it. With the fraction it is **red 5 runs out of 5**. Density, not more runs.

Mutation-checked one rule at a time:

| mutation | reddens |
|---|---|
| delete the fit | the crossing law |
| a second above-frame producer | the label law |
| — | natural size pinned as independent of the stored box |

## Verification

node projects **4431 passed** · apps/web jsdom **2314 + 77** · browser projects **654** · widget smoke **PASS** · `pnpm -r typecheck` and `pnpm knip` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved whiteboard text sizing to accurately account for natural content dimensions and padding.
  * Prevented layout results from depending on a node’s previously stored height.
  * Improved rendering containment for text, file, link, and group nodes, reducing elements crossing frame boundaries.

### DIFF
--- a/apps/web/src/components/spatial-editor/scene-render.ts
+++ b/apps/web/src/components/spatial-editor/scene-render.ts
@@ -6,11 +6,7 @@
  */
 
 import type { MeasureText, ResolvedReference } from '@kamiazya/whiteboard-canvas-render'
-import {
-  layoutSpatialCanvas,
-  SPATIAL_THEME_GEOMETRY,
-  sceneBounds,
-} from '@kamiazya/whiteboard-canvas-render'
+import { naturalNodeContentSize, SPATIAL_THEME_GEOMETRY } from '@kamiazya/whiteboard-canvas-render'
 import type { SpatialCanvas, SpatialNode } from '@kamiazya/whiteboard-model'
 import type { ResolvedTheme } from '../../hooks/useThemeMode.js'
 import { createEditorAppearance } from './editor-appearance.js'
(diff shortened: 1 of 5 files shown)


kamiazya:
両方とも妥当で、片方は**このPRが追加した関数の実バグ**でした。修正しました（e90e9e5f）。

### Major: `naturalNodeContentSize` が箱に依存していた

指摘どおりです。`fitBodyInNode` の退化ボックス判定が fitting フラグより手前に残っていたため、`fitToBox: false` でも file ノードがラベルにフォールバックしていました。測定値:

```
tiny(h=10)    = { w: 16,  h: 16 }   ← ラベル
roomy(h=1000) = { w: 184, h: 40 }   ← 本文
```

「保存された箱に依存しない」ことが全てである関数が、箱に依存していた。あの判定は**レイアウトではなく fitting の一部**なので、そう書き直しました。

**捕まえるはずのプロパティが、2重に効いていませんでした。**

1. text ノードしか見ていなかった（関数は汎用なのに）→ 全種に拡大
2. 拡大してもまだ緑のまま。生成器の `resolveReference` がラベルだけを返しており、**file ノードが markdown シームに一度も入っていなかった**ため。箱に内容を入れる3シームのうち1つが、全プロパティから素通しでした。→ 本文を解決する resolver に変更し、ミューテーションは **5回中5回赤**

### Minor: 枠上ラベルの判定がカウントだけだった

これも指摘どおりで、text / link に新しい枠上プロデューサが生えても通ってしまいます。両者は **0件**を要求し、コンテナ形状の2種については単一要素が `textRun` であることを要求するようにしました。link ノードに枠上ラベルを生やすミューテーションで赤を確認しています。

検証: node 4432 / browser 654 / widget smoke PASS / typecheck・lint クリーン。

PR #884, opened 2026-08-18
feat(server-core): edit a canvas with a batch of ops in one transaction

kamiazya:
Second layer of the stack (base: `server-core-canvas-snapshot`, #881). This is the batch tool itself.

## The problem, stated precisely

Building a ten-node diagram took twenty-odd MCP round trips, each one a separate document load and save, with the model tracking ids across all of them — and inventing four integers of geometry per node, since `x`/`y`/`width`/`height` are all required by `sharedNodeFieldsSchema`.

The tools were not *too many*. They were the wrong *shape*.

(Worth noting what the reference point is **not**: Playwright MCP has no batch mechanism and 40+ imperative tools. What makes it work is that its snapshot is addressable and its action results carry fresh state — which is what #881 and this PR's `snapshot` field are for. Batching is this domain's own need, because a human clicks one button at a time and a diagram does not.)

## What lands

`wb_canvas_edit` takes an op list — `node.add` / `node.patch` / `node.remove` / `edge.add` / `edge.patch` / `edge.remove` / `node.lock` / `edge.lock` / `tidy` — and applies it as a single transaction: one load, one save, every op or none.

A refusal names the failing op by index **in the message** as well as on the error class, because only `.message` survives the MCP error path and a model repairing a batch needs to know which op it got wrong.

## Three things worth reviewing

**Geometry is optional, and the stored schema is untouched.** JSON Canvas 1.0 still requires integers; only the tool's *input* is relaxed. A node with no geometry is placed in a fixed grid below the existing content and the chosen position comes back under `geometry`. The placement is deliberately dumb so an agent can predict it — the `ponytail:` comment names free-region packing as the upgrade path if quality ever beats predictability.

**The result carries the resulting board.** It reuses `wb_canvas_snapshot`'s projection (extracted to `projectCanvasSnapshot` in this PR) rather than a parallel one, so the two can never disagree field-for-field.

**`node.remove` / `edge.remove` are new capability, not sugar.** There was no delete tool at all — the only way to remove a node was a whole-document replace through `wb_document_set`. Removing a node takes its edges with it, because a dangling edge stores a canvas that `spatialCanvasSchema` refuses on the next read.

Locks are accumulated in memory and written only once every op has applied, so a rejected batch leaves neither content nor locks behind. Registered under the per-document write lock: this tool reads the whole canvas, decides ids and placements against what it read, and writes it back.

## Verification

- `server-core-node` — 15 new tests, including all-or-nothing rollback, a lock set *inside* the batch binding later ops, unlock-then-patch in one batch, cascade edge removal, and placement wrapping.
- `pnpm test --project mcp-node` — 292 files pass.
- `pnpm smoke:e2e` — `tools/list → 23 tools`, plus two new assertions: that what the tool *reports* placing matches what its own snapshot says it *stored*, and that a rejected batch leaves nothing behind **through the real wire**.

## Reach

`blastRadius: none` — additive; no existing tool changes behaviour. `projectCanvasSnapshot` is an extraction from #881 with its behaviour pinned by that PR's tests.
`userReach` — registered and callable from any connected MCP client on merge.

Next in the stack retires the seven tools this replaces.

---

**Naming note.** This shipped as `wb_canvas_apply` in an earlier push and was renamed before review. `apply` is the verb of declarative reconcilers (kubectl, Terraform), where you hand over the desired state and the server computes the diff — a design that was explicitly considered and rejected, because it deletes whatever a human added that the agent's payload does not mention. A name that promises semantics the tool does not have is worse than a bland one. `wb_canvas_view` was also considered and rejected: `canvas_view` already exists as the MCP Apps widget tool, and two tools differing only by a `wb_` prefix is a selection trap.


kamiazya:
@coderabbitai review

PR #887, opened 2026-08-18
refactor(mcp-server)!: retire the seven spatial-mutation tools into wb_canvas_edit

kamiazya:
Top of the stack (base: `server-core-canvas-apply`, #884). This retires the seven single-purpose spatial-mutation tools that #884 replaced.

`wb_node_add`, `wb_node_patch`, `wb_edge_add`, `wb_edge_patch`, `wb_node_lock`, `wb_edge_lock`, `wb_canvas_tidy` — gone, no aliases. The registered surface goes **21 → 16 tools**, and the spatial-mutation surface goes **7 → 1**.

## The coverage moved before the code did

Of the 44 cases across the seven retired test files, **11 asserted behaviour `wb_canvas_edit`'s own tests did not yet reach**:

- a locked *node* not freezing the edges touching it
- an edge lock not leaking onto a node that happens to share its id
- tidy's scope restriction, fixpoint, and lock-avoidance
- the kind-undefined write declaring `spatial`
- schema-level refusals (negative width, invalid arrowhead end)
- `label` patched onto a text node being silently stripped

Those are ported and green **in the commit before the deletion**. That ordering is the point, not a formality: `annotate` was removed the same way once and nothing replaced it — `canvas-viewer`'s sticky-note control still sits unmounted, because every submission failed at the host with an unknown-tool error while the control looked live.

## The prose half is not a rename

`drawing-visuals` told agents there is *"no way to delete a single node or edge"* and to work around it by patching an unwanted node *"into something harmless"*. It also taught that *"every node needs an id, integer x/y/width/height"*. Both are now the opposite of the truth. A find-and-replace would have left every one of those claims standing.

**Three runtime messages were pointing callers at tools they can no longer reach** — `document-set`'s two content-loss refusals, `document-get`'s no-recorded-kind error, and `NodeLockedError`'s *"unlock it with `wb_node_lock`"*, shipped to an agent at the moment it is stuck. `NodeLockedError`, `EdgeLockedError` and `EdgeNotFoundError` turned out to have no consumers left and are deleted rather than corrected.

**`mcp-claude-cli-smoke` asked Claude to `call wb_node_patch with type=rectangle`** — a tool that never took a `type`, never created a node, and a node type this product does not have. Excalidraw-era text that outlived two migrations because the smoke only checks that a versionId is printed. It now asks for a `node.add` op that can actually succeed.

## Also here

- **`packages/mcp-server/skills/` is deleted.** A byte-identical duplicate of the repo-root `skills/`, tracked but read by nothing: both plugin manifests resolve `./skills` at the root, the npm `files` array deliberately excludes it, and #850 — the commit that decided skills ship through the plugin only — edited the duplicate instead of removing it. The guard only ever scanned the root, so a sweep that fixed one copy left the other stale with every assertion still green. `skills-tool-surface` now pins its absence.
- The seven names join `BANNED_PATTERNS`, beside `Excalidraw` and `/api/debug`.
- **ADR-0010** records the decision and the rejected alternatives (declarative desired-state, intent-level layout, keeping `wb_canvas_tidy` standalone, and the `apply` / `view` / `draw` names). **ADR-0009 is not rewritten** — it named these seven tools and a decision record is history.

## Verification

- `pnpm test --project mcp-node` — 292 files pass; `server-core-node` + `canvas-render-node` — 97 pass.
- `pnpm smoke:e2e` — `tools/list → 16 tools match expected set`, both smokes green. Every refusal the retired tools used to prove (duplicate id, dangling endpoint, locked node, locked edge, unknown edge id, markdown document) is re-asserted through `wb_canvas_edit`, plus rejected-batch rollback.


kamiazya:
@coderabbitai review

kamiazya:
Taking this over — the session that opened it is gone. Reconciled with current `main` and pushed as `70852a05`.

**Merging #881 + #884 broke `main`, and this branch carried the same fault.** All three imported `server-core/src/render/load-spatial-canvas.js`, which #886 had already deleted when it consolidated document loading into `tools/document-io.ts`. Fixed on `main` by #901; this branch is repointed the same way.

Worth naming why nothing caught it: both PRs reported `MERGEABLE`, both had 19/19 green, and `git merge-tree` found zero conflicts — because **neither PR touched the deleted file**, so there was nothing to conflict with. Their CI was green against their own base, which still had it. Every pre-merge signal answered a different question than "does the merge result compile".

## What the merge-result check then found

Typechecking and testing the *merge result* (rather than either side) surfaced two more defects that no branch could see on its own:

1. **`snapshot-not-found.contract.test.ts` imported `./node-add.js`** — a tool this PR retires. Its intent ("the read and write sides raise ONE class, not a parallel pair") is worth keeping exactly, so the write-side probe is now `wb_canvas_edit` instead of `wb_node_add`. What is under test is the shared loader, not the tool.
2. **`canvas-edit.test.ts` expected `DocumentNotFoundError`** where #886 renamed the store-level miss to `SnapshotNotFoundError`. #884 was written before that rename.

## Conflict resolutions, so a reviewer can check them rather than trust them

- `tool-profiles.ts`, `mcp-smoke-coverage.ts`, `testing.md`, `architecture.md`, `mcp-e2e-smoke.mjs` → **this branch's side**, which already carries `wb_canvas_edit` / `wb_canvas_snapshot` and drops the retired seven.
- `canvas-snapshot.ts`, `canvas-snapshot.test.ts` → **main's side** (#901's repointing).
- `create-server.ts`, `index.ts`, `errors.test.ts` → drop only the retired tools' registrations/exports/tests, keep main's `SnapshotNotFoundError` handling from #886. Verified field-by-field rather than accepting the `rerere` replay, since that replay was recorded by the dead session against an older `main`.
- `node-patch.test.ts` → deleted (its tool is retired). `EdgeNotFoundError` is confirmed to have no remaining references anywhere.

Verified on the merge result: `pnpm -r typecheck` clean, `server-core-node` + `mcp-node` + `canvas-render-node` + `arch-lint-node` **400 files / 4608 passed**. Two failures in an earlier full-parallel run (`cannot bind`, `IMPORT_MAX_CHUNK_BYTES`) both pass in isolation — the known load-dependent family, unrelated to this diff.

Note for the merge gate: CodeRabbit has never reviewed this PR. Its check reads `pass`, but the comment is "Review limit reached" — the same for every PR in this stack.

PR #889, opened 2026-08-18
feat(mcp-server): let a human see what the agent just did, and where

kamiazya:
Fourth layer (base: `retire-spatial-mutation-tools`, #887). The server half of making an agent's edits visible to the human watching the same board.

## The pieces were mostly already here

An agent editing a board a human is looking at was invisible. Almost everything needed to fix that existed and was never connected:

- **`viewport_request`** has been in the WebSocket protocol since the HTTP viewport route landed.
- **`operatorInfoSchema`** already distinguishes `kind: 'ai'` from `'human'`.
- **`routes/viewport.ts`'s own no-client hint** told callers to *"run `viewport_set`"* — a tool that did not exist. This PR makes that hint true.

## What lands

`ServerDeps` grows an optional `clientNotifier`. **Optional is the point:** every existing composition and every test stays a valid server without one, and a tool that needed a browser to be present would stop being headless.

- **`wb_canvas_edit` announces what it touched and follows it with a viewport fit.** Opt out with `follow: false` — moving someone's viewport is an interruption. A batch that moved no node moves no viewport either: `mode: 'fit'` with an empty list fits the *whole* board, a jarring jump for an edit that only removed an edge.
- **`wb_viewport_set`** exposes the message an agent could never reach. It answers `delivered: false` rather than failing when nobody is watching — a headless daemon is the normal case, and an agent that could not tell the difference would read every headless run as broken.

**Nothing notifies until the write is committed** — a human must never be shown an edit that was refused — and both calls are wrapped, because the batch is already on disk by the time anyone is told and a broken socket must not report a failure for an edit that succeeded.

## Two design points worth reviewing

**`agent_activity` is ONE message, not a begin/end pair.** A batch is atomic and lands in milliseconds, so a paired form would only flicker, and an `end` lost to a dropped socket would strand the indicator on forever. Presence becomes the client's job: hold "an agent is editing" for a few seconds after the last message and let it lapse. It is also never cached for replay (unlike `viewport_request`) — a tab that connects later already has the edit, so replaying the announcement would highlight a change it has always had.

This is also why `wb_node_lock` could not have been reused for presence: a lock is durable sidecar state, so an agent that crashes leaves one behind for a human to clear.

**`viewport` joins ADR-0009's entity list**, and the guard that caught it now carries the reason. A viewport belongs to a client rather than to stored content, so it is not a document-model noun like the rest — but `wb_viewport_set` is still an agent asking the *daemon* to do something, which is what the `wb_` plane is. ADR-0009 point 7's exemption is narrower than it looks: it covers tools the MCP Apps **host** renders (`canvas_view`, `canvas_open`), not everything that touches a UI.

## Verification

- `server-core-node` + `mcp-node` — 324 files pass. 6 new tests on `wb_canvas_edit`'s notification behaviour (including: says nothing when the batch was rejected; a throwing notifier does not fail a batch already on disk), 4 on `wb_viewport_set`, 6 on the bridge.
- **Mutation-checked** the bridge's load-bearing claim: breaking the documentId→path resolution turned 2 tests red, and the restore was verified before re-running green.
- `pnpm smoke:e2e` — `tools/list → 17 tools`, and `wb_viewport_set → delivered:false with no browser attached` asserts the headless path is a success rather than an error.

Next and last in the stack is the apps/web half: the presence indicator and the highlight that consume `agent_activity`.


kamiazya:
@coderabbitai review

PR #892, opened 2026-08-18
feat(web): show who moved the board — agent presence and a fading outline

kamiazya:
Top of the stack (base: `mcp-agent-activity-notifier`, #889). The browser half of making an agent's edits visible: the daemon announces what it touched, this is what a human sees.

## What lands

- **A chip at the top of the canvas** — "Agent editing — added 3, tidied the layout". `role="status"` + `aria-live="polite"`, because someone editing under you is worth announcing and one message per applied batch is not enough traffic to talk over anything. `assertive` would interrupt, which an edit somebody else made does not warrant.
- **The nodes it touched, outlined** for three seconds.

## Presence is held here and allowed to lapse

Nothing ever says "the agent is done" — the server sends one message per batch and no more. So both timers restart on each announcement and expire on their own. **That is what makes a crashed agent disappear** instead of leaving an indicator a human has to dismiss, and it is why a durable lock could not have carried this.

The two timers are deliberately different lengths, and **only the outline clears at the shorter one**: dropping the summary with it would blank the label of a chip still on screen, which reads as a bug rather than as a fading highlight.

## Two smaller choices

**The outline reuses `MemberOutlinesOverlay`** rather than adding a second overlay component — "these boxes, outlined" is the same drawing, and only the colour and the meaning differ. It gains two optional props (`stroke`, `testId`) and nothing else. It passes **no edges**: an agent reports the edges it touched, but an edge outline is far less legible than a box one and the nodes are what actually moved. Noted in a comment as the thing to add if edge-only batches turn out to be real.

**`onAgentActivity` is optional on `DocumentBackendHandlers`** because it is news rather than protocol — a backend that never emits it, and a page that does not care, are both valid. Only the daemon path wires it; a browser-local document has no agent to announce.

## Verification

- `apps/web` jsdom — 208 files pass. 7 tests on the hook (lapse, burst-restart, most-recent-only, unmount cleanup), 3 on the page wiring.
- **Mutation-checked** the page wiring: unsubscribing the page's `onAgentActivity` turns 2 of the 3 page tests red, with the restore verified before re-running green. That join is exactly what the hook's own tests cannot see.
- `web-browser` — 119 files pass (this PR touches `SpatialEditor`).
- `mcp-node` + `server-core-node` — 324 files pass. `pnpm -r typecheck`, `pnpm lint`, `pnpm knip` clean.

## Note for the reviewer

`gh stack submit` failed three times on `test-node` before landing. The failing test is `0011-import-fs-blobs > chunks a blob larger than IMPORT_MAX_CHUNK_BYTES...`, timing out at **10455ms against a 10000ms budget** — unrelated to this PR, an overrun rather than an assertion, and only under the extra load of the hook running against a 5-ref atomic push. It passes standalone and in every single-branch push. Flagging rather than fixing: a 4%-over per-test budget in a migration test is its own lane.


kamiazya:
@coderabbitai review

PR #895, opened 2026-08-18
refactor(mcp-server)!: merge wb_scene_digest into wb_canvas_snapshot

kamiazya:
Top of the stack (base: `web-agent-presence`, #892). Answers a question raised on the stack: **can the two reads be merged?**

Yes — and my earlier reasoning for keeping them apart was wrong in an instructive way.

## The wrong test

I argued "neither subsumes the other's information". That is true, and it is **not the test**. The seven mutation tools this stack retired each had a defensible payload too, and were still the wrong *shape*. Two reads that both answer with a node list put the same selection problem one layer below the one this stack just fixed.

## What lands

`wb_canvas_snapshot` now answers both read questions:

- **default** — what is ON the board: node types, text, stored geometry, lock state, every edge.
- **`layout: true`** — additionally whether the board is TIDY: overlap / containment / cluster / free-region analysis of the laid-out scene.

`wb_scene_digest` is retired. Registered surface: **17 → 16**.

## The part that is a fix, not tidying

`sceneDigest` **degraded silently** past its guards. A board over `PAIRWISE_MAX_ENTRIES` reported `overlaps: []` — which reads as *"nothing overlaps"*, the most misleading possible answer to "is this board tidy" — and a scene too large for the free-region grid reported no free space. It now reports `truncated`, joining the snapshot's existing rule that **a capped read must never look complete**.

## Two design points

**The layout half is opt-in** because it costs a full layout pass (text measurement, edge routing), while "what is on this board" is the far more common question.

**The merged `nodes` carries STORED geometry.** The two halves genuinely disagree about what a box is: the analysis reports where content is actually *drawn* after text sizing; `x`/`y`/`width`/`height` are what an edit *writes back*. The node list carries the one you act on; the analysis keeps its own.

## Coverage moved before the code, again

The digest's pinned literal, its id-naming guard, its index-row markdown refusal and its no-recorded-kind case are all `wb_canvas_snapshot` tests now. The reference-stability characterization in `canvas-render-svg.references.test.ts` — that a canvas's analysis must not move when a *different* document is edited — was ported rather than deleted.

## Two stale claims the sweep exposed

- **`auditing-workspaces` taught a falsehood.** It said a digest *"reads the spatial containers without checking the document's kind, so a markdown document digests as `{ nodes: [], edges: [] }` with no error"* — untrue since the digest gained `assertSpatialDocument`, and the skill's entire "do not probe with it first" section rested on it. Rewritten around what is actually true.
- **The convergence smoke polled the digest purely as a node-id list**, which the default read does without paying for a layout pass.

## Verification

- `mcp-node` + `server-core-node` + `canvas-render-node` — 390 files pass. `pnpm -r typecheck`, `pnpm lint`, `pnpm knip` clean.
- `pnpm smoke:e2e` — `tools/list → 16 tools`, both smokes green. The new `wb_canvas_snapshot(layout:true)` call is the runtime guard that a laid-out scene still validates through the real MCP SDK — the job `wb_scene_digest`'s smoke step used to do.


kamiazya:
@coderabbitai review

PR #896, opened 2026-08-18
fix(server-core): measure text once, so wb_scene_render draws what the exporter draws

kamiazya:
`wb_scene_render` and `wb_scene_digest` laid every scene out with a constant-ratio
text estimate while the PNG/SVG exporter — same process, same canvas — used
opentype.js glyph metrics. The two disagreed on where every line wraps, so the SVG
an agent read back was not the picture a user exported.

The estimate itself was fine; there was simply no way to pass the real measurer in.
Chasing that down first turned up a smaller problem underneath: three composition
roots had each grown their own copy of that estimate, with three different constant
sets.

## Visual repro

Same canvas, same pipeline, only the measurer differs. Before, the second node's
text is painted straight through its own right edge (layout believed it fit) and
the heading wraps for no reason:

After — identical to what the exporter produces for this canvas:

## What changed

**1. One constant-ratio measurer, not three** (`598c249c`)

| | advance | ascent | descent | lineGap |
|---|---|---|---|---|
| `server-core/render/fallback-measure.ts` | 0.55 | 0.8 | 0.2 | 0.1x |
| `mcp-server/server/export/measure-text.ts` | 0.55 | **0.75** | **0.25** | 0.1x |
| `canvas-viewer/measure-text.ts` | **0.6** | 0.8 | 0.2 | **0** |

Same role, same formula, silently divergent numbers — so the same canvas measured
differently depending on which degraded path produced it. `constantRatioMeasureText`
now lives in canvas-render, the package that defines the `MeasureText` seam, and
the other two copies are deleted. The promoted version is mcp-server's, the only
one that clamped a non-finite `sizePx`.

The real measurers are untouched: opentype.js glyph metrics and Canvas `measureText`
are different mechanisms, not duplicate implementations of one.

**2. A way to inject the real one** (`43d67ed2`)

`ServerDeps.measure?: () => Promise<MeasureText>`. Async and a factory because
parsing a font file is, and memoized inside the measurer, so a server that never
renders pays nothing at startup. Absent, the tools degrade exactly as before —
server-core still cannot load a font itself.

**3. Regression coverage** (`4e1309cb`)

`render-parity.test.ts` runs one canvas through both producers and asserts the text
runs land at identical wrap points. The fixture alternates `MMMM` and `iiii` on
purpose: a constant-ratio measurer cannot tell them apart, so average-width text
would let the two agree by accident. Mutation-checked — drop `measure` from
`resolveServerDeps` and it goes red.

**4. Unrelated flake fix** (`5f71fe08`)

`edge-crossing-min.test.ts` failed three consecutive full-suite runs while passing
every time its project ran alone. `Test timed out in 5000ms` on an in-body
`await import('./spatial-canvas.js')` — the third flake shape in
`.claude/rules/integrator-flow.md`, verbatim. The file mocks nothing, so the import
had no reason to be dynamic. Hoisted.

## Worth knowing before merging

`wb_scene_digest` values move once, for every document, since the bboxes are now
measured with the real font. Nothing pinned them, and no test or consumer in the
repo broke — but anything outside using a digest as a change signal will see one
spurious change.

## Verification

- `pnpm test` — 8236 passed, 0 failed
- `pnpm -r typecheck`, `pnpm build`, `pnpm knip`, `pnpm lint` clean
- Manual: rendered the same canvas through `wb_scene_render` and the headless
  exporter; text runs identical (`runs match: true`). With the injection removed,
  the tool wraps into 4 lines where the exporter wraps into 3.


kamiazya:
#897 just merged and conflicts with this branch directly — it modifies `packages/server-core/src/render/fallback-measure.ts`, which this branch deletes. Flagging the requirement rather than the file, because resolving the conflict the obvious way (keep the deletion) silently drops the fix.

**Both PRs are right, and landing them naively gives the worst of both.** Measured through the real stdio MCP server, one node `{width: 200, height: 32, text: 'これは日本語のテキストで折り返します'}` → `wb_scene_digest`:

| | per-kana advance @16px | digest verdict |
|---|---|---|
| main before #897 (0.55 uniform) | 8.8 | `truncated` absent |
| **#897** (script-aware estimator) | **16.0** (correct) | `truncated: true` |
| **this branch** (real opentype measurer injected) | **7.1** | `truncated` absent |

The 7.1 is not a measurement — it is Roboto's `.notdef` advance. The vendored face carries no CJK glyphs, and `measure-text.ts` has no per-code-point guard: `constantRatioMeasureText` only fires when the whole font fails to parse (L116/L132), never for a missing glyph, so L56's `getAdvanceWidth` returns `.notdef` for every kana, ideograph and Hangul syllable:

```
"abcdefghij"                          -> 74.6  (full-em 160)
"これは日本語のテキストで折り返します" -> 127.7 (full-em 288)
"あ"  ->  7.1 (16)    "漢字" -> 14.2 (32)    "한국어" -> 21.3 (48)
```

Verified end-to-end on this branch: the same node comes back `{"id":"jp","bbox":{...},"z":0}` — `truncated` absent, same as unfixed main. So this branch achieves parity between digest and export by making both wrong for CJK at ~44% of true width, which is worse than what main does today.

**The reconciliation makes both goals true, and is small:**

1. Fold #897's script-awareness into the promoted `constantRatioMeasureText` — it already imports from canvas-render, where `isFullWidthCodePoint` now lives (`measure.ts`). Charge a full em for fullwidth code points, keep the ratio for the rest.
2. Give `createOpentypeMeasureText` a **per-code-point** fallback: `font.charToGlyphIndex(char) === 0` → measure that character with the estimator instead of `.notdef`. Detect by glyph index, not by comparing advances.

With both, parity holds *and* CJK is measured correctly — and it fixes the export-path defect too, which is otherwise its own backlog item (a Japanese PNG export is laid out against `.notdef` advances today; if the host supplies a CJK face to resvg, the glyphs paint at full width against quarter-width boxes and run visibly past their own node).

`render-parity.test.ts` is the right place to pin it — add a Japanese fixture alongside the `MMMM`/`iiii` one. A constant-ratio measurer and a `.notdef`-returning one agree with each other on CJK, so a Latin-only parity fixture passes while both are wrong.

kamiazya:
Reconciled and pushed to this branch as `5378caf2` (merge of `origin/main`, which now carries #897). Both halves kept; each turned out to be what the other needed.

**`constantRatioMeasureText` is now script-aware.** A fullwidth code point costs a full em. It is still an estimate matching no real font — that property is unchanged — but one ratio for every character is a wrong *model* rather than a coarse one, and no choice of constant fixes mixed text. Iterating code points instead of UTF-16 units fell out of the same loop and stopped an astral code point being charged twice.

**`createOpentypeMeasureText` falls back per code point** for anything the face does not carry, detected with `charToGlyphIndex(char) === 0` rather than by comparing advances (a legitimately narrow glyph can share `.notdef`'s width). A string the face carries entirely still takes exactly one `getAdvanceWidth` call, so Latin measures byte-identically to before this commit.

End-to-end through the real stdio MCP server, the node from my previous comment now answers `{"id":"jp",...,"truncated":true}`. Parity holds *and* CJK is right — and exported Japanese is fixed as a side effect, so `issues/export-measurer-notdef-cjk` is resolved rather than deferred.

**Two things worth flagging in review:**

`fallback-measure.truncation.test.ts` from #897 survives as `degraded-measure-truncation.test.ts`, repointed at `constantRatioMeasureText`. It pins the digest verdict for Japanese, which is the behaviour #897 existed to fix, so deleting it with the file it was named after would have dropped the guard silently.

`render-parity.test.ts` gains a Japanese node — **and my first version of that assertion was worthless.** A constant-ratio measurer and a `.notdef`-returning one *agree with each other* on CJK, so parity is satisfied while both are wrong; the fixture has to separate the two models, not just exercise them. The 26-character string I reached for first happened to wrap at both widths and passed against the unfixed code. The mutation check is what caught it. The current fixture is 60 fullwidth characters in a ~184px content box — 6 lines at a full em, 3 at `.notdef` — and reports `expected 3 to be greater than or equal to 4` when the glyph guard is reverted.

Verified: `mcp-node` + `server-core-node` 3840 passed / 328 files, `canvas-render-node` + `canvas-viewer-node` + `canvas-viewer-jsdom` + `arch-lint-node` 896 passed, `pnpm -r typecheck` clean, `pnpm lint` exit 0 (3 warnings, all pre-existing in `apps/web`), `pnpm knip` exit 0, `pnpm smoke:e2e` ALL OK.

kamiazya:
@coderabbitai review

PR #899, opened 2026-08-18
feat(server-core): reconcile a group's contents with a region.set op

kamiazya:
Top of the stack (base: `merge-canvas-reads`, #895). The last piece of ADR-0010, and the one it deferred.

## The deferred question answered itself

`region.set` — "this group should look like this" — is the only op that deletes something it was **not** told about. It was deferred because of one question: *what if a human is mid-drag across the group's edge?*

The scope rule answers it. A node is in the region only when its stored box is **strictly enclosed** by the group's. A node straddling the edge is not enclosed, so it is out of scope and survives untouched — which is precisely the mid-drag case, handled *by the rule* rather than special-cased around it. Edges follow the same rule (in scope only when both endpoints are), so an edge leaving the region is never collateral damage.

**Mutation-checked:** relaxing strict containment to mere overlap turns the straddling-node test red, and nothing else.

## Two more rules, to make it usable rather than merely safe

- **A declared node that already exists is MERGED, not replaced.** Omitting geometry leaves it where it is, which is what makes re-applying the same region a no-op. The cost is that this op cannot *clear* a field — `node.patch` can, and the schema comment says so.
- **A locked element anywhere in the region refuses the whole batch, up front** — before anything is removed. Deleting by omission and silently dropping a locked node would be the worst possible reading of this op.

## One wrinkle worth reviewing

Coordinate-less declared nodes are placed **inside the group** by their own cursor, not the board-level one. The board cursor places below all existing content, which for a region op would land the node *outside* the region it was declared in — and therefore out of scope on the next call, breaking idempotence. That is a small second placement routine rather than a parameter on the first, because the two anchor differently (a box vs. the content bbox).

`nodes` requires an `id`. Reconciliation is matching by id; a declared node without one could only ever be a create, which is `node.add`'s job.

## Verification

- `server-core-node` — 40 tests on `wb_canvas_edit` now, 8 of them on this op: replace-inside/leave-outside, the straddling case, never deleting its own group, edge scoping, the locked refusal, in-group placement bounds, idempotence, and a non-group `within`.
- `mcp-node` + `server-core-node` — 323 files pass. `pnpm -r typecheck`, `pnpm lint`, `pnpm knip` clean.
- `pnpm smoke:e2e` — exit 0. The new step drives the full reconcile through the real wire (declare two, then declare one, assert the other is gone) **and** asserts the rest of the board is untouched, which is the property that makes deleting-by-omission safe to hand an agent at all.

## Summary by CodeRabbit

* **New Features**
  * Added `region.set` for reconciling a region’s nodes and edges to an explicit list.
  * Omitted content within the region is removed, while content outside or crossing its boundary is preserved.
  * Existing nodes are merged, and new nodes without positions are placed within the region.
  * Locked regions or content prevent the entire operation from being applied.
* **Documentation**
  * Updated canvas editing guidance and architecture documentation with `region.set` behavior and usage.


kamiazya:
@coderabbitai review

kamiazya:
Taking this over — the session that opened it is gone. Merged current `main` and fixed the review findings; pushed as `f49cc8e2`.

**Both Major findings are real, and I reproduced each with a failing test before touching the code.**

**1. `touchedEdges` as region-deletion state.** Confirmed. `touchedEdges` accumulates across the whole batch, so:

```js
ops: [
  { op: 'edge.add', edge: { id: 'outside', fromNode: 'far1', toNode: 'far2' } },
  { op: 'region.set', within: 'g', nodes: [], edges: [] },
]
```

deleted `outside` — undeclared, and touched by the add — with both endpoints 900 units from `g`. Silent data loss. Region removals now go in a set scoped to the op, exactly as suggested; `touchedEdges` keeps its batch-wide reporting job.

**2. Declarations outside the region.** Confirmed, and the lock angle is the sharp end: the preflight only walks what is IN scope, so an out-of-region target was never lock-checked. A `region.set` could therefore modify a **locked** node without unlocking it. A declared node must now land strictly inside `within`, and a declared edge must have BOTH endpoints inside it.

**3. Strict containment comparisons — not applied, and applying it would be a regression.** The finding says the contract requires strict enclosure. The code's own behaviour says otherwise: `placeWithin`'s fallback places a node at exactly `{ x: bounds.x, y: bounds.y }`. Under `>` / `<`, a node **this very op just placed** would be out of scope on the next call, so re-applying the same region would stop being a no-op — breaking the idempotence the op documents and `is idempotent: applying the same region twice changes nothing the second time` pins. Inclusive is correct here; a node flush against the group's inner edge is inside it.

**4. `SKILL.md`.** Applied, and extended: it now states all four rules — full enclosure, both-endpoints-for-an-edge, declarations must land inside, and a locked in-scope element refusing the batch before anything is written.

Each of the three guards is **mutation-checked separately** — reverting any one reddens exactly the test that pins it and leaves the other two green:

```
M1 (scoped deletion reverted) → × leaves an edge an earlier op touched alone…
M2 (node guard reverted)      → × refuses to move a node that is not inside the region
M3 (edge guard reverted)      → × refuses an edge whose endpoints are not both inside the region
```

**One more thing the merge hid.** Merging `main` produced **zero conflicts** and six failing tests: main had repointed `canvas-edit.test.ts`'s import to `loadDocument`, while this branch's added region tests still called `loadSpatialCanvas`. `ReferenceError: loadSpatialCanvas is not defined`. Nothing conflicted because the two sides edited different parts of the file. Repointed.

Verified on the merge result: `pnpm -r typecheck` clean, **401 files / 4638 passed**, `pnpm knip` clean, `pnpm smoke:e2e` ALL OK.

PR #902, opened 2026-08-18
feat(web): a document browser — columns, real renders, move, create in place

kamiazya:
The files tab showed a tree and a preview: a folder was something you could only expand in place, no surface answered *"what is in this folder"*, the preview poured OKF source into a `<pre>`, and the hierarchy it displayed could not be added to or rearranged from the UI at all.

It is now a browser.

| pane | answers |
|---|---|
| sidebar | **where** — folders, at every depth |
| contents | **what is here** — one level, as cards |
| preview | **what it is** — the document drawn, its path, kind and age, and what you can do to it |

**Two column modes, not a breakpoint.** One column is the whole tree, folders and documents together, reachable without navigating anywhere. Two columns splits it. Neither is a subset of the other, and selecting in either fills the same preview, so they are two ways into one browser.

## The model

`folderContents` is the whole of it. A folder has no record of its own, so it is a **shared path prefix**: its identity is the prefix, its name is the prefix's last segment, its count is everything below it. The prefix is anchored at a segment boundary, which is what keeps `design-system` out of `design`. A document can be a folder at the same time — the sidebar shows `design` as a folder, and the pane listing its *parent* shows it as a document. Neither picks one, because both are true.

## Two renditions, and why that is not the inconsistency it looks like

Cards and the preview draw the **real render** — the same SVG at two sizes, so a thumbnail is always a small version of what the preview enlarges. One-column rows draw a **minimap** instead, because a faithful render carries no information at row size. Measured on one document at four sizes and four aspects:

A square 24px is a single horizontal dash; 40×24 and 56×32 are smears; only around **80×44** do the heading, the bullet block and the second heading separate. A card is 144×64 and clears it. A row will not. The minimap is a reduction of the same layout the card renders, not a second opinion about it.

## Acting on a document

- **Move.** #888 made the route subtree-safe and its own comment recorded that nothing called it — *"no MCP tool, CLI command, or apps/web affordance calls this route yet"*. This is the caller. A move re-reads the list rather than patching it, and the selection follows the document, because losing it blanks the preview exactly when someone wants to see the move landed. The server's refusal is forwarded verbatim: a subtree move collides on a **produced** path, so the path typed is often free, and the integration test asserts on a path the caller never typed.
- **Create in place.** Creation always landed at the workspace root; the browser showed a hierarchy it could not add to. Numbering is per folder — `design/untitled` and `untitled` do not collide, so letting one push the other to `-2` would make the counter jump for no visible reason. No name is collected (ADR-0006), and no path is derived from one (ADR-0008).
- **Duplicate and delete**, so retiring the grid loses nothing. Both stay owned by the page, which already has the duplicate and the confirmation dialog. The pane asks; it never deletes.

## Also measured: what a `[[reference]]` survives

Nothing rewrites a reference on a move or a rename — there is no backlink index anywhere. The three forms fail in complementary ways, now pinned in `reference-survival.test.ts`:

| written as | path move | display-name change |
|---|---|---|
| the path | breaks | survives |
| the name | survives | breaks |
| the id | survives | survives |

This is the price of a move affordance and nothing stated it. Dropping the path form would not be a fix — it trades that for the more frequent breakage in the other column.

## Evidence

Every rule is mutation-checked: reverted, confirmed RED, restored, with occurrence counts printed on both sides. That is ~35 rules across `folderContents`, both trees, the breadcrumb, the contents cards, `fitSvgToBox`, both loaders, the thumbnail, the preview, the worker's `markdown-render`, the move, and the create.

Six checks were **vacuous when first written** and are fixed here rather than trusted — each found by its own mutation staying green:

- two `waitFor`s that succeeded on the first render, while the kind icon was still up, and never observed the state under test;
- one that could not tell "drawing" from "cannot draw";
- one reading the kind off the thumbnail placeholder instead of the subtitle;
- one asserting on a card's whole text, where the word appears twice;
- one nested-`<svg>` fixture with no viewBox, so the root anchor it pinned was never reached.

Two guards were **deleted rather than tested**, both unreachable: `rest === ''` in `folderContents` (the boundary filter drops that row first) and the direct-children filter in `newDocumentPathIn` (every descendant's remainder contains a slash; no name it asks about does).

And the size measurement above was **wrong the first time** and said every size failed: it waited for `svg[viewBox]`, and the lucide kind icon it was waiting to replace is itself an `<svg>` with a viewBox.

The spatial render has a browser test of its own — snapshot bytes → LoroDoc → the pool's `layout` → SVG — because the capture that appeared to show it working had mocked the snapshot route to 404 and was showing the markdown half twice. It asserts node text, since an empty scene also serialises to a valid `<svg>`.

`apps/web` as CI runs it: **2523 passed** (jsdom) + **77 passed** (node). `pnpm lint`, `pnpm -r typecheck`, `pnpm knip`, `arch-lint-node` all clean.

## Deliberately not here

**The grid/tree toggle stays.** Retiring the grid is not the mechanical deletion it looks like: `DocumentListView` carries search (`filterDocumentsBySearch`) and pinned-first ordering, and the browser has neither. `DaemonIndexPage.test.tsx` is 38 tests, most of them driving the grid's cards. That belongs in its own PR whose whole diff is "delete the grid, move its tests".


kamiazya:
Applied. The finding was correct, and it is the exact invariant this PR's description claimed ("each pane is driven by exactly one to its left, so the three cannot disagree") and did not have — a preview could outlive the folder it came from, with no row left in the contents pane to mark it.

Fixed in afa5ef27 with the red test first (`drops the preview when navigating to another folder`), then all three navigation routes — sidebar, breadcrumb, folder row — through one `selectFolder` handler. Mutation-checked both halves: removing the `setPreview({kind:'idle'})` and leaving the breadcrumb on the raw setter each go red.

One deviation from the proposed fix: I considered the more precise rule (keep the preview if the document is still listed in the new folder) and dropped it as unreachable. A document can only be opened from the folder it lives in, so `parentFolder(preview.path)` always equals the folder you are leaving — the keep branch could never run.

PR #903, opened 2026-08-19
ci: let every required check run on merge_group

kamiazya:
Groundwork for enabling a **merge queue** on `main`. Inert until the queue is turned on in branch protection — `merge_group` never fires before that.

## Why this, and not "require branches to be up to date"

That setting would have caught the incident below, and it is the wrong tool here: with several dev-loops in flight it makes every merge invalidate every other open PR, forcing each to update and re-run CI. That serialises exactly the parallel development this repo's flow is built on, so `strict` stays `false`.

A merge queue answers the same question without that cost: it tests the **speculative merge** — main, plus anything queued ahead, plus this PR — at merge time, batched. Branches never need updating by hand; only the final verification is ordered. `gh stack merge` composes with it (a stack is added to the queue).

## Why it is needed

`on: pull_request` already builds a merge commit, so CI does test a branch against main — but only as main stood **when the PR last moved**, and nothing re-runs it when main advances.

| | |
|---|---|
| #881 head pushed | `2026-08-18T11:37:31Z` |
| #886 merged | `2026-08-18T13:06:55Z` |

#881 stayed green that whole time against a base that no longer existed. Merging it landed a `main` that did not compile — three files importing `server-core/src/render/load-spatial-canvas.js`, which #886 had deleted (fixed by #901). `MERGEABLE`, a zero-conflict `git merge-tree`, and 19/19 green all said it was safe, and none of them was answering the question: **a branch that merely imports what the trunk deleted has nothing to conflict with.**

The same shape appeared four more times while reconciling that stack — a contract test importing a retired tool, a stale error-class expectation, and two more stale import paths — plus a clean auto-merge that duplicated a tool-registration block (twelve MCP tests failing as HTTP 500s that never named the cause) and another that silently dropped a per-node signal. Conflict markers are not a proxy for correctness when a branch predates a rename on the trunk.

## The prerequisite this PR supplies

Every required check must also run on `merge_group`, or the queue waits forever for one that cannot fire. Required today: `verify`, `pr-title`, `check`, `test-unit (1)`, `test-unit (2)`, `test-jsdom`, `test-browser`, `widget-smoke`.

- `ci.yml` owns all of them but one — a single trigger covers it.
- **`pr-title` is the trap.** It lives in its own workflow deliberately (so it can re-run on `edited`, which `ci.yml` must not), so it needs the trigger separately — and a merge group has no PR title, so the validation step short-circuits rather than pretending to check something. Adding a required check in future means adding the trigger with it.

## To enable

Settings → Branches → `main` → **Require merge queue**. Squash is already the only permitted method, which is what the queue should use. Worth knowing before flipping it: the queue runs the full required set per batch at merge time, and this repo's CI is not cheap — batching is what keeps that affordable, so leave the batch size above 1.

## Summary by CodeRabbit

* **Improvements**
  * Added validation for merge-queue commits to help ensure changes remain safe to merge.
  * Updated automated checks to support merge-queue workflows.
  * Added guidance for handling stale checks and performing manual pre-merge validation.

### DIFF
--- a/.claude/rules/dev-flow.md
+++ b/.claude/rules/dev-flow.md
@@ -110,4 +110,40 @@ Native **Task list** = live board (in-flight / blocked / done; main session owns
 
 The integrator's push is guarded on two sides. **Local (pre-push, `lefthook`)**: `pnpm -r typecheck` + `pnpm test --project mcp-node` + `pnpm lint:noconsole` run before the commit leaves the machine (pre-commit only formats staged files, to not slow the dev-loop's worktree commits). **Cloud (post-push)**: GitHub Actions CI (`verify`) + CodeRabbit + AccessLint + WIP + CodeQL — monitor with the `Monitor` tool and triage with the `ci-triage` workflow/skill into Tasks / whiteboard canvases.
 
+**A green PR check does not say the MERGE is green.** `on: pull_request` builds
+a merge commit, so CI already tests a branch against main — but only as main
+stood when the PR last moved, and nothing re-runs it when main advances.
+Measured: #881's head was pushed at 11:37Z, #886 merged at 13:07Z, and #881
+stayed green the whole time against a base that no longer existed. Merging it
+broke `main` outright — three files importing a module #886 had deleted — and
+`MERGEABLE`, a zero-conflict `git merge-tree`, and 19/19 green all called it
+safe. None of them was answering the question: a branch that merely IMPORTS
+what the trunk deleted has nothing to conflict with. The same shape appeared
+four times in one stack, twice more as a clean auto-merge that duplicated a
+registration block and silently dropped a signal.
+
+"Require branches to be up to date" is the obvious fix and is deliberately NOT
+used (`strict: false`, user decision 2026-08-19): with several dev-loops in
+flight it makes every merge invalidate every other PR, which is the parallelism
+this whole flow is built on. The answer is a **merge queue** — it tests the
+speculative merge (main + anything queued ahead + this PR) at merge time,
+batched, so development stays parallel and only the final verification is
+ordered. `gh stack merge` composes with it: a stack is added to the queue.
+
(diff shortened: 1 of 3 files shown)


kamiazya:
Closing: the premise does not hold. **Merge queue is not available on this repository**, so these `merge_group` triggers would be config for a feature that can never fire.

Measured rather than assumed:

| probe | result |
|---|---|
| classic branch protection API | exposes no merge-queue field at all |
| ruleset with `required_status_checks` only | **created successfully** |
| same ruleset **plus** `merge_queue` | `Invalid rule 'merge_queue': ` — reason empty |
| `merge_queue` targeting a non-protected branch pattern | same failure, so not a clash with classic protection |
| `merge_queue` with no parameters | `Invalid merge method 'merge'. Not allowed for this repository.` |

That last one proves the rule type is recognised and that the payload shape was right — lowercase `squash` fails schema validation while `SQUASH` passes it, and the request is then rejected semantically with no reason given. With the `required_status_checks`-only control succeeding, the rulesets API and the `admin: true` permission are both fine; `merge_queue` alone is refused. The repository is public but **user-owned**, which is the remaining difference.

The probe ruleset was deleted; the repo has zero rulesets again, exactly as before.

**What we will do instead** (user decision, 2026-08-19): keep `strict: false`, and run `gh pr update-branch` on the single PR being merged, immediately before merging it. Same effect as "require branches to be up to date" for the branch that matters, without invalidating every other open PR — which is the cost that made that setting unacceptable with several dev-loops in flight.

Worth recording alongside it: `gh run rerun` cannot substitute. It replays the ORIGINAL event payload, so it never sees a newer base — `pr-title.yml`'s own header comment already says exactly this for a different reason. The branch has to move.

And there is a backstop already in place that nobody was watching: `ci.yml` runs on `push: [main]`, so main going red is visible immediately after a bad merge. Checking it after each merge would have caught the incident that started this.

PR #911, opened 2026-08-19
feat(mcp-server): say which characters the export fonts cannot draw

kamiazya:
A PNG export draws with the vendored Roboto and nothing else (`loadSystemFonts: false`), and resvg does not substitute a face it was not given. So a Japanese canvas exports as tofu boxes — **silently**.

## What it looks like

Same canvas, same SVG, the only difference being which files reach `fontFiles`:

**This PR does not ship the second picture** — which font to distribute is a separate decision. It makes the first one stop being silent.

## Why silence is the actual defect

Measurement is already correct: since the per-code-point fallback landed, the box is the right size, the text wraps in the right places, and `truncated`, `overflows` and the digest all report a perfectly healthy render. The layout is right. The only thing wrong is that the reader cannot read a word of it, and nothing anywhere says so.

`undrawableCharacters` answers it by reusing the question the measurer already asks internally — `charToGlyphIndex(char) === 0` — instead of leaving it an implementation detail that is computed, used for estimation, and thrown away. Both export results carry it, and a non-empty answer is logged: the same discipline as the existing `Roboto TTF not found` warning, one level finer. That one is per-install and logged once; this is per-canvas and logged per render.

```
こんにちは世界 -> undrawable: ["こ","ん","に","ち","は","世","界"]
Hello world    -> undrawable: []
```

```json
{"level":"warning","scope":"headless-renderer","count":7,"characters":"こんにちは世界",
 "msg":"export fonts have no glyph for some characters; they render as tofu"}
```

Deliberately about the **renderer's fonts**, not about the text: an SVG export keeps the characters as `<text>`, so a viewer whose system carries the face reads them normally, and only rasterisation loses them. Both result types document which they mean.

## Two things this turned up on the way

**`headless-renderer.test.ts` mocks `measure-text.js` wholesale in three places**, so a new export from that module is a runtime failure rather than a type error. They now answer `loadExportFont` too.

**`document-backend-contract.test.ts` reported `tsc failed:` with an empty body.** It interpolated `result.stderr`, and tsc writes diagnostics to **stdout** — so a build break announced that something broke without saying what. It now includes both streams, and that is how I found my own logger misuse (`log.warning(msg, data)` instead of `log.warning(data, msg)`) at all: the first report of it was a blank error.

Verified: `mcp-node` + `server-core-node` + `arch-lint-node` **333 files / 3940 passed**, `pnpm -r typecheck` clean, `pnpm knip` clean, `pnpm smoke:e2e` ALL OK.

## Summary by CodeRabbit

* **New Features**
  * PNG and SVG exports now identify characters that cannot be rendered by available fonts.
  * Export results include missing characters, with warnings when detected.
  * Font loading failures are handled gracefully, allowing exports to continue.

* **Bug Fixes**
  * Build failure diagnostics now include both standard output and error details.

* **Documentation**
  * Documented font distribution and rendering behavior for non-Latin characters.

* **Tests**
  * Added coverage for missing glyph detection and export handling.

### DIFF
--- a/docs/contributing/adr/0011-font-distribution.md
+++ b/docs/contributing/adr/0011-font-distribution.md
@@ -1,0 +1,138 @@
+# ADR-0011: Fonts are installed by whoever needs them, not bundled
+
+**Status:** Accepted
+
+## Context
+
+The export path draws with four vendored Roboto faces and nothing else. resvg is
+given their paths with `loadSystemFonts: false`, deliberately — the system-font
+scan dominates first-call latency, and no other family appears in the SVG
+`canvas-render` produces.
+
+Roboto is Latin-only. resvg does not substitute a face it was not given. So a
+Japanese canvas exports as tofu boxes:
+
+| | |
+|---|---|
+| `Hello world` | renders |
+| `こんにちは世界` | □□□□□□□ |
+
+**The measurement is already right**, which is what makes this worth an ADR
+rather than a bug fix. `createOpentypeMeasureText` falls back to the
+constant-ratio estimator per code point for anything the face lacks, so the box
+is the right size, the text wraps in the right places, and `truncated`,
+`overflows` and the digest all report a healthy render. Layout is correct and
+the reader cannot read it.
+
+Three constraints shape the answer, and each has already rejected an obvious fix
+elsewhere in this repo:
+
+- **Offline.** The widget smoke asserts that rendering triggers **no network
+  fetch at all**. The export path is likewise local-only.
+- **Byte-identical output.** `canvas-render`'s headline guarantee is that the
+  same scene renders identically on Node, in the browser and on Workers.
+- **Package size.** The published `dist` is **5.2 MB**. Noto Sans CJK Regular
+  alone is **18.6 MB** — bundling it makes the package 4.6x larger for every
+  user, including everyone who never writes a non-Latin character.
+
+## Decision
+
+**Fonts beyond the Latin default are separate packages, installed by whoever
+needs them.** The daemon resolves whatever is present and reports what it could
+not draw.
+
+### 1. The unit is a SCRIPT, not a language
+
+A `lang-ja` and a `lang-ko` package would each carry the same CJK face, because
+Noto CJK covers Japanese, Chinese and Korean in one file. Scripts are what fonts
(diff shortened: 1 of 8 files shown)


kamiazya:
Both findings verified against the code. One is right and is fixed; one is not, and here is why.

## ✅ `loadSystemFonts: true` contradicts the ADR — correct, and a good catch

`headless-renderer.ts:203-205`:

```ts
const fontOption = faces.regular
  ? { fontFiles, loadSystemFonts: false, defaultFontFamily: EXPORT_FONT_FAMILY }
  : { loadSystemFonts: true }
```

The ADR said the export path draws with the vendored faces **"and nothing else"** and rejected `loadSystemFonts: true` unconditionally. Both overstate: host fonts can change exported pixels today, in exactly that branch. That is the kind of thing an ADR being wrong about is worse than not saying at all, because the next reader trusts it.

Recorded as **decision 5** rather than deleted, with the two bounds that make it tolerable:

- **Per-install, not per-canvas.** The condition is "this deployment has no vendored font", not "this text needs a glyph". A working install never reaches it.
- **`undrawable` answers `[]` there**, because nothing was measured against a known face — the same "absence means NOT MEASURED, never fine" rule `overflows` follows. Reporting every character as undrawable would be a louder and wronger answer.

Failing closed is named as a real option and deliberately not taken: it is a behaviour change for a condition nobody has hit, and this ADR is about distribution. The rejection is narrowed to what was meant — rejected as the *answer* for a missing script, not as the last-resort branch.

## ❌ "Use `mcp-node` for this server behavior test" — the test already is

`packages/mcp-server/vitest.node.config.ts`:

```ts
name: 'mcp-node',
include: [
  'src/daemon/**/*.test.ts',
  'src/di/**/*.test.ts',
  'src/server/**/*.test.ts',      // ← undrawable-characters.test.ts lives here
  ...
```

The file is `src/server/export/undrawable-characters.test.ts`, which that glob collects. Measured rather than argued:

```
$ pnpm test --project mcp-node -t "undrawableCharacters"
  Tests  4 passed | 3578 skipped (3582)
```

The guideline it cites is satisfied by placement — `mcp-node` is a vitest *project* over `packages/mcp-server`, not a separate harness a test opts into. No change.

PR #934, opened 2026-08-20
feat(canvas-render): add an injectable content cache for text-node bodies

kamiazya:
## Summary

- `SpatialLayoutOptions.contentCache` memoizes a text node's laid-out body (parse + wrap + fit). Content is laid out **origin-relative** and placed by `placeInNode`, so a cached value is position-independent by construction — the key is only `(width, height, text)`. Validity for everything else (`measure`, theme, `geometry`, `parseBody`, mdast seams) is the caller's documented contract: one cache per document+theme, dropped on theme/font switches.
- Only the success path is cached: the parse-failure fallback recomputes every run so `onDegrade` keeps firing — a cache changes timings, never what the caller is told.
- **Instrument-first, and honest about the ceiling**: on the scene-diff corpus with the arithmetic test measurer, edge routing is **77–86%** of full layout time, which a content memo cannot touch. Measured interleaved (40-node corpus): content-only layout **13.9ms → 0.3ms warm (−98%)**; full layout **53.3ms → 38.1ms (−29%)**. Under a real Canvas/opentype measurer the content share — and the win — is larger. Routing incrementality is recorded as the bigger lever, deferred alongside decision #10's parallel-executor seam.
- `userReach: foundation` — wired by the editor DOM-patch layer (the next stack layer), which owns the per-document cache lifetime (including the layout worker's per-theme map).

Stack layer 6 — base `claude/svg-generation-improvement-uigd90-5` (#933). Landing: top-PR-only merge, see #926.

## Test plan

- [x] New or updated `mcp-node` / `mcp-jsdom` / `mcp-browser` / `web-browser` tests added — `layout/spatial-canvas-content-cache.test.ts` (red-first): byte-equal scenes cold vs warm **including after a move**; zero `measure` calls on a warm run; key sensitivity to text/width/height (**mutation-checked**: dropping `height` from the key turns it red); degraded path uncached with `onDegrade` still firing per run; plus a fast-check property that a shared warm cache never changes any layout
- [x] `pnpm test` passes locally — canvas-render-node **778 passed** (full project), typecheck 0, knip 0, biome clean
- [x] Manual verification completed (describe below)
- [ ] E2E coverage added or extended where applicable — not applicable: opt-in seam, no caller wired yet by design

Manual verification: interleaved timing runs (cold/warm/cold/warm) on the 40-node corpus, numbers above; the no-cache path is bit-for-bit the previous code path (option absent → `cacheKey` undefined → identical control flow).

## Visual evidence

[...]

PR #933, opened 2026-08-20
test(canvas-render): add the scene-diff scoreboard for incremental rendering

kamiazya:
## Summary

- The **instrument-first** half of the editor-incrementalization work: before any keyed-group DOM patching exists, pin how much of a laid-out scene actually changes under a typical single edit. Rendering is a pure per-entry function of the top-level scene entries, so a changed entry is exactly a changed emitted group — the reuse ceiling for a patch layer.
- Deterministic 40-node/57-edge corpus (fake measurer, light theme), every count pinned **exactly** (an improvement in layout stability is as loud as a regression):

| edit | changed / total |
|---|---|
| move far-corner leaf node | 11 / 178 |
| move 4-edge hub node | 8 / 178 |
| edit visible text | 1 / 178 |
| edit past the truncation cut | 0 / 178 (free reuse + truncation-leak tripwire) |
| add disconnected node | 2 / 180 |
| delete one edge | 1 / 177 (surviving routes untouched) |

**Verdict**: single edits touch **0.6–6%** of the scene — the feared edge-routing globality does not materialize for small edits on this corpus, so keyed per-group replacement has a 94–99% reuse ceiling. This is the go/no-go number the DOM-patch layer (next stack layers) was waiting on.

One deliberate scope cut, recorded rather than silent: DOM-update **wall-clock** cost is not measured here — `pnpm bench` is node-only in this repo, and the patch layer's own browser harness will carry that number when it lands.

Stack layer 5 — base `claude/svg-generation-improvement-uigd90-4` (#932). Landing: top-PR-only merge, see #926. (The `<marker>` layer originally slotted here is deferred with reasons recorded: post-gzip savings ≈ 0 against a second-geometry-producer parity cost.)

## Test plan

- [x] New or updated `mcp-node` / `mcp-jsdom` / `mcp-browser` / `web-browser` tests added — `scene-diff-quality.test.ts` (7 pinned cases; the truncation-cut case was found by the instrument itself: an appended-text edit measured 0 because the corpus bodies overflow their boxes, verified against a minimal non-truncated repro before pinning)
- [x] `pnpm test` passes locally — canvas-render-node green including the new scoreboard; typecheck + knip clean
- [x] Manual verification completed (describe below)
- [ ] E2E coverage added or extended where applicable — not applicable (test-only layer)

Manual verification: the "changed: 0" reading was treated as suspect and root-caused (truncation absorbing the edit) with a separate minimal script before being pinned as a deliberate case — not accepted at face value.

## Visual evidence

[...]

PR #935, opened 2026-08-20
feat(canvas-render): add the keyed SVG projection for per-group DOM patching

kamiazya:
## Summary

- `renderSceneToKeyedSvg` (`svg/keyed.ts`) emits **the same document** `renderSceneToSvg` does, cut at the seams a DOM patch layer needs: one `<g data-wb-key="…">` group per top-level scene entry, `#defs`/`#background` pseudo-groups for the chrome outside any entry, plus the root open tag and raw root attributes (for `setAttribute`-style envelope updates on a mounted root). **String equality is change detection** — the serializer is canonical and deterministic, so a patch layer compares group strings and replaces only what differs, with every byte that reaches the DOM still serializer-produced (single-producer preserved).
- Both renderers share the new `buildSvgDocumentParts`, so they cannot drift. Keys come from `scene-entry-keys.ts`, **promoted to production as the one key producer** — the scene-diff scoreboard (#933) now counts by exactly the keys the patch layer will patch by, and its pinned numbers are unchanged by the swap.
- Plain `renderSceneToSvg` output is untouched: the wrappers are an editor-mode projection; exports and the MCP render tool keep the keyless bytes.

Stack layer 7 — base `claude/svg-generation-improvement-uigd90-6` (#934). The next (final) layer mounts this in `apps/web`: mount-once ref container, per-group replacement, the patch≡full-render convergence property, and wiring of #934's content cache. Landing: top-PR-only merge, see #926.

## Test plan

- [x] New or updated `mcp-node` / `mcp-jsdom` / `mcp-browser` / `web-browser` tests added — `svg/keyed.test.ts` (7 cases, red-first): shared-key shape incl. preamble fallback, per-entry wrapping, exact `rootOpen + groups + '</svg>'` assembly identity, **byte-identity to `renderSceneToSvg` with wrappers removed in both envelope modes** (exact reconstruction, no regex), root-attr carry, and single-group change under a single-entry edit
- [x] `pnpm test` passes locally — canvas-render-node **787 passed** (full project; scoreboard pins unchanged after the key-producer swap); typecheck 0; knip clean; biome clean
- [ ] E2E coverage added or extended where applicable — the apps/web layer above carries the browser-mode convergence tests

Manual verification: the assembly-identity and strip-identity tests are constructive proofs over the real pipeline (shared `buildSvgDocumentParts`), not mocks.

## Visual evidence

Plain output byte-identical (full suite unchanged); keyed output is a new opt-in projection consumed by no surface yet.

## Checklist

- [x] Docs updated if this changes user-visible behavior or a published contract — internal package surface, no behavior change

https://claude.ai/code/session_01Kn5gTK3WGQcNrahPppzvHg

---
_Generated by [Claude Code](https://claude.ai/code/session_01Kn5gTK3WGQcNrahPppzvHg)_

### DIFF
--- a/packages/canvas-render/src/scene-diff-quality.test.ts
+++ b/packages/canvas-render/src/scene-diff-quality.test.ts
@@ -1,6 +1,7 @@
 import type { CanvasEdge, SpatialCanvas, SpatialNode } from '@kamiazya/whiteboard-model'
 import { describe, expect, it } from 'vitest'
 import { layoutSpatialCanvas } from './layout/spatial-canvas.js'
+import { sceneEntryKeys } from './scene-entry-keys.js'
 import type { Scene } from './scene-graph.js'
 import { createFakeMeasure } from './test-utils/fake-measure.js'
 import { createSpatialTheme } from './theme/spatial-theme.js'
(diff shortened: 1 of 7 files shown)


kamiazya:
Landed in c23871d via the single squash of the stack's top PR #939 (its diff contained this PR). Closing without a separate merge. (The red `verify` here was the known stdio-exit smoke runner-load flake, diagnosed and recorded; the same suite is green on the landed commit.)

---

PR #940, opened 2026-08-20
refactor(web)!: retire the bespoke browser-local store for the DocumentIndex port

kamiazya:
Slice 3 of the ports initiative. `apps/web` no longer has a bespoke record shape and access path for browser-local documents: everything goes through `packages/ports`' `DocumentIndex`, implemented over IndexedDB (#916), plus a thin app-side layer for the two things the port deliberately does not own.

`apps/web/src/lib/browser-local-store.ts` is **deleted**, with no shim — per the standing "no backward compatibility for internal names" rule.

## What replaced it

| was | is |
|---|---|
| `BrowserLocalStore` (one object) | `DocumentIndex` + `DefaultDocumentPointer` + `ContentClock` |
| `save()` (upsert, metadata) | `createDocument` / `setDocumentName` |
| `del()` (compare-and-delete against the pointer) | `deleteDocument` by path, with the pointer check moved into the controller |
| `generateId()` | the index mints ids |
| `load()`'s three outcomes (`ok`/`not-found`/`corrupted`) | a snapshot or `null` — whether the *content* reads is `LoroStore.load`'s answer |
| `updatedAt` on the metadata row | the content record's envelope (the #919 fix, now the only source) |

## Three production gaps the conversion exposed

Each was found by a test, not by reading the diff.

**1. Nothing created the `local` workspace.** The port answers a list against an unknown workspace with `WorkspaceNotFoundError`, not `[]` — so a fresh browser failed its first list *and* its first create. `ensureLocalWorkspace` now runs on every create and before the list page's first read.

**2. v9 left `documentIndex` empty**, on the recorded promise that the bespoke `documents` store would keep serving reads *"until its call sites move"*. This PR moves them. **v10 backfills the index from `documents` and drops it** — without it, an upgrading user opens the app to an empty list with their bytes still on disk. Two tests pin it, including the empty-workspace case.

**3. The `'untitled'` title sentinel stopped matching.** Auto-names are paths now, and the second one is `untitled-2` — an ordinary path, not a sentinel. A name equal to its own path is the one nobody chose, which is exactly how the index spells it (`name` omitted), so that is the rule the title box uses and the sentinel is deleted rather than widened.

## Two tests that had quietly stopped checking

- Every create path seeds a content record now, so `expect(keys.length).toBeGreaterThan(0)` is true *before any edit lands*. Two persistence tests polled on that, stopped retrying immediately, and then asserted on a canvas that was empty (`expected [] to include 'reload-regression-node'`). They now wait on the record's **content**, with the re-fire inside the same wait.
- The real-IndexedDB browser suites are back on `IdbDocumentIndex`. An in-memory double there would have made "survives a remount" vacuous.

## Also

- The controller takes **named deps** instead of five positionals — the bespoke store's single object became four collaborators, three of them optional.
- `BrowserLocalIndexPage` now creates through the editor's own `createSeededDocument` instead of a second, subtly different create path.
- `startFresh` clears the pointer before dropping the stale document: the old `del` cleared it as a side effect, and deleting by path has no such coupling, so a failed repoint would otherwise leave the pointer aimed at a deleted document.
- `noContentClock` and `fake-indexeddb` dropped (knip).

## Verification

| suite | result |
|---|---|
| `apps/web` jsdom | 2595 passed |
| `web-browser` (real Chromium, real IndexedDB) | 707 passed |
| `mcp-node` + `arch-lint-node` | 3707 passed, 6 skipped |

All run against the branch **after** merging current `origin/main`.

No visual evidence attached: this change has no intended pixel effect, and the user-visible surfaces it touches (list, title box, degraded banner) are pinned by the browser suites above.

## Summary by CodeRabbit

* **New Features**
  * Added browser-local document summaries with names, paths, timestamps, and default-document handling.
  * Improved local document creation, loading, switching, duplication, deletion, and persistence.
  * Added migration support for existing local documents and workspaces.

* **Bug Fixes**
  * Improved recovery from missing documents and failed storage operations.
  * Fixed timestamp reporting and ensured storage transactions no longer hang after errors.


kamiazya:
## CodeRabbit findings — all nine verified, all nine real, all nine fixed

Each was checked against the code rather than accepted or dismissed. Two were **production defects**, not test hygiene.

### 🔴 The list page re-read the index on every render (Critical — confirmed with a number)

`clock = idbContentClock()` as a **default parameter** is evaluated on each render, so the load effect's `[index, clock]` deps changed every time → `setSnapshots` → re-render → repeat. `App.tsx` passes only `index` and `onOpenDocument`, so this was the path production takes.

Reproduced on that exact wiring before fixing: **464 index reads and still climbing** after half a second. Defaults hoisted to module level (the controller already did this); `BrowserLocalIndexPage.defaults.browser.test.tsx` mounts the page the way production does and asserts the count *stops*, rather than that it is small — a fixed bound could pass by accident on a slow machine.

### 🟠 Deleting a canvas left the default pointer naming it (Major — my comment was false)

The code carried a comment claiming the editor's resume path "already falls back safely". It does not: it reports `The canvas data could not be read.` So an ordinary delete handed the user an error screen the next time they opened the editor. Red test first (`expected '0CFJNRVY…' to be null`), then resolve the pointer before the delete and clear it when it matches.

### Three tests that had stopped exercising anything

All the same shape — stubbing a method on the **test double** while the code under test reads the **index**:

| test | stubbed | actually read |
|---|---|---|
| import panel's `listDocuments` rejection | `store.listDocuments` | `index.listDocuments` |
| stale-link held enumeration | `store.listDocuments` | `index.listDocuments` |
| deep-link read failure | `index.listDocuments` | `index.resolveDocument` |

Each is repointed and **mutation-checked**: break the production `catch` and they go red, restore it and they go green. Before the fix they passed under both. The deep-link test was also renamed — it says "path read" now, because that is what it stubs.

### The rest

- Both test doubles register the local workspace, so an unseeded one **lists** where production returns an empty list instead of throwing `WorkspaceNotFoundError`.
- Three doc comments named the wrong collaborator — damage from the bulk conversion's `sed`, caught here.

Local after the fixes: jsdom 2596, node 77, `web-browser` 708, `knip`/`typecheck`/`lint` clean.

PR #943, opened 2026-08-21
feat(web)!: store document files as content-addressed blobs

kamiazya:
ports slice 1. `BlobStore` gets its IndexedDB implementation, and `DocumentFileStore` becomes a **fileId → `BlobRef` mapping** over it.

## Why the fileId stays

The fileId is not an internal handle — it IS an address. The string `newImageRef` builds is written **into the document**, and the daemon's file route validates it. Moving documents to content addresses changes a published shape, which the vocabulary rule calls its own increment.

So the bytes moved and the address did not, and this mapping is what sits between them.

## What that buys — both named as deliberate follow-ups in the old store's own comment

| | before | after |
|---|---|---|
| same image in two documents | two copies, nothing could tell | **one copy** |
| deleting a reference | *"Records are never deleted here, so the store grows unbounded"* | **dropped, and the bytes go once nothing names them** |

Deleting one of two references must *not* blank the other's image — that half is what made `delete` impossible to bolt onto the old store, and it has its own test.

## Old records

A record written before this reads as `v1` and is **converted in place on first read** — the bytes are already in hand there, so it costs one write the first time an old image is displayed. Not an upgrade-transaction migration, because `crypto.subtle.digest` cannot be awaited inside a versionchange transaction without losing it.

`ponytail:` a v1 record nobody ever reads is never converted. That is acceptable because it is exactly the record a reference-sweeping GC would collect anyway.

## The port's guarantees are a conformance suite now, not prose

`describeBlobStoreConformance` lives beside the port, and **all three** implementations call it — in-memory, filesystem, IndexedDB — so "content-addressed" cannot mean three things in three places. The two Node suites keep only what is theirs to ask: on-disk sharding, the base64 envelope, filesystem failures.

The suite was mutation-checked against a known-good store before being trusted — drop the defensive copy, invent a `contentType`, break the digest, skip the delete: **each goes red**, and the restore goes green. The mapping layer's three new rules were mutation-checked the same way.

Also `inTransaction`/`request` moved to `idb-tx.ts`. A port's atomicity guarantee is bought by the shape of those two functions, and the blob store made them the second caller.

`DB_VERSION` 11 adds the `blobs` store. Purely additive.

## Verification

| suite | result |
|---|---|
| `apps/web` jsdom | 2591 passed |
| `web-browser` (real Chromium, real IndexedDB) | 724 passed |
| `mcp-node` + `arch-lint-node` + `ports-node` | 3851 passed, 6 skipped |

All run after merging current `origin/main`.

No visual evidence: the user-visible surface is an image that displays exactly as before. What changed is how many copies of it are on disk, which the browser tests assert by counting rows.


kamiazya:
## Manual verification on the Pages preview

Tests cover dedup and delete against real IndexedDB in real Chromium. What they cannot cover is the **v10 → v11 upgrade running on a browser that already holds data**, so that was checked against the shipped bundle at `ports-blobstore.kamiazya-whiteboard.pages.dev`.

`controlled: false` and one unregistered worker first — this is the PR's code, not a cached bundle.

**1. The schema shipped.** `whiteboard` opens at version `11` with stores `blobs, documentFiles, documentIndex, loroDocuments, meta, workspaces`.

**2. Upgrading a real v10 database with a v1 record.** Seeded a v10 database holding the OLD `documentFiles` shape (`{v:1, mimeType, created, blob}`, 11 bytes), then loaded the app:

| | result |
|---|---|
| version after load | `11` |
| `blobs` store | created |
| the v1 record | **survived intact** — still `v:1`, still holds its 11-byte Blob |
| `blobs` contents | empty |

The empty `blobs` store is the design, not a miss: the conversion is **read-through**, so it fires when that image is first displayed, not at upgrade time. Purely additive, nothing at risk during the upgrade itself.

**3. The write path still works.** Created a canvas from the empty state:

```json
{ "url": "/local/untitled",
  "createdDocuments": [{ "path": "untitled", "kind": "spatial", "hasName": false }],
  "contentRecords": 1,
  "defaultPointer": ["defaultDocumentId"],
  "errorBanner": null }
```

Three things worth noting there, all from the slices before this one and all confirmed in a real browser for the first time: `hasName: false` is the index storing "no name of its own" as absence; `contentRecords: 1` is the content seed every create path now writes; and the empty list rendered its empty state with **no error banner**, which is `ensureLocalWorkspace` doing its job — before #940 a device with no workspace answered `WorkspaceNotFoundError` and the page read as a failure.

The fixture database, caches and worker were removed from the preview origin afterwards.

PR #948, opened 2026-08-21
feat(web)!: put the browser's Loro persistence behind the DocumentStore port

kamiazya:
ports slice 4, **foundation half**. The port gets its third implementation and, first, the suite that says what the port actually promises.

## Writing the suite found a real disagreement

`DocumentStore`'s guarantees were spread across three files: the in-memory double's tests, the libSQL store's tests, and a **parity property** between them. That arrangement answers *"do these two agree?"* but not *"what is the contract?"* — so a third implementation had nothing to be held to.

Writing it out found that the in-memory double returns chunks in **insertion order** where the real store sorts by index (`order by chunkIndex`, and its own test says *"Insert out of index order to prove loadSnapshot re-sorts"*). The parity property missed it because its generator only ever produced chunks in order.

**Two implementations agreeing is not the same as a contract** — and the double is what most tests run against.

## Mutation-checked before being trusted

Six mutations against a known-good store, each expected red:

| mutation | caught |
|---|---|
| fold a saved-empty snapshot into "absent" | ✅ |
| merge chunks instead of replacing them | ✅ |
| stop moving the frontier on append | ✅ |
| actually filter by `sinceFrontier` | ✅ |
| delete only half a document | ✅ |
| hand back the caller's array | ✅ |

Two earlier attempts **did not apply cleanly** — one patch matched 0 occurrences, one mutation was a no-op that fell through to the same value — and were **redone rather than counted**. A mutation that never landed proves nothing.

Three more were run against the *browser* store, to check the suite binds to that implementation and not only to the double.

## `docRefKey` moves to `ports`

It is a **stored** key. Two stores that spell it differently cannot read each other's documents, with nothing to say so at compile time — which its own doc comment already argued, from beside one implementation. The daemon's migrations keep their frozen literals.

## One record, not four tables

The daemon needs `documentSnapshots` + `documentSnapshotChunks` + `documentFrontiers` + `documentDeltas` and a SQL transaction across them, because a **row** is its unit. A **record** is the unit here, so all four live under one key. Two things the daemon has to be careful about therefore cannot happen: a replaced snapshot cannot orphan a chunk, and `deleteDoc` cannot half-succeed.

## A comment that is a measurement, not an assumption

Removing the defensive `copy()` leaves the **whole suite green** — IndexedDB clones during `put` and hands each `get` its own buffers, so a caller mutating after awaiting reaches neither. What it *does* close is the window between `saveSnapshot` being called and its internal read resolving. The comment says exactly that instead of claiming more.

## Reach

**`foundation:` nothing calls `IdbDocumentStore` yet.** Wired by the filed follow-up *"ports 5: LoroStore を DocumentStore の上に載せ替える"*, which also has to settle a dependency this slice surfaced: `idbContentClock` reads `updatedAt` off the `loroDocuments` envelope, and the document list's timestamps go with it. Deciding that **before** the swap is written into the ticket, because not deciding it drops every listing to the epoch.

`DB_VERSION` 12 adds `syncDocuments`. Purely additive, and deliberately **not** where `LoroStore` writes — two writers on one shape during a transition is how a subtle corruption gets in.

## Verification

| suite | result |
|---|---|
| `apps/web` jsdom | 2597 passed |
| `web-browser` (real Chromium, real IndexedDB) | 745 passed |
| `mcp-node` + `ports-node` + `arch-lint-node` | 3869 passed, 6 skipped |

All run after merging current `origin/main` (including #947, which touches the same `browser-idb.ts` — the merge keeps both its test-isolation seam and this slice's `DB_VERSION` bump).

No visual evidence: nothing here reaches a pixel. The user-visible check belongs to the follow-up that wires it.

## Summary by CodeRabbit

* **New Features**
  * Added browser-based document storage with snapshots, deltas, frontiers, compaction, and document deletion.
  * Added clearer handling for missing, malformed, and unsupported document data.
  * Added shared document-storage behavior across supported storage implementations.

* **Bug Fixes**
  * Improved snapshot chunk ordering, data isolation, and transactional persistence.
  * Migrated existing browser data to the updated storage format while preserving document content and timestamps.

* **Tests**
  * Expanded migration, persistence, compaction, error-handling, and multi-document coverage.


kamiazya:
## Manual verification of the v11 → v12 upgrade

The store itself is foundation-only, but `DB_VERSION` 12 is a schema change **every existing user's browser runs**, and that is the part tests cannot fully derisk. Checked against the shipped bundle on the Pages preview.

Service worker unregistered and caches cleared first, so this is the PR's code rather than a cached one.

Seeded a **real v11 database holding data** — a workspace, an index row, its `loroDocuments` content record, and the default pointer — then opened the app at that document's deep link:

| | result |
|---|---|
| version after load | `12` |
| `syncDocuments` | created |
| the index row | **untouched** — `pre-v12` / `Pre v12` / `spatial` |
| its content record | **untouched** — `v:1`, 5 bytes, original `updatedAt` |
| `syncDocuments` contents | **empty** |
| deep link | resolved; `<h1>` read `Pre v12`; no error banner |

The empty `syncDocuments` is the claim, not a gap: this slice writes nothing into it. That is what "purely additive" means here, and it is why the upgrade cannot put existing data at risk — there is no conversion step to get wrong. The conversion belongs to the follow-up that wires `LoroStore`, and it will need its own verification of exactly this kind.

The fixture database, caches and worker were removed from the preview origin afterwards.

kamiazya:
## Slice 5 folded in — nothing here lands unwired

Per the decision to do the wiring in the same PR, `LoroStore` is now a thin layer over `IdbDocumentStore`. Title updated (this is a breaking storage change, not just an addition).

### Two contract gaps the wiring hit, both decided before implementing

**`saveCompactedSnapshot` is a new port operation.** Folding a delta log into a snapshot and dropping the log was not expressible — `deleteDoc` was the only thing that cleared a log, so a log had **no way to stop growing while the document lived**. That would have regressed a fixed bug (`COMPACT_DELTA_BYTES` carries the measurements). One operation rather than save-then-clear, because the two-call version has a window where a concurrent append is dropped.

**`StoredDocumentUnreadableError` is a new port failure.** `T | null` folded *"written by a newer build"* into *"you have no such document"* — opposite sentences to say about someone's own data, and four user-facing messages depended on telling them apart.

**One requirement in the suite was wrong and was corrected while implementing it.** The unreadable-record seam originally took a `code` naming which shape to write. `unsupported-version` is **not a state the libSQL store can be in** — its records are typed columns, with no envelope version to be wrong about. The shared bar is now "throws rather than answering null"; which codes an implementation distinguishes is its own test's business.

### `updatedAt` moved to its own store

It belongs to neither port — `DocumentStore` knows only frontiers, `DocumentIndex` deliberately holds only placement and naming. It now sits beside the default-document pointer, and the listing no longer parses a whole snapshot record to read one string.

`DB_VERSION` 12 carries every `loroDocuments` record across: snapshot, delta log **and** timestamp. The log travels because a record that had one is a document whose snapshot alone is not its current state.

### Three defects found by measuring, not by arguing

**A test that never waited for the write.** Going through the port made `appendDelta` three IndexedDB round trips instead of one, widening a window the node-lock suite had always raced. Three hypotheses were tested and **refuted** first — a vacuous wait, `waitFor`'s act batching, and the remount opening a different document — before a trace of the whole chain (`push → append → deliver snapshot+1`) showed both appends running while the reload read one delta and no lock. The test waits on stored state now.

**A cross-realm `Uint8Array`.** Under `fake-indexeddb` a stored array reads back with `constructor.name === 'Uint8Array'` and `instanceof Uint8Array === false`, so a schema gated on `instanceof` rejects every good record as corrupt. Measured directly rather than guessed — I had dismissed this hypothesis once before testing it. `storedBytesSchema` asks the internal brand instead, and copies into this realm. (main hit the identical trap in #961 and reached the same conclusion independently, which is a useful second opinion on the diagnosis.)

**Waits budgeted longer than their own test.** The six suites main moved to jsdom kept 10s waits under a 5s default — a wait that actually has to wait could never finish. Latent before; my change made it bite.

Two fixtures were also building their own IndexedDB schema by hand, so they were constructing databases without the store under test in them. Both use the real opener now.

### Verification, after merging current `origin/main` (including #961)

| suite | result |
|---|---|
| whole root suite (`pnpm test`) | **8931 passed**, 6 skipped |
| `apps/web` jsdom + node | 2618 + 77 |

The v11 → v12 upgrade still needs the same real-browser check the earlier commit got; I will run it on the refreshed preview.

kamiazya:
## CodeRabbit triage — nine of ten real, and two were the new contract broken by its own implementations

Each finding was checked against the code. Two were defects in the operation this PR *adds*, which is the most useful kind of review to get.

### 🔴 `saveCompactedSnapshot` was unsafe as specified

*"Save this and drop the log"* cannot be right, and the reviewer's interleaving point is what showed it. The caller folds a log it has **read**; an append arriving before the store writes is neither in that snapshot nor superseded by it — so clearing the whole log loses an edit, **in every implementation**, not just the one flagged.

The input carries `supersededDeltaCount` now and each store drops exactly that prefix.

**The conformance case that proves it went through two drafts**, and the first one is worth naming: it allowed an empty log as an acceptable outcome, and so passed against a store that demonstrably loses the update. The assertion was at fault, not the store. Tightened, it goes red against the in-memory store and green after the fix.

### 🔴 The in-memory store had the exact window the operation exists to close

It delegated to `saveSnapshot` and then cleared; the `await` between them let a concurrent append land and be thrown away. One map write now.

### 🟠 `LoroStore.load` was serialised against nothing

Two port calls — snapshot, then log — with a compaction able to land between them, handing back a pre-compaction snapshot with a post-compaction *empty* log. Nothing re-reads afterwards, so that stale answer is what the editor would open.

### 🟠 The v12 migration destroyed records it could not parse

It skipped them and then deleted the store, so an envelope from a newer build went from **"unreadable, and recoverable"** to gone with its bytes — undoing the very distinction the rest of this PR adds. Carried verbatim now, with a regression test; mutation-checked (skipping instead of carrying goes red).

### 🟠 The kind-collision case never collided

`TREE.workspaceId` and `DOC.documentId` were different strings, so a store keying on the id alone passed. Same string now — dropping `kind` from `docRefKey` goes red, and did not before.

### The rest

| finding | |
|---|---|
| schema accepted records it could not serve | fixed — manifest/chunk agreement, and a snapshot must carry its frontier (one direction only: a delta-only document is legal) |
| `vi.waitFor` ignored the file's configured budget | fixed — 1000ms for the one condition a remount depended on |
| two comments described arguments that no longer exist | fixed |

**Not taken:** `ponytail:` is this repo's marker for a deliberate shortcut with a named ceiling (AGENTS.md's Source Comment Discipline), not an editing artifact.

Whole root suite after the fixes: **8935 passed**, 6 skipped. `typecheck` / `lint` / `knip` clean.

kamiazya:
## Manual verification of v11 → v12, on the real bundle

Service worker and caches cleared first, so this is the PR's code.

Seeded a **real v11 database** — the Loro bytes were produced by the app itself, then rewritten into the old envelope — holding two documents: a normal one, and one whose envelope this build cannot parse (`{v: 99, …}`), which is the case the migration used to destroy.

| | result |
|---|---|
| version after load | `12` |
| `loroDocuments` | **gone** |
| stores | `blobs, contentTimestamps, documentFiles, documentIndex, meta, syncDocuments, workspaces` |
| the real document's snapshot | carried — 81 bytes, envelope `v:1` |
| its `updatedAt` | carried — `2026-08-01T00:00:00.000Z` |
| **the unparsable record** | **carried verbatim** — `{v: 99, writtenByANewerBuild: true}` |
| deep link | resolved; `<h1>` read `Carried`; no error banner |

That last row is the CodeRabbit finding, confirmed fixed in a real browser rather than only in a test: before, the migration skipped it and then deleted the store, so an envelope from a newer build went from *"unreadable, and recoverable"* to gone with its bytes.

### One thing this surfaced that is NOT in scope here

Opening the future-version document shows an **empty canvas with no banner**, even though `LoroStore.load` correctly answers `unsupported-version` and the backend forwards it. The bytes are safe (`bytesStillThere: true`) — the gap is that the classification never reaches the screen, so the user is told their document is empty rather than that their app is old.

Pre-existing wiring, not introduced here: this PR's change was to stop the classification being *lost*. The copy already exists (the import panel prints "Canvas data uses an unsupported storage version."). Filed as its own ticket rather than widened into this diff.

The fixture database, caches and worker were removed from the preview origin afterwards.

PR #950, opened 2026-08-21
feat(web): lay the markdown preview out as a page, not as a node

kamiazya:
> **Top of a 2-PR stack** (#951), on top of #949. Only this layer's diff is shown here.

## Summary

The preview pane runs to a readable measure and was rendering at **node** metrics — a heading scale cut so it cannot eat a third of a 280px box, applied to a full-width column. #949 made the metrics a parameter; this passes the document one.

## Set on `layoutScene`, not on the preview's own call

`renderMarkdownPreview` and `layoutMarkdownOutline` both go through `layoutScene`, and the minimap rail maps the blocks the second one returns. Two scales here would leave the rail marking a document the reader is not looking at, at heights that do not match the pane.

That is **pinned rather than trusted**: a test asserts the rail's blocks equal the pane's.

## Test plan

- [x] Two tests: the heading lands at the document size (and *not* the node one), and the rail's blocks equal the pane's.
- [x] **Mutation-checked** — reverting the one line drops the heading back to node size and fails the surface test.
- [x] `apps/web` both suites: **2593 + 77 passed**; `pnpm -r typecheck` clean; `pnpm lint` exit 0.

## Checklist

- [x] Commit message follows Conventional Commits


kamiazya:
@coderabbitai review

PR #966, opened 2026-08-21
feat(web): one document browser for both modes, with match highlighting

kamiazya:
The three-pane document browser (tree / folder contents / preview) was **daemon-only** because `WorkspaceFilesPanel` called the daemon's client functions directly — the single obstacle the one-browser-both-modes ticket named, cleared by the ports work. This PR puts the seam in and wires both sides, and upgrades the search results per the design review.

## The seam: `WorkspaceFilesSource`

Five operations, measured from the panel's imports rather than assumed: list, create, rename, and the two content reads its thumbnails and preview need. Duplicate and delete stay with the page, which owns the confirmation dialog.

| adapter | backing |
|---|---|
| `createDaemonFilesSource` | the same client calls the panel used to make; its 404 becomes the seam's `WorkspaceMissingError` |
| `createLocalFilesSource` | the same `DocumentIndex`/`LoroStore` pair the editor uses — browser and editor cannot disagree about what exists |

The content reads take the whole **entry**, not a field chosen by the loader: the daemon addresses a snapshot by path and OKF by id, the local store keys everything by id. Choosing the field in the loader is exactly the decision that made it daemon-only.

## Local mode gets the tree

`BrowserLocalIndexPage` gains the same grid/tree toggle the daemon page has had since #902. Phase 2 (retiring the grid + migrating its 38 tests) stays its own increment, per the standing ticket — it needs no further design input now that search has one.

## Search results: highlight + list/grid, per the ten-app survey

Design reviewed against Finder, Explorer, VS Code, Obsidian, Notion, Drive, Dropbox, Figma, Linear and Slack (full survey in the session):

- **Match highlighting** — near-universal in mature search UIs. First case-insensitive match marked in the title *and* the path; first only, because four highlights in one title read as damage.
- **The path stays on every row in BOTH layouts.** The two surveyed apps that flatten hierarchy without per-row location context — Finder and Figma — both carry standing user complaints about it. The grid does not get to reintroduce that gap.
- **List first**, grid as the toggle: the row form carries the path inline, the grid trades density for a bigger picture, Finder-icon-view style.
- Cross-workspace scope unchanged and deliberate (`search-documents.ts` documents why). Body search and a scope toggle are filed separately — full-text needs an **index**, not a UI.

## A test premise refuted while writing it

The local source's missing-workspace case cannot be reached on a fresh database: the v8+ upgrade seeds the `local` workspace row unconditionally. The test pins the reachable behaviour instead (fresh database lists empty), and the `WorkspaceMissingError` mapping stays as a guard for hand-edited stores.

## Verification

| suite | result |
|---|---|
| `apps/web` jsdom + node | 2640 + 77 |
| `web-browser` | 736 passed (1 isolated-green typing flake from the documented family, zero reachability to this diff) |
| workspace-files component suites | 104 |
| `typecheck` / `knip` | clean |

Daemon-mode behaviour is preserving-by-construction (the adapter wraps the same calls) and held by `DaemonIndexPage.test.tsx`'s 38 tests. Local-mode manual verification on the Pages preview is below — daemon mode is origin-locked off previews.

## Visual repro (local mode, Pages preview)

Verified live on the branch preview (fresh origin, service worker uncontrolled). The whole loop was driven through the new seam: create → move to `plans/roadmap` → edit in the real editor → browse → search → delete-with-refresh.

**The three-pane browser now serves local mode** — tree, folder contents with real markdown thumbnails, preview rendering the document body, and Move/Open/Delete:

**Search replaces the pane with a flat cross-folder list and marks the match** — `marketing-plan` matches in the title, `plans/roadmap` only in the path, and both say so:

**The grid layout trades density for the picture and keeps the path on the card** (the Finder/Figma gap stays closed):

Also exercised, not pictured: rename to a nested path grows the tree a `plans` folder; deleting a document from a search result closes the dialog and drops the row without a reload.

### DIFF
--- a/apps/web/src/components/workspace-files/SearchResults.test.tsx
+++ b/apps/web/src/components/workspace-files/SearchResults.test.tsx
@@ -1,0 +1,63 @@
+/**
+ * What a search result row must carry: the match, made visible.
+ *
+ * Prior-art survey (Finder, Explorer, VS Code, Obsidian, Dropbox, Notion,
+ * Slack): highlighting the matched substring is near-universal in mature
+ * search UIs, and the two surveyed apps WITHOUT per-row location context
+ * (Finder, Figma) both have standing user complaints about it. This file pins
+ * the two things that make a flat cross-workspace result list legible: the
+ * highlighted match and the path on every row — plus the list/grid toggle.
+ */
+import { cleanup, fireEvent, render, screen } from '@testing-library/react'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import type { WorkspaceDocumentEntry } from './document-entry.js'
+import { SearchResults } from './SearchResults.js'
+
+afterEach(cleanup)
+
+const entries: WorkspaceDocumentEntry[] = [
+  { documentId: 'd1', path: 'plans/roadmap', name: 'Road map', kind: 'markdown' },
+  { documentId: 'd2', path: 'sketches/map-of-rome', kind: 'spatial' },
+]
+
+describe('SearchResults', () => {
+  it('highlights the matched substring in the title', () => {
+    render(<SearchResults results={entries} query="road" onSelect={vi.fn()} />)
+    const marks = screen.getAllByTestId('search-match')
+    // Case-insensitive, and in the TITLE — the visible answer to "why is
+    // this row here".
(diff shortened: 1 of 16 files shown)


kamiazya:
@coderabbitai review

kamiazya:
@coderabbitai review

PR #989, opened 2026-08-22
feat(web): a document card's actions open at the card — right-click context menu

kamiazya:
Increment D of the **Document First** direction (A = #971, B = #978, C = #982): OOUI's noun-then-verb, at the object. Right-clicking a document card (folder pane) or a search result row opens **Open / Duplicate / Move… / Delete** at the cursor.

- Every item fires the **same handler the preview pane uses** — no second set of rules. Duplicate appears only when the page provides it; Delete carries the kind so the dialog says note/canvas correctly (#978's contract).
- **Move…** selects the document and opens the preview's move form in one gesture. The bump-token effect is deliberately defined AFTER the preview's selection-change reset — defined earlier, the reset ran second and undid the edit in the very commit where selection and token change together (caught red in the new test).
- The menu is the existing `ContextMenu` (role=menu, Escape handling, focus-on-open — already shared: MarkdownEditor imports it too) with its hardcoded aria-label made a prop: **"Document actions"** here, the spatial surface's "Canvas actions" default untouched.
- Folders and empty space open no menu (native browser menu preserved).

Six red-first jsdom tests pin: the verb set, Open/Delete dispatch (kind included), Duplicate omission, Move-opens-the-form, search-row parity + Escape close, folder-no-menu.

Verification: apps/web jsdom 2679 + node 77 green (one earlier run had a single load-dependent failure in an untouched file; full rerun clean), workspace-files browser files + flow + spatial context-menu browser test 8/8, lint / knip clean. Live preview verb-by-verb check follows as a comment.

### DIFF
--- a/apps/web/src/components/spatial-editor/ContextMenu.tsx
+++ b/apps/web/src/components/spatial-editor/ContextMenu.tsx
@@ -73,6 +73,11 @@ export interface ContextMenuSeparator {
 export type ContextMenuItem = ContextMenuActionItem | ContextMenuOptionsItem | ContextMenuSeparator
 
 export interface ContextMenuProps {
+  /**
+   * Accessible name of the menu. The default names the spatial surface;
+   * object menus (a document card's) pass their own.
+   */
+  readonly label?: string
   /** Screen position relative to the editor root. */
   readonly x: number
   readonly y: number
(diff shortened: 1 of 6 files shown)


kamiazya:
**Live verification** (Pages preview, fresh origin, SW uncontrolled; driven through the real DOM — the verification browser tab was backgrounded, so evidence is the rendered state):

- Right-click on the seeded note's card → `role="menu"` "Document actions" with exactly `["Open","Move…","Delete"]` — **Duplicate correctly absent** (the local index page provides no duplicate handler).
- **Move…** → `{"moveFormOpen":true,"value":"untitled","menuClosed":true}` — one gesture selects and opens the preview's path form.
- **Delete** → the kind-aware dialog from #978: `Delete "untitled"? This permanently removes the note from this browser.` → cancelled.
- **Open** → navigated to `/local/untitled`.

PR #996, opened 2026-08-22
feat(web): one Rename dialog for a document's name and its path

kamiazya:
Increment E of the **Document First** direction (A #971, B #978, C #982, D #989), per the **Q4 decision: one dialog** rather than explaining a split.

A document has a **name** (what it is called; optional — readers fall back to the path's last segment) and a **path** (where it lives; moving takes the subtree). Both were editable, in two different places, with nothing on screen saying how they differ. **Rename…** now opens one dialog carrying both fields, each with a line explaining its role:

> **Name** — What it is called, everywhere it appears. Leave empty to show the last part of the path instead.
> **Path** — Where it lives in the workspace. Moving it takes everything under it along.

## The model underneath is unchanged

- The name still lives in the **workspace**: the seam gains `setDocumentName`, mapping to the daemon's `PUT …/name` (whose "clear" spelling is the empty string) and to `DocumentIndex.setDocumentName` by id (whose clear is *absence*). Both spellings are pinned by adapter tests.
- The path still moves through the existing `renameDocumentPath`.
- **Neither field derives the other** (ADR-0008).
- Saving applies **only what changed**; a collision shows the server's own message, because it names the PRODUCED path — on a subtree move often not the one typed.

## What this retires

The preview pane's inline move form and the context menu's `Move…`-plus-bump-token plumbing both go; each now opens this dialog. The token's ordering hazard from #989 (an effect that had to be declared *after* the selection reset or the edit was undone in the same commit) disappears with it.

## Tests

Eight new dialog tests (prefill + both explanations, name-only, clear-as-absence, path-only, both, server refusal keeps the dialog open, context-menu parity, cancel touches nothing), preview tests migrated to the Rename… affordance, `files-tab` move tests re-pointed at the dialog, adapter tests for both clear spellings.

Verification: apps/web jsdom 2697 + node 77, workspace-files browser + flow 6/6, lint / knip / typecheck clean. Live preview walk-through follows as a comment.

### DIFF
--- a/apps/web/src/components/workspace-files/DocumentPreview.test.tsx
+++ b/apps/web/src/components/workspace-files/DocumentPreview.test.tsx
@@ -1,4 +1,4 @@
-import { act, cleanup, fireEvent, render, screen, waitFor } from '@testing-library/react'
+import { act, cleanup, fireEvent, render, screen } from '@testing-library/react'
 import { afterEach, describe, expect, it, vi } from 'vitest'
 import { DocumentPreview } from './DocumentPreview.js'
 import type { WorkspaceDocumentEntry } from './document-entry.js'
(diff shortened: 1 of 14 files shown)


kamiazya:
**Live verification** (Pages preview, fresh origin, SW uncontrolled; driven through the real UI and read from the live DOM — the verification tab was backgrounded, so the evidence is rendered state rather than pixels). Two seeded documents, all four paths through the dialog:

1. **Opens prefilled, both roles explained** — Path `untitled`, Name empty, and the two help lines render verbatim: *"What it is called, everywhere it appears. Leave empty to show the last part of the path instead."* / *"Where it lives in the workspace. Moving it takes everything under it along."*
2. **Name only** → card reads `Release plan`, the path in the preview is unchanged.
3. **Clearing the name** (whitespace) → card falls back to the segment: `untitled`.
4. **Path only** → `plans/roadmap`: breadcrumb becomes `Workspace/plans`, the pane lists `roadmap`, the preview follows the document.
5. **Collision** (moving the second document onto the occupied path) → the dialog **stays open**, shows the server's own words — `Document path "plans/roadmap" already exists in workspace "local"` — and the Path field keeps what was typed.

One note for anyone reproducing this from a script: the closed dialog's node lingers with `data-state="closed"` in a backgrounded tab, because Chrome freezes CSS animations there and Radix unmounts on animation end. `data-state`, not node presence, is the honest check.

PR #1044, opened 2026-08-23
feat(mcp): wb_workspace_edit — one call for a batch of document operations

kamiazya:
`wb_canvas_edit` already lets an agent change a spatial document in one call (ADR-0010). Everything a workspace *contains* still took one round trip per document: create, set, delete — each its own tool call, each its own turn. Seeding a workspace with five notes cost ten calls.

`wb_workspace_edit` applies an ordered list of ops in one call and reports what it did per op:

```jsonc
{ "workspaceId": "default", "ops": [
  { "op": "document.create", "path": "notes/a", "kind": "markdown", "markdown": "---\ntype: note\n---\nalpha" },
  { "op": "document.create", "path": "board",   "kind": "spatial", "name": "sketches" },
  { "op": "document.set",    "documentId": "…", "markdown": "…" },
  { "op": "document.delete", "documentId": "…" }
]}
```

The ids it mints come back in `results`, so a follow-up op in the *same* batch, or the next call, needs no round trip to learn them.

## It does not roll back, deliberately

A workspace is not one CRDT. Each document is its own `LoroDoc` with its own snapshot, so "undo the ops that already landed" would mean a second batch of writes that can fail exactly the way the first one did — and a half-undone undo is a worse state than the one it replaces.

So the failure is *reported* rather than hidden. The error names the index that failed, says how many ops before it stand, and tells the caller to resume from that index:

```
ops[2] (document.create) could not be applied: path "board" already exists.
2 op(s) before it were applied and stand; ops after it were not run.
Resume from this index rather than resending the batch.
```

Saying "nothing was written" here would be false, and a caller acting on it would create the first two documents twice.

## Also fixes the orphan CodeRabbit found in #1041

`wb_document_create` with a malformed markdown body created the document, *then* failed applying the body — leaving an empty document squatting the requested path while the caller held an error saying the create had not happened. The retry then collided with a ghost. The body is now parsed before anything is written, and after the workspace bootstrap so a missing workspace still reports itself first.

Reproduced first (`document-create-orphan.test.ts`), and mutation-checked: neutering the preflight turns it red.

## Verification

- `server-core-node` 393 passed; `mcp-node` mcp suite 124 passed
- `pnpm smoke:e2e` — three new steps, including one asserting the partial-application message is reported and not hidden
- Registration is a four-place list (`ALL_REGISTERED_TOOLS`, a coverage category, `TOOL_PROFILES`, the smoke's `EXPECTED_TOOLS`); the pre-push gate caught two I had missed, which is the gate working

Backend-only — no visual surface.

https://claude.ai/code/session_01Cmbb1zNhiZitB7tn9LQ5f3


kamiazya:
@coderabbitai review

PR #1084, opened 2026-08-27
feat!: collapse the dual document plane — the workspace tree becomes the address book

kamiazya:
## Summary

- **The workspace tree is the address book.** Listing, path resolution, names, pins, kinds, timestamps and branch HEAD all read from the workspace record; the `documents` table survives only as a frozen legacy inbox the boot fold reads, and the daemon works with it empty. `DualPlaneDocumentIndex` and the SQLite row index are deleted; DI binds `LoroWorkspaceDocumentIndex` (wrapped by the composition root's `CacheCoherentDocumentIndex`, which keeps the doc-cache coherent and holds the workspace write lock around every mutator).
- **Versions and branches are re-keyed on `workspaceId`** (migration 0015), with `documentId` as a plain column: migration 0016 rebuilds both tables without the FK to `documents` (a tree-created document has no row to reference). Delete-completeness moves from the cascade into the `documentTeardown` bracket's explicit cleanup, with a compensating boot-time orphan sweep for the crash window. `DocRef`'s document arm gains a required `workspaceId`.
- **Contested paths are surfaced, not auto-resolved**: two replicas merging into one path list the later entries as `shadowed` (badged in the gallery, noted in the preview — threaded end-to-end on both the browser index and the daemon HTTP/MCP surface); an agent resolving a contested path gets an error, never a silent suffix. Compaction now pins **branch tips** alongside version rows (Loro refuses a checkout before the shallow start, so a cut past a branch tip was data loss), tolerates pre-fold branch rows whose frontiers name the retired per-document oplog, and the retention policy is documented: history reaches back to the oldest version and the oldest branch tip.
- Includes the measured-change instruments (workspace-record growth scoreboard with exact byte pins; listing-convergence property), the permanent `rebuild-from-scratch` acceptance test (whole document surface works with zero `documents` rows, mutation-checked), ADR-0007 addendum, and docs sync.

## Self-review hardening (adversarial pass over this PR's own diff)

- Entry bundle: App.tsx no longer imports the concrete browser index, keeping loro-crdt's WASM off the first-paint critical path (158.9 → 118.9 KB gzip); `entry-graph-loro-free.test.ts` pins the boundary at source level.
- Lock coverage: every workspace-record mutator (index create/rename/restore, names-store rename/pin) holds `withWorkspaceWriteLock`; `workspace-write-lock-coverage.test.ts` pins each writer as blocked while the lock is held.
- Fold-skipped documents (content unreadable by this build) stay listed, resolvable and deletable through a legacy-row fallback in `FoldingBrowserIndex`, so the classified error ("saved by a newer version" vs "corrupt") reaches the screen again; the `unreadable-content` regression suite runs against the production wiring.
- The fold's pre-kind delete sweeps `versions`/`branches` explicitly (0016 dropped the cascade), and the boot fold ends with a conservative orphan-row sweep.

## Test plan

- [x] `pnpm test` passes locally (only the environment-known failures: 4 root-EACCES permission tests in `mcp-node`, 9 undici `Response(blob)` tests in `web-jsdom` — both fail on `main` in this container too)
- [x] E2E coverage added or extended where applicable (`mcp-e2e-smoke` gained the workspaceId-carrying version tools; `mcp-http-convergence-smoke` caught and now covers the tool-created-workspace registration gap)

Manual verification: real dev daemon over the folded data dir — boot fold, listing, rename round-trip, pin/unpin and branch-HEAD switch confirmed byte-level against the stored record; folded DB verified to hold zero `document:%` rows. Both e2e smokes (spawned real daemon, HTTP + WS + MCP, no mocks) green; `test:browser` green in real Chromium against real IndexedDB.

New guards worth naming: `rebuild-from-scratch.test.ts` (empty-table acceptance, mutation-checked both directions), `compact-branch-pin.test.ts` (branch tips survive compaction; foreign pre-fold tips cannot disable it), `workspace-record-growth.test.ts` (exact-pinned growth scoreboard), `workspace-tree.convergence.property.test.ts` (replica convergence), contested-path conformance (the error names the colliding path, pinned across all index implementations), `entry-graph-loro-free.test.ts` (entry closure stays loro-free), `workspace-write-lock-coverage.test.ts` (every record writer holds the lock).

## Visual evidence

Visual evidence: none — the one user-visible surface change is the `shadowed` badge on gallery cards and the preview notice, a state only reachable by merging two replicas that concurrently created the same path; it is constructed and asserted in `web-browser`/`web-jsdom` tests (`FolderContentsList` card badge, `DocumentPreview` notice) rather than staged for a screenshot. Everything else is storage/contract work with unchanged rendering, covered by the suites above.

## Checklist

- [x] Docs updated if this changes user-visible behavior or a published contract (`docs/explanation/domain-model.md`, ADR-0007 addendum, retention policy)

https://claude.ai/code/session_016W8zfiQN2orUM1hJt5ygs2

---


kamiazya:
CI status on b04ab12: `test-unit (2)` is red with two 10s **timeouts** — `backup-restore.test.ts` ("non-destructive: source survives…") and `workspaces.test.ts` ("maps a thrown CorruptStoredDataError to 500", the PUT/path case). Every other check is green; `verify` shows skipped only because it depends on the test jobs.

Evidence this is runner load, not this PR's diff:

- Both tests pass on this exact commit locally — in isolation twice and under the full parallel `pnpm test` run.
- Measured local durations are 130ms and 180ms against the 10s budget (a ~75x inflation would be needed to time out).
- No new lock-order cycle exists: all tree-index serialiser entries sit inside the workspace write lock, and the document lock is a disjoint key space (`canvas-doc:*`).
- The two timeouts hit different test files in different workers at the same moment, and the job's own import phase took 45s — the shape of a stalled shared runner, matching the load-dependent timeout family documented in `.claude/rules/integrator-flow.md`.

I attempted `rerun-failed-jobs` for run 33094912275 and got 403 (no permission from this session), so I cannot re-run the shard myself. A re-run of the failed job should settle it; if it fails again on the same tests I'll treat that as real and root-cause it.

---

PR #1125, opened 2026-08-30
refactor(web): declare the markdown editor's verbs once, and make the coverage ledger a convention

kamiazya:
## Summary

- Extracts the coverage-ledger mechanism from `editor-state.property.test.ts` into `apps/web/src/test-utils/coverage-ledger.ts`, so a second screen reuses it instead of copying the four-direction assertion.
- Writes the convention down as `.claude/rules/coverage-ledger.md` — path-scoped to `apps/web/**` and `packages/*/src/**`, so it loads when someone is editing the code it governs.
- Applies it to the **markdown editor**: its verbs become one declared table, and a property test pins them.

Follows #1119, which introduced the pattern, and #1123, which was the first time the mechanism caught a real feature addition.

Two further commits came out of CI and are described at the bottom.

## What the convention says, and why the negative half matters

The rule states when a surface earns a ledger — it grows, something models it, a missed member would be silent — and, at more length, **when it does not**: a closed vocabulary like `ThemeMode` cannot forget a member; a ledger with no model underneath has nothing to tally and decays into a hand-kept list of names; and coverage is not a percentage, so `not modelled: <reason>` is a permanent, first-class answer. What a ledger forbids is the *un-decided* member, not the uncovered one.

It also fixes an ordering the single existing example hid. **Declare the surface, model it, then pin it.** The spatial editor skipped straight to pinning only because its surfaces were already closed unions (`EditorCommand['kind']`, `GestureEvent['type']`, `ShortcutId`). Measured across `apps/web`: it is the *only* place with declared union command/event surfaces. Everywhere else the surface is undeclared, so "add a ledger" is not the first move there — declaring is.

## The markdown editor, as the worked example

Its verbs were declared **twice**. `markdownStyleKeymap` bound `Mod-b` to `wrapSelectionWith('**')` while the catalog's Bold row carried its own `'**'`; Heading and Link were catalog-only; and the separators between bands were literal positions in an array, so inserting a verb silently moved them.

`MARKDOWN_EDITOR_VERBS` is now the one declaration. The keymap derives from it, the catalog derives from it (rows, plus separators from each verb's own `band`), and the delimiters are written once. The table stays React-free so a node test can drive every verb without a renderer; icons live at the render site keyed by verb id, which buys a second exhaustiveness check for free.

Two things fell out: `SourcePaneApi.wrapSelection` had no callers left once every verb arrived as an ordinary command, so it is **deleted**; and the verbs are now a closed union, which is what a ledger can ride on.

## Test plan

- [x] New or updated `mcp-node` / `mcp-jsdom` / `mcp-browser` / `web-browser` tests added — `editor-verbs.property.test.ts` (5 properties + the ledger), `wait-for-saved.test.ts`
- [x] `pnpm test` passes locally — see the note on pre-existing failures below
- [x] Manual verification completed — the real catalog is exercised by `editor-catalog.browser.test.tsx`, which drives the menu through a real browser and passed unchanged
- [ ] E2E coverage added or extended where applicable — not applicable; no routing, websocket or persistence behaviour is involved

`editor-verbs.property.test.ts` states what the verbs guarantee rather than sampling them:

| | property |
|---|---|
| **V1** | no verb changes the line count — this is what drives every verb, so the ledger has something to tally |
| **W1** | wrapping never loses a character (a wrap is insert-only by design) |
| **H1** | setting a heading level twice is a no-op the second time |
| **H2** | promoting a body line and demoting it returns the original |
| **T1** | toggling a task checkbox twice is the identity |

**All five are mutation-checked**, each going red under a targeted mutation and naming its own property. The ledger itself is verified in **all four directions**: a new union member fails `tsc` (TS1360), a stale entry fails `tsc` (TS2353), a `covered` verb the run never drives fails with the message that says how to fix it, and a `not modelled` verb that *is* driven fails as stale (`produced it 241 times — the entry is stale`).

Two of those checks were wrong on the first attempt, in ways worth recording:

- The "covered but never driven" direction first came back **green**, because my harness applied its two edits *sequentially with a restore between* — the other property still reached `italic`. Applied simultaneously, it is red. A mutation check that does not remove every path to the subject proves nothing.
- The vacuity check first came back green for the same class of reason. Its real subject is a **newly added verb the generator cannot reach**: a synthetic verb acting only on lines starting with `@@@` is driven 24 times by V1, changes nothing, satisfies the line-count property, and is caught — *"the generator cannot reach the arrangement it acts on, so its properties are vacuous"*. A tally counts attempts; a separate effects counter is what makes that distinct.

### H1's first counterexample was the test's bug, not the code's

`{"doc":"## deeper\nweekly review","at":9}`. Measured before concluding, rather than read off the assertion:

```
1st at 9 -> line 1 "## deeper"     => "# deeper\nweekly review"    handled true
2nd at 9 -> line 2 "weekly review" => "# deeper\n# weekly review"  handled true
```

The demote shortened the document by one character, so re-applying at the same **offset** landed on the next line and promoted that instead. H1 now addresses a line, and `lineStart()` says why. `setHeadingLevel` was correct throughout.

### Pre-existing failures

`pnpm --filter @kamiazya/whiteboard-web test` reports 9 failures in 5 files — `DocumentThumb`, `VersionThumbnail`, `MergeDialog`, `VersionTimeline`, `daemon-file-adapter`. **Not this PR**, and verified rather than asserted: the identical 9 failures in the same 5 files reproduce on a clean `origin/main` worktree. They are object-URL/blob behaviour in this container. Everything this PR touches is green — the full `web-browser` project at **147 files / 829 tests**, plus `pnpm knip`, `biome`, `arch-lint` (111) and `pnpm -r typecheck`.

## The two commits CI asked for

**`867559a0` — the markdown save wait answers for the LAST edit, not any edit.** CI hit `expected '# From ' to contain '# From the list'`. The product code is correct: `scheduleSave` sets `pending` synchronously on every keystroke. What was wrong is the wait's premise, that "the editor shows the full text" implies "a save for it is pending" — the keystroke → `setBody` chain runs CodeMirror → Loro commit → doc subscription, which is asynchronous. Under load, typing outlasts the 500ms debounce, so a partial body is written first and (this being the only call site) that partial write is what first sets `data-last-saved-at`. A single-sample wait then returns describing a write that predates the last edit. #1116 closed "there is no signal"; this is one step past it — the signal did not say *which content* it covers. `waitForMarkdownSaved` moves to `test-utils` with its own test that scripts the indicator through the exact failing sequence and asserts which write it settled on; mutation-checked against the pre-fix logic.

**`f58e4acb` — the WAL snapshot test's budget, sized on a measurement.** `snapshot.test.ts` timed out at the project's 10s on a CI shard. This PR changes **zero** files under `packages/`, and that test arrived in #1118 — but the repo owner confirmed the budget exists to catch the unexpected and is fine to raise with a reason, so it is raised here rather than left. Measured: **966ms** idle, **1266ms** with the whole `mcp-node` project in parallel (4113 tests), **>10000ms** on the CI shard. Locally it never approaches the budget, so what consumed it was runner contention, not the test's own work — 3000 sequential awaited round-trips are I/O- and IPC-bound and degrade hard under CPU contention. 60s matches this package's existing precedent and still fires on a hang against ~1.3s of real work. Only the 3000-row case takes the override (the other two seed 50 and 10); verified wired and per-test by forcing it to 1ms. Batching the inserts was rejected: `c.batch` turns 3000 commits into one transaction, changing the WAL state the test exists to produce.

## Visual evidence

**Visual evidence: none — no pixels move.** The catalog renders the same rows, in the same order, with the same separators; only where the list comes from changed. That claim is not an assertion: `editor-catalog.browser.test.tsx` drives the real menu in a real browser (opens "Editing actions", clicks Bold, checks the document) and passes unchanged, as does the rest of the browser suite.

## Checklist

- [x] Docs updated if this changes user-visible behavior or a published contract — `AGENTS.md` now points at the rule instead of restating the mechanism a third time; `dev-flow.md` indexes it

## What is deliberately not here

The **source-scan** ledger variant (`editor-state-surface.test.ts`) is *not* moved into the shared helper. It has one call site and its coverage vocabulary differs — it needs a `view only:` state. Extract it when a second one exists, not before; the rule says so explicitly so the next person does not read the asymmetry as an oversight.

No ledgers were added to the other screens. Settings, workspace-files and the document list have no declared surface to pin yet, and adding one to each would be the sweep this convention is meant to make unnecessary — the rule now tells whoever next touches them what to do first.

---


kamiazya:
## CI status on `867559a0`

Two failures so far, one fixed here and one that is not this PR's.

### 1. `test-browser (2)` — fixed, on `867559a0`

`BrowserIndexPage.flow.browser.test.tsx`: `expected '# From ' to contain '# From the list'`.

Root-caused rather than re-run. **The product code is correct** — `scheduleSave` sets `pending` synchronously on every keystroke. What was wrong is the wait's premise: that "the editor shows the full text" implies "a save for it is pending". The keystroke → `setBody` chain runs CodeMirror → Loro commit → doc subscription, which is asynchronous. Under load, typing outlasts the 500ms debounce, so a partial body is written first — and since this was the only `waitForSaved` call site, that partial write is what first sets `data-last-saved-at`. A single-sample wait then sees a settled `saved` with a timestamp and returns, describing a write that predates the last edit.

#1116 closed "there is no signal for a markdown save". This is one step past it: the signal did not say *which content* it covers.

`waitForMarkdownSaved` settles, waits out a full debounce period, and settles again. It moved to `test-utils` with its own test, which scripts the indicator through the exact sequence the failing run was in (`saved@partial → pending → saved@full`) and asserts which write it settled on. Mutation-checked — with the pre-fix logic it fails with `settled on the partial write — the wait answered for a save that predates the last edit`.

Validated locally after the fix: **147 files / 829 tests green** in the full `web-browser` project.

### 2. `test-unit (2)` — not this PR's, and I cannot re-run it

`packages/mcp-server/src/server/store/db/snapshot.test.ts` → `Test timed out in 10000ms`.

Why it is not this PR's, measured rather than argued:

| evidence | |
|---|---|
| files this PR changes under `packages/` | **0** (14 files, all `apps/web/**`, `AGENTS.md`, `.claude/rules/**`) |
| where that test came from | #1118, not here |
| its cost on an idle tree | **966ms of test time, against a 10s budget** — 10× margin, twice measured |

It is the load-dependent family: `seedWalDatabase(3000)` performs 3000 sequential awaited `INSERT`s plus `VACUUM INTO` and an integrity check, so its cost scales with machine load, and this shard reported 205s of test time in an 89s wall clock. Nothing about it is broken; the budget is simply not sized for a saturated shard.

**There is no fix to port, and I am not fixing it here** — it is in a package this PR does not touch at all, and picking between "raise the budget" and "make the seeding cheap" is the owning package's call, not a drive-by in a coverage-ledger PR.

Proposed patch, for whoever picks it up: the 3000 round-trips are the whole cost, and batching them preserves what the test is actually about (commits still sitting in the `-wal`):

```ts
await c.batch(
  Array.from({ length: rows }, (_, i) => ({
    sql: 'insert into t (v) values (?)',
    args: [`row-${i}`],
  })),
)
```

I also could not spend the one allowed re-run: `rerun-failed-jobs` returns `403 Resource not accessible by integration`. Noting that here rather than treating it as someone else's problem — this PR stays watched, with a check-in scheduled, until it is green, merged or closed.

---

kamiazya:
Following up on point 2 above: the repo owner confirmed the budget exists to catch the unexpected and is fine to raise with a reason, so it is raised here in `f58e4acb` — my earlier "not fixing it here" no longer stands.

Sized on a measurement rather than on the failure:

| condition | time |
|---|---|
| idle tree (×2) | 966ms |
| whole `mcp-node` project in parallel (4113 tests, 149s of test time) | **1266ms** |
| CI shard | **>10000ms** (timed out) |

Locally it does not come close to the budget even with everything else running, so what consumed the 10s was runner contention, not the test's own work — 3000 sequential awaited round-trips are I/O- and IPC-bound, and that degrades hard under CPU contention rather than in proportion to it. 60s matches the existing precedent in this package (`document-backend-contract.test.ts`) and, against ~1.3s of real work, still fires on a hang.

Only the 3000-row case takes the override; the file's other two seed 50 and 10 rows and stay on the project default. The override is verified wired and per-test — forced to 1ms it reports `Test timed out in 1ms` and fails that case alone, the other two passing.

**I did not batch the inserts**, which was my earlier suggestion and was the worse idea: `c.batch` turns 3000 individual commits into one transaction, changing the WAL state the test exists to produce. That would have been a cheaper test of something else.

---

PR #1128, opened 2026-08-30
fix(web): tell the truth about deleting, and name a note after its own title

kamiazya:
## Summary

A dogfooding pass over document management, driven as a real user in a real browser rather than read off the source. Four frictions found, three fixed here; the fourth turned out to be a feature gap rather than a fix and is written up at the bottom instead of half-done.

## F1 — the delete dialog claimed an undo that exists

Measured, in this order:

```
DIALOG : Delete "untitled"? This permanently removes the note from this
         browser. There is no undo.
after  : Trash (1) appears
restore: "Weekly review / untitled / markdown · 6s ago" is back, content intact
```

The claim was false in **all three** places it appeared. `deleteDocument` evacuates a document into the trash *before* removing anything — its own comment says `EVACUATE FIRST` — and the daemon takes the same path, which `document-store.ts` describes as keeping "the same recoverability promise the agent-facing port makes".

The new copy says what each keeper actually does, and the two differ:

| | now says |
|---|---|
| Browser (index + document page) | The note moves to the Trash, where you can restore it. |
| Daemon (index) | …plus: its versions and branches are deleted, and restoring does not bring them back. |

That asymmetry is checked, not assumed: a browser workspace keeps no versions at all (`VersionTimeline` reads the daemon's HTTP API), while `documentTeardown` really does delete the daemon's version and branch rows, and the trash holds only the tree subtree.

**This is the worst direction for copy to be wrong in.** It tells someone a safe action is irreversible, so they keep documents they meant to tidy away. Two tests pinned the old wording — worth noting on its own: a test guarantees a claim stays *consistent*, never that it is *true*.

## F2 — a note's name never learned from its content

Typing `# Weekly review` left the workspace calling it `untitled` in the card, the URL and every search result, while the card rendered "Weekly review" right above that name from the body preview. The app already knew the title. Three notes in, the list read `untitled / untitled-2 / untitled-3`.

`seedNameFromTitle` now runs in the workspace host's save, so name and body land in one write. Two gates, and **the second is the load-bearing one**:

- `name` absent means nobody named it — but it is *also* what the rename dialog leaves behind when someone deliberately clears a name to show the path instead. Re-seeding over that would be this codebase's recurring defect: state keyed on something that has since changed.
- So the path must still be a **generated** one. Anyone who cleared a name on a document they had also placed has engaged with naming, and is left alone.

The **path is never touched**. ADR-0008 measured deriving one from a display name and found every non-Latin title collapsing to `untitled-N`; ADR-0007's addendum retracted it. Seeding only the name means a heading edit can never move a document out from under a link.

`isGeneratedDocumentPath` sits beside the generator it mirrors, and a property drives the generator and feeds every path it produces to the predicate. **That property failed immediately**: the first regex was `-[2-9]\d*`, which rejects `untitled-10` and every later one, because they begin with `1`. An example test would have had to guess that case.

## F3 — not fixed here, and not a wiring gap

`Duplicate` and `Pin` are offered in a daemon workspace and absent in a browser one, which is the default no-account experience the landing page advertises. My first read called this unwired. **That was wrong, and the correction is the point**: there is no `duplicateDocument`/`copyDocument` anywhere in `ports`, `workspace-index` or `apps/web/lib`, and no `pinOrder` in the browser index. The daemon composes duplicate at the page level from `getDocumentSnapshot` + create, against its HTTP API.

So bringing them to the browser means new operations in the shared packages, not a prop. That is a feature increment with its own product question — should a browser workspace have them at all? — and stapling it onto a copy-and-naming PR would have been the wrong shape.

## F4 — nothing on the fast path says "folder"

The New-document menu offered Canvas, Markdown note, and "Name and location…", and someone looking for a folder has no reason to open the third. There is deliberately **no "New folder" to add**: folders here *are* the path, and `pruneEmptyFolders` removes one the moment it holds nothing, so a folder created empty could not exist. The entry now says the word someone would search the menu for.

## Test plan

- [x] New or updated tests added — `title-from-body.test.ts` (8 examples + 2 properties), `new-document-path.test.ts` (+ a generator/predicate agreement property), two `web-browser` regressions for the naming seed and its gate, and the delete-dialog assertions inverted
- [x] `pnpm test` passes locally — see the pre-existing note below
- [x] Manual verification completed — every finding was *found* in a real browser and re-checked there after the fix, before any test was written
- [ ] E2E coverage added or extended where applicable — not applicable; no routing, websocket or daemon composition is involved

Verified after the fix, in the same real browser that found the problems:

```
card title after typing "# Weekly review" : ["Weekly review"]
same, for a note that opens with prose    : ["Weekly review","untitled-2"]
delete dialog                             : The note moves to the Trash, where you can restore it.
```

Gates: `web-jsdom` 209 (touched areas), `web-browser` 56 (pages), `arch-lint` 111, `pnpm knip`, `biome` over 1951 files, `pnpm -r typecheck` — all green. The naming seed is **mutation-checked**: removing the call fails the positive case alone, with the negative one still green.

**Pre-existing failures, not this PR:** `mcp-node` reports 3 failures in `migrator.test.ts` and `0011-import-fs-blobs.test.ts` — all three assert on `EACCES`/permission-denied, which cannot happen running as root in this container. Same three, unchanged, before and after this branch.

## Visual evidence

Before/after figures were composed for F1 and F2 and reviewed, but **could not be attached**: the `gh image` extension needs the `gh` CLI, which this environment does not have, and `.claude/scripts/compose-figure.mjs` needs ImageMagick, which is not installed here either — so the figures were composed with a headless-browser stand-in that keeps the part that matters (it refuses two identical panels and prints both digests; `790cc794…` vs `c3ea72a0…` for the delete pair, `72908581…` vs `83a81563…` for the naming pair).

For F1 the substance is text anyway, and it is quoted verbatim above: the old dialog said "There is no undo" and the trash restored the note. F4 is one word in a menu. F2's effect is the card title in the block above.

---

## Summary by CodeRabbit

- **New Features**
  - Untitled documents now automatically use the first Markdown heading as their title.
  - Title detection supports headings with surrounding blank lines and normalized spacing.
  - Documents deleted from the browser can be restored from Trash.

- **Bug Fixes**
  - Explicitly named documents are no longer overwritten by Markdown headings.
  - Plain-text documents remain titled “untitled” instead of receiving an invented title.
  - Delete confirmations now accurately describe restoration and permanent version removal.

### DIFF
--- a/apps/web/src/components/workspace-files/NewDocumentMenu.tsx
+++ b/apps/web/src/components/workspace-files/NewDocumentMenu.tsx
@@ -17,8 +17,10 @@ import { NewDocumentDialog } from './NewDocumentDialog.js'
  *
  * A kind is fixed for the document's life — nothing in the codebase writes
  * `kind` a second time, so the only way back from the wrong one is delete
- * (behind a "there is no undo" confirmation) and start again. An
- * irreversible choice must not be committed by a single press on an
+ * and start again, carrying whatever was written across by hand. (The delete
+ * itself is recoverable — it evacuates to the Trash — but the KIND is not:
+ * restoring returns the same document, of the same wrong kind.) A choice
+ * that cannot be taken back must not be committed by a single press on an
  * unlabeled glyph, which is what the two icon buttons this replaces did:
  * their names lived in tooltips, and a tooltip needs a pointer that a phone
  * does not have.
@@ -127,7 +129,13 @@ export function NewDocumentMenu({
               className="gap-2"
             >
               <SlidersHorizontal aria-hidden="true" className="size-4" />
-              Name and location…
+              {/* Says "folder" because that is the word someone looking for
+                  one searches the menu for, and this entry is the only place
+                  they can put a document in one. Folders here ARE the path
+                  (the dialog's Path help says so, and an emptied one is
+                  pruned), so there is no "New folder" to offer instead — a
+                  folder with nothing in it cannot exist to be created. */}
+              Name and folder…
             </DropdownMenuItem>
           </>
         )}
--- a/apps/web/src/components/workspace-files/new-document-path.test.ts
+++ b/apps/web/src/components/workspace-files/new-document-path.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it } from 'vitest'
-import { newDocumentPathIn } from './new-document-path.js'
+import { fc, fcTest, withDefaults } from '../../test-utils/fast-check.js'
+import { isGeneratedDocumentPath, newDocumentPathIn } from './new-document-path.js'
 
 describe('newDocumentPathIn', () => {
   it('creates in the folder the browser is standing in', () => {
@@ -37,4 +38,30 @@ describe('newDocumentPathIn', () => {
     expect(newDocumentPathIn('design', ['design-untitled'])).toBe('design/untitled')
     expect(newDocumentPathIn('design', ['design-system/untitled'])).toBe('design/untitled')
   })
+})
+
+describe('isGeneratedDocumentPath', () => {
+  it('recognises what newDocumentPathIn produces, at the root and in a folder', () => {
+    expect(isGeneratedDocumentPath('untitled')).toBe(true)
+    expect(isGeneratedDocumentPath('untitled-2')).toBe(true)
+    expect(isGeneratedDocumentPath('design/untitled-17')).toBe(true)
+  })
+
+  it('does not claim a path a person chose', () => {
+    expect(isGeneratedDocumentPath('weekly-review')).toBe(false)
+    expect(isGeneratedDocumentPath('untitled-notes')).toBe(false)
+    expect(isGeneratedDocumentPath('design/untitled/child')).toBe(false)
+    // `-1` and `-0` are never generated: the counter starts at 2.
+    expect(isGeneratedDocumentPath('untitled-1')).toBe(false)
+  })
+
+  // The pair has to agree, or the seeding gate opens on paths nobody generated.
+  fcTest.prop([fc.integer({ min: 0, max: 30 })], withDefaults())(
+    'every path the generator produces is recognised by the predicate',
+    (howMany) => {
+      const taken: string[] = []
+      for (let i = 0; i < howMany; i++) taken.push(newDocumentPathIn('design', taken))
+      for (const path of taken) expect(isGeneratedDocumentPath(path)).toBe(true)
+    },
+  )
 })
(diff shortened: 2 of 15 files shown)


kamiazya:
## CI found a real defect in this PR's own naming change

Worth writing down, because it is the same race the branch below this one just fixed, arriving in a second disguise.

**`test-browser (2)`** failed with a card whose title was the single letter **`F`**. Reproduced deliberately in a real browser rather than inferred:

```
type "# From"  →  pause past the 500ms debounce  →  type " the list"
before: the note is called "From", permanently
after:  the note is called "From the list"
```

Typing outlasts the debounce on a loaded machine, so a save lands while the heading is half written — and the first version of the seeding stopped there, because a name being *present* was what closed its gate. **That is worse than the `untitled` it replaced**, since a wrong name looks deliberate.

A name this code produced may now be replaced, and the test for "this one is ours" is that the title still *starts with* it — exactly the shape a half-typed heading leaves behind. Any other name is a person's and is never touched. Both halves are pinned, the counterexample as its own case before the fix, and it is mutation-checked: restoring the old rule fails the counterexample alone with the other sixteen green.

It also broke `BrowserIndexPage.flow.browser.test.tsx`, and that one is worth naming: it clicked a card through its title span, which vitest's `userEvent` locates **by text** — and the row's decorative SVG miniature renders the same words once the note carries a real name. Two matches, and a strict-mode violation that reads like the card is missing. It clicks the row now.

## And a sweep that was too narrow

`test-jsdom` then failed on two assertions pinning the *old* delete copy:

```
/removes the note, including its versions/     DaemonIndexPage.test.tsx
/removes the note from this browser/           BrowserIndexPage.test.tsx
```

My earlier sweep grepped for `no undo|cannot be undone|permanently removes` and matched neither, because both assert on a middle fragment of the sentence. Grepping for the phrase I happened to remember found four of six sites; running `pnpm --filter @kamiazya/whiteboard-web test`, the command that matches CI, found the rest. Suite is now 2901 passing with 9 failures, all the pre-existing objectURL/blob group that reproduces unchanged on a clean `origin/main` worktree.

## On the review's merge-risk note

> for a generated-path note whose name was intentionally cleared, a later save may unexpectedly restore that name

Correct, and it is a deliberate trade-off rather than an oversight — the reasoning is in `seedNameFromTitle`'s comment. `name` absent cannot be told apart from "cleared on purpose" without storing a marker saying the name was derived, which is a new stored field and a change to the rename path. The still-generated path is the proxy standing in for it, and the residual case it does not cover is: clear the name of a note you never placed, so the card shows `untitled`, then keep typing. Flagging it here for explicit awareness rather than burying it — if that case is worth closing properly, the marker is the way, and it is its own increment.

---

PR #1132, opened 2026-08-30
feat(web): show every variation's lane in the version timeline

kamiazya:
## Summary

- **The timeline shows every lane, not only the one HEAD is on.** It filtered its rows to the active branch, so the only way to see another variation's history was to switch onto it first — the one action a person consults history in order to decide.
- **That filter made a documented rule unreachable.** `mini-graph.ts` has said "rows on other branches use a ring dot" since it was written, and no call site could reach it: every row the timeline drew was active by construction, so the renderer did not even branch on `row.active` — it hard-coded a solid fill and said so in a comment. A rule with no reachable subject passes every test there is.
- **A row on another lane names its variation; a row on yours does not.** A ring answers "not the lane you are on" and stops there, and colour is not a name.

**Restore stays on HEAD's lane.** Showing another variation's history is not the same as offering to restore from it, and what restoring one variation's version *into* another means is undecided — an affordance acting on an undecided semantic is worse than none. So a row on another lane is a plain container rather than a disabled button: disabled announces "unavailable", which is the wrong story about a row that is doing its job. There is nothing to enable later, so there is nothing to grey out.

## Test plan

`pnpm check:local` → **exit 0**, with the exit code captured directly to a file rather than read through a pipe.

`pnpm --filter @kamiazya/whiteboard-web test`: **9 failed | 2925 passed**. All nine are the pre-existing jsdom `Blob`-has-no-`.stream()` family (`DocumentThumb`, `VersionThumbnail`, `MergeDialog`, `daemon-file-adapter`, and one `VersionTimeline` thumbnail case) — the same set `main` shows, unchanged in count and membership. `pnpm test` is therefore not claimed.

`web-browser` for the new file: **5/5**.

### Manual verification

The two rules jsdom answers only approximately are verified in a real browser instead of by hand, which is what the third commit is (see below): whether an SVG circle actually paints as a ring, and whether the non-restorable row is genuinely unreachable by keyboard.

## Visual evidence

**Visual evidence: none — no image upload path in this session.** GitHub access here is through the MCP server rather than the `gh` CLI, so the `gh image` step `AGENTS.md` specifies cannot run, and a figure has nowhere to be hosted.

What a figure would show, stated instead:

- **before** — one lane. Every dot solid, every row a button, and the `feature` branch's versions simply absent from the list.
- **after** — three rows for the same document: two solid dots on `main`, one ring on `feature` carrying that lane's own `BranchMeta.color` on its stroke, with `feature` named beside it. The ring row is not a button.

The browser test asserts exactly that composition (3 circles, 1 with `fill="none"` and the feature colour on `stroke`, 1 lane name, 2 row buttons), so the claim above is machine-checked rather than described from memory.

## Mutation checks

Every rule here was pinned red-first, and the two that a green-on-first-try test could have faked were mutated on the production side afterwards:

- Drawing every dot solid → the browser test fails with **0 rings where 1 is required**.
- Making every row interactive → it fails with **3 row buttons where 2 are required**.
- Labelling every row instead of only the inactive ones → the jsdom test fails with **3 lane names where 1 is required**.

Each restore was verified by re-running, not assumed.

## Four existing tests were rewritten, not adjusted until green

The rules they pinned are the rules this change moves, so adjusting them would have been the wrong move:

- the active-branch filter test (replaced),
- the lane-colour test, which asserted *the dot is always solid*,
- the empty-state copy test — the list is the document's history now, not one lane's, so "No versions on «main» yet" became "No versions yet",
- the 15s-poll test, whose evidence had to change from *which rows exist* to *which row is solid*.

Two error-path tests needed only the copy reference updated. One reference lived outside the timeline's own test file (`DaemonDocumentPage`'s history-toggle test) and went red the moment the copy changed; it is fixed in the same commit as the copy.

## The browser commit, and why the fixture asserts its own subject

The third commit exists because the repo's workflow treats jsdom as the start rather than the end. Its fixture asserts the subject is **present** — three dots, one lane name — before judging it: a two-lane scenario that silently rendered one lane would satisfy every rule in that test while checking nothing.

## Checklist

- [x] Docs updated if this changes user-visible behavior or a published contract — `apps/web/DESIGN.md`

---

## Summary by CodeRabbit

* **New Features**
  * Version history now displays entries across all branches or lanes.
  * Active-lane entries use solid markers; other lanes use hollow markers and show their lane name.
  * Restore actions are available only for entries on the active lane.
  * Entries from other lanes remain visible as non-interactive context.

* **Bug Fixes**
  * Updated empty-history messaging to reflect the complete document history.


kamiazya:
## `test-unit (2)` was red on `b827e65`, and it is not this PR's

```
FAIL  mcp-node  src/server/store/file-gc-loop-availability.test.ts
      > a file-GC pass > keeps handing the event loop back while it scans
AssertionError: expected 20 to be greater than 20
```

Established rather than assumed, on three independent grounds:

1. **This PR's diff cannot reach it.** Five files, all under `apps/web`: `DESIGN.md` and four version-timeline files. Zero files under `packages/`.
2. **The test does not exist on this branch's base.** It arrived with #1130 (`feat(server): sweep unreferenced uploads in server mode, and unblock the loop while it runs`), which merged after this branch was cut. CI tests the merge with `main`, which is how it got into this run at all.
3. **`main`'s own CI is green on the same code** — `test-unit (1)`, `test-unit (2)` and `verify` all `success` on `882cc8bf`. So this is not the "red on the base branch too" case either.

Run locally on the merged tree, **6 out of 6 passed**. The assertion is `expect(availability.samples).toBeGreaterThan(20)` and it got exactly **20** — a one-sample miss on a threshold with no margin, which is the load-sensitive-measurement shape `integrator-flow.md` catalogues.

### What I did, and what I did not

I have **no permission to re-run the job** (`rerun-failed-jobs` → `403 Resource not accessible by integration`), so the one allowed re-run was not available to me.

Instead I merged current `main` into this branch — warranted on its own, since the branch was two commits behind — which re-runs CI against the base that introduced the test. That is a real base update, not an empty commit to kick CI.

I did **not** touch the test. Making it robust would mean editing `packages/mcp-server` from a PR about `apps/web`'s version timeline, on a file that merged twenty minutes ago — widening this PR into another package and another subject. Its fragility is filed separately as follow-up: a loop-availability instrument whose pass/fail hangs on one sample will keep costing unrelated PRs a cycle, and the fix belongs with the person who set the threshold.

Gates on the merged head: `pnpm check:local` → **exit 0** (exit code captured directly). Touched suites **382/383**, the one failure being the pre-existing jsdom `Blob`-has-no-`.stream()` thumbnail case described in the PR body.

---

PR #1133, opened 2026-08-30
fix(web): declare destructive confirmation copy once, and stop it calling a note a canvas

kamiazya:
## Why this, and not something else

Picked by what actually caught a defect. In #1128 correcting the delete-dialog copy meant editing **six** places, and my grep for the old phrasing found **four**. The two it missed were tests asserting a *middle fragment* of the sentence, which no pattern written from the production sites could match — only the full suite disagreed, long after the sweep was believed done.

The cost was not a shipped bug. It was a **sweep believed complete**, and no mechanism in the repo could have told me otherwise.

## The declaration

`apps/web/src/lib/destructive-copy.ts` holds the sentence a delete dialog shows about the user's data. Scope is the **description** only — that is the line that says whether the thing comes back, the half that has been wrong, and the half a reader decides on. Titles carry no promise and have never drifted.

`DeleteDocumentDialog` now takes an **action id**, not a `description: string`:

```diff
-  description: string
+  action: DestructiveActionId
```

A caller that can pass a sentence is a caller that can write a second copy of it, which is exactly how the browser sentence came to exist in two files. It is no longer expressible.

Both index pages and `BrowserDocumentPage` build from the catalog; both test files now assert against the same builder, so they cannot drift from the copy they check.

## The scan, and the thing it had to be redesigned around

`destructive-copy-surface.test.ts` scans every source file under `apps/web/src` (`?raw` glob — apps/web is browser-only) and fails when the copy appears anywhere but the declaration.

**It scans five-word RUNS, not whole sentences, and that was a measurement, not a preference.** The first version compared whole fragments. Run against the tree it flagged the three production sites and **neither test file** — it would have repeated the #1128 miss exactly:

```
./pages/BrowserDocumentPage.tsx carries "moves to the Trash, where you can restore it."
./pages/BrowserIndexPage.tsx    carries "moves to the Trash, where you can restore it."
./pages/DaemonIndexPage.tsx     carries "moves to the Trash, where you can restore it."
```

Word runs catch the prefix case. Four words collides with ordinary English; six misses `/^The note moves to the Trash/`. Probes are **derived** from the copy by splitting on a sentinel subject — a hand-picked "distinctive phrase" beside the copy would be one more string to keep in step, the same defect one level up.

### Mutation-checked in all four directions

| mutation | result |
|---|---|
| a site hardcodes the sentence again | RED — 5 probes, names the file |
| a test asserts a prefix fragment (**the #1128 shape**) | RED — `DaemonIndexPage.test.tsx carries "moves to the Trash, where"` |
| the glob stops reaching the source | RED — `expected 0 to be greater than 300`, the count guard, not a false all-clear |
| the sentinel stops recovering static halves | RED — `expected 0 to be greater than 0` |

## Deliberately NO coverage ledger

Per `.claude/rules/coverage-ledger.md`, a ledger is step 3 of declare → model → pin, and **nothing models confirmation copy** — there is no property or table-driven run to tally, so a ledger here would be the hand-maintained list of names that rule exists to replace. Declaration + scan is the complete answer, not a partial one.

The rule gains that as its worked example for a surface that earns a scan and not a ledger, plus the word-run measurement, plus why two source scans still do **not** share a helper: they have four lines in common (glob + count assert) and diverge entirely past that. *Extract when two call sites want the same judgement, not when two of them read files.*

It also now states the scan's **floor** honestly: a three-word prefix assertion is under any run length that avoids false positives. That is acceptable because a single definition removes the *sweep* rather than perfecting the *grep*.

## Second fix — found by writing the declaration down

`triggerCleanup`'s two failure messages hardcoded `"canvas"`. Deleting a **note** that could not be flushed reported:

> The canvas could not be safely removed. Your copy has been kept.

`kindNoun` exists precisely so user-visible copy never calls a note a canvas (`vocabulary.md` retired the container sense), and these two strings predate it. Red test first — it reproduced that exact string — then routed through `kindNoun`.

## Third fix — CI, and not this PR's subject at all

This PR's `test-unit` went red twice on a defect it cannot reach, in `packages/mcp-server`. Both failures were the **same class**, one file apart:

```
file-gc-loop-availability.test.ts        expected 20 to be greater than 20
workspace-tail-loop-availability.test.ts expected  5 to be greater than  5
```

Both floored an absolute **sample count** — calibrated on an idle machine — beside a stall assertion that is deliberately a ratio. A count is a statement about how busy the runner is. Base `882cc8bf` was green and `rerun-failed-jobs` answers **403** here, so it was reproduced rather than waited out: eight CPU hogs on four cores, and the *first* local run gave `samples=20, elapsed=427ms` — the identical message.

Two instances is the class, so the rule moved into the helper that owns the measurement as `loopTurnShare`, and both tests assert against it:

| | file-GC | workspace-tail |
|---|---|---|
| yields removed (the defect each guards) | 3 of 279 = **0.011** | **0 samples** |
| under load | 0.18 – 0.24 | 0.37 – 0.64 |
| idle | 0.35 | 0.55 |

Floor at **0.05** — 4× above the blocked case, 5× below the worst loaded one. Mutation-checked separately against each production yield removed: `expected 0.0116 to be greater than 0.05` and `expected 0 to be greater than 0.05`. The message is now the share itself rather than arithmetic.

`loop-availability.test.ts`'s own `samples > 10` is **left alone**: its body is a bare `setTimeout` with no CPU work and it passed three loaded runs, so there is no evidence to weaken it on.

## Test plan

- [x] New or updated tests added — the scan (5 cases), a red-first regression for the wrong noun, and the two re-floored loop-availability guards
- [x] `pnpm --filter @kamiazya/whiteboard-web test` — **2932 passing**, `web-node` 77/77. The 9 failures are the standing objectURL/blob group, **A/B-confirmed** identical on a clean `origin/main` (`882cc8bf`) worktree in this container; none is in a file this diff touches
- [x] `arch-lint-node` 111/111 · `mcp-node` 4168 passing (its 3 failures are the chmod/EACCES tests that cannot fail as root, pre-existing) · both loop-availability guards pass four runs under load
- [x] Manual verification in a real browser against `pnpm dev` — all three delete dialogs after the refactor: document page/note *"Delete this note?" / "The note moves to the Trash, where you can restore it."*, document page/canvas *"Delete this canvas?" / "The canvas …"*, index page *"Delete \"untitled\"?" / "The note …"*. (The daemon variant needs a daemon; it is covered by the jsdom test.)
- [x] `pnpm knip`, `biome`, `secretlint`, `intent:validate`, `pnpm audit --prod`, `pnpm -r typecheck` green. `check:local`'s only failure is `compose-figure.test.mjs`, which prints its own reason — ImageMagick is absent in this container
- [ ] E2E coverage — not applicable; no routing, websocket or daemon composition involved

## Visual evidence

**Visual evidence: none — the rendered strings are unchanged for a canvas, and the one string that does change is a failure message reachable only by making a save fail mid-delete.** The delete dialogs render byte-identical copy before and after; the change is where that copy comes from, which the browser walk above confirms. The note-vs-canvas fix has no staged capture worth trusting either — the honest evidence is the red test, which reproduced `The canvas could not be safely removed.` on a `markdown` document before the fix and fails on any reintroduction.

---

### DIFF
--- a/.claude/rules/coverage-ledger.md
+++ b/.claude/rules/coverage-ledger.md
@@ -115,9 +115,43 @@ Assert the scan found a plausible COUNT in its own `it`. A regex that stops
 matching otherwise reports itself as "every entry is stale", which sends the
 reader to the wrong file entirely.
 
-This variant is deliberately NOT in the shared helper: one call site, and
-its coverage vocabulary differs (it needs a `view only:` state). Extract it
-when a second one exists, not before.
+There are two scans now, and they still do NOT share a helper. What they
+have in common is four lines — the `?raw` glob and the plausible-count
+assertion — and past that they diverge completely: `editor-state-surface`
+CLASSIFIES each thing it finds against a three-value vocabulary it needs a
+`view only:` state for, while `destructive-copy-surface` asserts a
+single-definition rule and has no vocabulary at all. A helper over that is
+a glob with a parameter. Extract when two call sites want the same
+JUDGEMENT, not when two of them read files.
+
+### The second one: copy declared once
+
+`lib/destructive-copy.ts` + `destructive-copy-surface.test.ts` is the worked
+example for a surface that earns a scan and explicitly does NOT earn a
+ledger. Nothing models confirmation copy — there is no property or
+table-driven run to tally — so a ledger would be the hand-maintained list of
+names this rule exists to replace. Step 1 of declare -> model -> pin, plus a
+scan holding it, is the whole mechanism, and that is a complete answer
+rather than a partial one.
+
+Two things it measured that generalise:
+
+- **Scan word RUNS, not whole strings.** The defect that motivated it was a
+  correction that had to reach six places and a grep that found four; both
+  misses were tests asserting a MIDDLE FRAGMENT of the sentence. Measured:
+  the first version of that scan compared whole fragments and flagged the
+  three production sites and neither test. Five-word runs, derived from the
+  copy by splitting on a sentinel subject, catch both. Four collides with
+  ordinary English; six misses the prefix case.
+- **Derive the probes, never list them.** A hand-picked "distinctive
+  phrase" beside the copy is one more string to keep in step — the same
+  defect one level up.
+
+A scan has a floor and saying so is part of the rule: a test asserting a
+three-word prefix is under any run length that avoids false positives. That
+is acceptable here because the single definition removes the SWEEP rather
+than perfecting the grep — with one source there is nothing to keep in
+step, and a stale assertion fails loudly instead of being missed quietly.
 
 ## Traps
 
--- a/apps/web/src/components/document-list/DeleteDocumentDialog.tsx
+++ b/apps/web/src/components/document-list/DeleteDocumentDialog.tsx
@@ -1,3 +1,4 @@
+import type { DocumentKind } from '@kamiazya/whiteboard-model'
 import {
   AlertDialog,
   AlertDialogCancel,
@@ -8,13 +9,21 @@ import {
   AlertDialogTitle,
 } from '@/components/ui/alert-dialog'
 import { Button } from '@/components/ui/button'
+import { DESTRUCTIVE_COPY, type DestructiveActionId } from '@/lib/destructive-copy'
+import { kindNoun } from '@/lib/kind-noun'
 
 export interface DeleteDocumentDialogProps {
-  // The canvas pending deletion, or null when the dialog is closed.
-  pending: { displayName: string } | null
+  // The document pending deletion, or null when the dialog is closed.
+  pending: { displayName: string; kind?: DocumentKind } | null
   busy: boolean
   error: string | null
-  description: string
+  /**
+   * Which promise this delete makes about the user's data. The dialog builds
+   * the sentence itself rather than taking one: a caller that can pass a
+   * string is a caller that can write a second copy of it, which is how the
+   * browser sentence came to exist in two files. See lib/destructive-copy.ts.
+   */
+  action: DestructiveActionId
   onCancel: () => void
   onConfirm: () => void
 }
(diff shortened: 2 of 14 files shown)


kamiazya:
`test-unit (2)` went red on `0ca2d78b`, in code this PR cannot reach:

```
mcp-node src/server/store/file-gc-loop-availability.test.ts
  > a file-GC pass > keeps handing the event loop back while it scans
AssertionError: expected 20 to be greater than 20
```

That file arrived two commits ago in #1130 (`bd19b9d8`); this diff is `apps/web` plus one `.claude` rule. Base `882cc8bf` was green, so "red on the base too" did not settle it, and `rerun-failed-jobs` answers **403** here — no re-run available.

So I reproduced it instead of waiting it out. Eight CPU hogs on four cores, and the **first** local run gave the identical message:

```
[probe] {"elapsedMs":427,"samples":20,"idealSamples":85,"worstStallMs":41}
AssertionError: expected 20 to be greater than 20
```

**The threshold was an absolute sample count in a test that is otherwise scale-free** — its own comment says why the stall assertion beside it is a ratio ("stable across machines in a way 'the worst stall is 38.8ms' is not"), and this one assertion was the exception. A count measures how busy the runner is.

| regime | samples / ideal | |
|---|---|---|
| yields removed (the defect it guards) | 3 / 279 | **1.1%** |
| under load | 20 / 85, 21 / 117 | 18–24% |
| idle | 61 / 173 | 35% |

Fixed in `0cf087db` as a floor at **5% of the samples the interval asked for** — 4× above the blocked case, 5× below the worst loaded one. The old `20` had *no* margin; a loaded run landed on exactly 20, which is what CI hit.

Mutation-checked: with the three `yieldToLoop()` calls removed the new assertion fails `expected 2 to be greater than 11.233`, so it still catches the defect it exists for. Four loaded runs pass with it in place, and full `mcp-node` is 4168 passing.

Flagging it because it is not this PR's subject: the same zero-margin threshold would have hit whichever PR ran next.

---

PR #1141, opened 2026-08-30
ci(web): gate an absolute LCP floor, deliberately far from the measurement

kamiazya:
## Summary

- `smoke-bundle-size.mjs` says the byte budget's four raises (108 → 114 → 126 → 138 KB) are "how a budget stops meaning anything", and that the real question is a **product call the gate file should not make silently**. This makes that call for the metric half.
- **LCP floor: 1000 ms**, against a current median of **492 ms**. The distance is the design, not slack.
- The instrument keeps its character — it asserts nothing unless a caller supplies a floor, so `measure:critical-path` still reports and fails nothing.

## Why the floor sits twice as far away as the measurement

On this rig LCP is dominated by CDP-emulated transfer and responds linearly at **4.9 ms per gzipped KB** — so it is very nearly a function of bytes. Placed near the measurement it becomes a second byte budget:

| | `smoke:bundle-size` | LCP on this rig |
|---|---|---|
| what it sees | transferred bytes | very nearly transferred bytes |
| resolution | **335 bytes** | ~2.5 KB (noise band ±2% ≈ ±10 ms) |
| variance | none (deterministic) | 2%, occasionally +11% cold |

Seven times coarser and less certain, for the same question, is not a gate worth having. So the two get **different jobs rather than different numbers**: bytes detect change, this states an absolute floor — *a mid-range phone sees the shell inside a second, whatever the code does*. A regression small enough to matter reaches the byte budget first, by design.

## What would make LCP independent, and why it is not available

The CPU-bound half — the parse/execute cost bytes cannot see. That quantity is `longTasks`, and it measured a **14–70% spread** across two 10-run sets. Signal and noise are the same size, so it gates nothing. Getting it honestly needs a steadier measurement than observed throttling (Lighthouse's lantern model, or field data), which is separate work and not a threshold choice.

## The mount check is part of the gate, and the numbers say why

Replacing the entry chunk with a no-op, so the app never mounts and only the boot splash paints:

```
run 1/3  lcp=460ms  mounted=false rootKids=1 text=""
```

**460 ms — faster than the working 492 ms.** A floor on its own would read a completely broken build as an improvement and pass. This is the one way this gate can be satisfied while measuring nothing, so it is refused explicitly.

Both halves of the check are needed and only one fires: that same run still reported `rootChildren=1`, because the splash is itself a child — so the child count alone would have missed it, and `shellMark` is what refuses it.

## Why the median of several runs, never one

Two 10-run sets on one build both gave a **492 ms median**, but the first set's opening run read **548 ms (+11%)** and the second set had no outlier at all. A cold-start effect that appears intermittently would flake a single-shot gate for reasons that have nothing to do with the code. The gate runs 5 and takes the median: one outlier cannot move it.

## Verification

| case | result |
|---|---|
| floor 1000 (as CI runs it) | **OK** — median 492 ms |
| floor 400 | **FAILED** — `496ms is over the 400ms floor` |
| entry chunk replaced with a no-op | **FAILED** — `3/3 runs never mounted the app` |
| restored | **OK** again — confirmed by re-running, not assumed |

`pnpm check:local` → **exit 0**. `local-gate-command.test.ts` still passes — it guards the `check` job, and this step joins `verify` alongside the other web smokes, so it is placed consistently with `smoke:bundle-size` and `smoke:pwa-precache`.

## What this does NOT fix

**The 108 → 114 → 126 → 138 ratchet.** That is the byte budget being set at "current + margin", and adding a metric gate does not change it. Fixing it means deciding what critical-path size is right as a product question — a separate decision from this one, and deliberately not smuggled in here.

## Test plan

- [ ] New or updated `mcp-node` / `mcp-jsdom` / `mcp-browser` / `web-browser` tests added
- [x] Manual verification completed (describe above)

No unit test: the subject is a CI gate script, and its four cases above are exercised by running it, which is what a test would do less directly. `pnpm test` is not claimed, for the pre-existing `apps/web` failures recorded in #1136.

## Visual evidence

**Visual evidence: none — this adds a CI step and changes nothing a human looks at.** The evidence is the measurement tables above.

## Checklist

- [x] Docs updated if this changes user-visible behavior or a published contract — the decision, its measurements and its limits live in the script's own docblock, where the next person to question the number will be

---

## Summary by CodeRabbit

- **Performance**
  - Added automated checks to verify the web app loads successfully and maintains a fast Largest Contentful Paint (LCP).
  - Performance checks evaluate multiple runs to provide more reliable results and flag regressions before release.

- **Reliability**
  - Added validation to detect cases where the application fails to mount during loading checks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -300,6 +300,13 @@ jobs:
       - name: Web app PWA precache gate
         run: pnpm --filter @kamiazya/whiteboard-web smoke:pwa-precache
 
+      # An ABSOLUTE floor, not a regression stop — the bundle-size gate above
+      # is the regression stop and is seven times finer. This one only refuses
+      # a shell that takes over a second to appear on a mid-range phone
+      # profile; the reasoning and the measurements are in the script.
+      - name: Web app LCP floor
+        run: pnpm --filter @kamiazya/whiteboard-web smoke:lcp-floor
+
       - name: Web app preview-origin behavioral smoke
         run: pnpm --filter @kamiazya/whiteboard-web smoke:preview-origin
 
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -18,6 +18,7 @@
     "smoke:artifact": "node scripts/smoke-artifact.mjs",
     "smoke:bundle-size": "node scripts/smoke-bundle-size.mjs",
     "measure:critical-path": "node scripts/measure-critical-path.mjs",
+    "smoke:lcp-floor": "LCP_FLOOR_MS=1000 MEASURE_RUNS=5 node scripts/measure-critical-path.mjs",
     "smoke:preview-origin": "node scripts/smoke-preview-origin.mjs",
     "smoke:pwa-precache": "node scripts/check-pwa-precache.mjs"
   },
--- a/apps/web/scripts/measure-critical-path.mjs
+++ b/apps/web/scripts/measure-critical-path.mjs
@@ -228,7 +228,18 @@ async function main() {
   }
   const { server, port } = await serveDist()
   const url = `http://127.0.0.1:${port}/`
-  const browser = await chromium.launch()
+  // `WHITEBOARD_CHROME_PATH` when the environment supplies its own Chrome,
+  // exactly as `smoke-preview-origin.mjs` does. CI's `verify` job installs no
+  // Playwright browsers — it sets that variable to the system Chrome — so a
+  // bare launch there fails with "Executable doesn't exist", which is a
+  // missing browser and not a slow one. Measured: that is how this gate first
+  // failed on CI, having passed every local run.
+  const browser = await chromium.launch({
+    headless: true,
+    ...(process.env.WHITEBOARD_CHROME_PATH && {
+      executablePath: process.env.WHITEBOARD_CHROME_PATH,
+    }),
+  })
   const runs = []
   try {
     for (let i = 0; i < RUNS; i++) {
--- a/apps/web/scripts/measure-critical-path.mjs
+++ b/apps/web/scripts/measure-critical-path.mjs
@@ -107,7 +107,30 @@ const RUNS = Number(process.env.MEASURE_RUNS ?? 10)
 // Set by the GATE caller only. Absent, this script reports and fails nothing,
 // which is what it was built to be; the assertion belongs to whoever states a
 // floor, not to the instrument.
-const FLOOR_MS = process.env.LCP_FLOOR_MS === undefined ? null : Number(process.env.LCP_FLOOR_MS)
+//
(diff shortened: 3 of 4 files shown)


kamiazya:
@coderabbitai review

---

kamiazya:
CodeRabbit's merge-risk finding was valid against my own code, and is fixed in `e4c180ea`.

## The defect, reproduced

`LCP_FLOOR_MS=abc` gave `Number('abc') === NaN`; `NaN === null` is false so the gate armed, and `median > NaN` is false so every run passed. Measured on the unfixed script:

```
LCP floor: OK — median 496ms, floor NaNms
exit=0
```

A typo in the env var would have left this job green over a gate that was no longer checking anything — the same guard-that-never-reaches-its-subject shape this PR's own mount check exists to refuse, one level up in the gate's own configuration. That makes it mine to fix here rather than defer.

## The fix

The floor is now parsed by a pure `parseFloorMs`, and a present-but-not-positive-finite value is refused **before** any build or browser work, so a misconfigured gate costs 0.48s instead of a full measurement run:

```
LCP_FLOOR_MS must be a positive number of milliseconds; got "1000ms". Refusing to run,
because an unparseable floor would arm the gate and pass everything.
exit=1
```

`main()` moved behind an `import.meta.url` check so the module is importable by its test, exactly as its sibling `smoke-bundle-size.mjs` already does.

## Why that guard needed pinning too

Breaking the `import.meta.url` expression makes the CI step measure nothing and exit **0** — a green job over a gate that no longer exists, which is strictly worse than the bug being fixed. So the new `measure-critical-path.test.ts` includes a subprocess case asserting both the refusal message and a non-zero exit; a guard that stops matching fails there instead of going quietly green. Asserting the *message* rather than just the exit code is load-bearing, because a missing `dist/` also exits 1.

## Mutation check (both directions, restore verified by re-running)

| mutation | result |
|---|---|
| revert the validation to bare `Number()` | **7 of 9 failed** |
| break the `import.meta.url` expression | **1 failed** (direct run prints nothing, exits 0) |
| restored (md5 confirmed identical to pre-mutation) | **9 passed**, 1.19s |

## Gates

`web-node` 86 passed · `apps/web` typecheck clean · `biome check` clean · `pnpm knip` clean. The real CI step `pnpm --filter @kamiazya/whiteboard-web smoke:lcp-floor` re-verified end to end after the refactor: `LCP floor: OK — median 496ms, floor 1000ms`. The gate still fails on a real breach (`LCP_FLOOR_MS=100` → exit 1), which is what proves the refactor did not defang it.

The one remaining pre-merge warning is Docstring Coverage 0% (1 function analysed). Declining it on the same grounds as previous PRs: comment discipline here is governed by AGENTS.md's Source Comment Discipline — enduring *why* over per-function docstrings — and `parseFloorMs` carries a 12-line rationale block already.

---

PR #1144, opened 2026-08-30
fix(web): stop a keystroke during a document load being written into the document that left

kamiazya:
## The defect

`use-markdown-document`'s load effect is careful in three ways — it guards its result with `cancelled`, flushes the outgoing debounce on cleanup, and unsubscribes. What it did **not** do is clear what it *holds* before the async load: `doc`, the body and `hostRef` were reset only when `documentId` became `null`, never when it changed from one document to another.

So for the whole of the next document's load, the hook still held the previous one's **write handle** and content under the new one's address:

- `setBody` writes through `hostRef.current`
- `scheduleSave` keys its scheduler on the **new** id

which makes a keystroke in that window a write into the previous document, queued under a name that is not its own.

Reproduced before touching anything, both halves:

```
expected 'written in c1' not to be 'written in c1'    ← c1's prose rendered under c2
expected [ { id: 'c1', … } ] to deeply equal []       ← save('c1', …)
```

The second is the severe one: **text typed under one document was saved into another.** Unlike #1140's stray delete there is no trash to undo it — it is a content write.

## The fix, and the rule it restores

Clear synchronously *before* the load. Both sibling screens already state this rule in their own comments:

> `DaemonIndexPage`: "Clear synchronously BEFORE the async load: leaving the previous workspace's rows visible during the switch lets a click pair the new workspace id with an old workspace's path — a mismatched identity."

> `VersionTimeline`: "Clear the previously loaded canvas's versions immediately on canvas change so a stale row … never renders under the new canvas while the refetch is in flight."

The `cancelled` guard stops a stale load from **landing**. It says nothing about what is still held while the next one is **in flight**, and that gap is the whole defect.

## The ledger gains refs — because a ref held this one

`scoped-screen-state.test.ts` scanned only `useState`. `hostRef` is a `useRef`, so the guard from #1143 would never have asked about it. The hook is added as a fifth case with ref scanning on.

Two things that made that guard wrong for this file, **both surfaced by it failing rather than by me reading it**:

| what | why it was wrong |
|---|---|
| it derived the setter as `set${Name}` | does not hold here — `[body, setBodyState]`, because `setBody` is the hook's *public writer*. The setter now comes from the declaration, and a ref's reset is recognised as an assignment to `.current` |
| its "unclassified" message still said *workspace-switch* | stale since #1143 gave the ledger a second scope |

The four screen cases keep ref scanning **off**, and the reason is on the function: 3–5 refs each, none read closely enough here to classify honestly, and a wrong `no subject:` in a guard people trust is worse than an absent one. Turning it on is a one-word change per case and its own increment.

### Mutations

| mutation | result |
|---|---|
| a new **ref** on the hook, unclassified | RED — `expected [ 'strayRef' ] to deeply equal []` |
| `hostRef` no longer cleared in the effect | RED — `expected [ 'hostRef' ] to deeply equal []` |
| a `survives:` entry downgraded to a bare `cleared on switch` | RED — names `saveState` |
| the fix reverted (clearing made conditional again) | RED — both reproductions, with their own messages |

Each was verified present in the file before its result was believed.

## One thing recorded rather than fixed

`saveState` is classified `survives:`, not quietly cleared. It **is** scope-bound — it describes the departed document's last save — but a reset alone would not settle the question: the outgoing flush reports asynchronously through the same setter, so its result lands *after* any reset. That needs its own reproduction, the way the write path got one, and the entry says so.

That is the third value in the classification, added here:

```ts
type ScopeCoverage = 'cleared on switch' | `no subject: ${string}` | `survives: ${string}`
```

A first-class permanent answer, like `not modelled:` elsewhere: what the ledger forbids is the **undecided** member, not the uncleared one.

## Test plan

- [x] Two red-first reproductions, then mutation-checked by reverting the fix
- [x] `pnpm --filter @kamiazya/whiteboard-web test` — 289 files / **2981** + `web-node` 13 / 86, zero failures
- [x] `pnpm test --project web-browser` — 148 files / **844**, zero failures
- [x] `pnpm check:local` — **exit 0**
- [x] All four mutations above
- [ ] E2E — not applicable; the switch is a prop change the hook's own suite models

## Visual evidence

**Visual evidence: none — the visible half is a brief empty editor where there used to be the wrong document's prose.** A screenshot of the fix is an editor with nothing in it for the length of a load, which is indistinguishable from a slow load. The evidence is the two assertions above; the half that matters was never visible at all, since a save into another document leaves no mark on the screen you are looking at.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Switching documents now immediately clears stale content and editing state.
  - Disabled document loading and missing-document states reset cleanly without displaying data from a previously selected document.
  - Prevented keystrokes made while a new document is loading from being saved to the previously active document.

- **Tests**
  - Added regression coverage for document switching during loading.
  - Expanded state-reset checks to verify document-related state is cleared correctly when changing scope.

### DIFF
--- a/apps/web/src/components/settings/PromoteWorkspaceSection.browser.test.tsx
+++ b/apps/web/src/components/settings/PromoteWorkspaceSection.browser.test.tsx
@@ -38,6 +38,8 @@ const DAEMON = { baseUrl: BASE, token: 'tok-1' }
 interface StubOptions {
   updateDelayMs?: number
   putDelayMs?: number
+  /** An update that stays in flight until the test releases it. */
+  updateGate?: Promise<void>
   failUpdateStatus?: number
   workspaces?: { workspaceId: string; segment?: string; displayName?: string }[]
 }
@@ -53,6 +55,7 @@ function daemonStub(target: LoroDoc, opts: StubOptions = {}): typeof globalThis.
     }
     if (url.endsWith('/workspace-document/update') && init?.method === 'POST') {
       if (opts.updateDelayMs) await new Promise((r) => setTimeout(r, opts.updateDelayMs))
+      if (opts.updateGate) await opts.updateGate
       if (opts.failUpdateStatus) {
         return Response.json(
           { title: `Workspace "ws-a" not found` },
@@ -137,6 +140,28 @@ async function seedPreFoldDocument(path: string): Promise<string> {
   doc.commit()
   await new LoroStore().save(entry.documentId, doc.export({ mode: 'snapshot' }))
   return entry.documentId
+}
+
+/**
+ * An update the test releases, rather than one that finishes on a timer.
+ *
+ * A test that has to observe the RUNNING phase is racing the flow it started:
+ * the progress element is on screen only until the update resolves, and the
+ * driver round trip that `userEvent.click` waits out afterwards is charged to
+ * that same window. A delay makes the window wide, not certain — and it is a
+ * wall-clock number sitting next to a cost that grows with the run (the same
+ * file's tests were measured at 1.5s alone and 30s+ with the whole browser
+ * project in flight). Held instead, nothing about the machine can close the
+ * window early.
+ */
+function heldUpdate(): { gate: Promise<void>; release: () => void } {
+  let release = (): void => {}
+  const gate = new Promise<void>((resolve) => {
+    release = () => {
+      resolve()
+    }
+  })
+  return { gate, release: () => release() }
 }
 
 const NO_INTERNAL_VOCABULARY = /loro|crdt|oplog|snapshot/i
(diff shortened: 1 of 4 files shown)


kamiazya:
### A fourth commit that is not part of the fix

`b51e59a1` touches `PromoteWorkspaceSection.browser.test.tsx`, which nothing else in this PR reaches. `test-browser (1)` went red on `139b7fc9` in that file, and this is what was done about it.

**Not established as this branch's**, three ways:

- the diff is `use-markdown-document.ts`, its new switch-race test, and `scoped-screen-state.test.ts` — none of them is imported by that file, and the two new tests are jsdom, not browser;
- `test-browser (1)` is green on `main` (`7c88a535`);
- the whole `web-browser` project passes locally on this head — 148 files / 844 tests — and the failure did not reproduce.

Re-running a job is not available to this session, so per the CI rules the flaky case was made robust rather than waved through.

**What the log actually showed.** The test asked for `promote-progress` and got a DOM already carrying `promote-last-result` ("Moved 2 documents to daemon workspace \"ws-a\"") — the flow had finished before the assertion looked. The window it was reading was a wall-clock `updateDelayMs: 120`, and the driver round trip `userEvent.click` waits out *after* dispatching the click is charged to that same window. A delay makes the window wide, not certain, and it is a fixed number sitting next to a cost that grows with how much of the suite is in flight.

So the update is now **held by the test** and released once the running phase has been read.

| check | result |
|---|---|
| flow still mid-run 2000ms after the click, with the gate in place | held — 16x the old window |
| `release()` moved before the read | RED, naming the new must-be-absent assertion |
| file in isolation | 11/11 |

The second row is the one that matters: it is what stops the added `promote-last-result` must-be-absent assertion from being decoration.

One thing not claimed: **the race itself does not reproduce on this machine**, so the "delete the `await` and watch it go red" mutation stays green here. What was measured is that the window no longer closes on the machine's schedule — not that the original failure was reproduced and fixed.

---

PR #1163, opened 2026-08-31
test(rules): pin the always-on context budget, and cut 15% of it

kamiazya:
## Summary

- Every session in this repo pays for the always-on rules before any work starts. Nothing counted them: **105,930 characters, about 26,500 tokens**, across `AGENTS.md` and the four `.claude/rules/` files with no `paths:` frontmatter.
- Adds the instrument that counts them, **in its own commit, before any cut** — the effect of this work is invisible in a diff and its failure mode is silent.
- Then cuts three sections whose detail already lives in a skill or a contributor doc, verified claim by claim first. **Always-on corpus 106,203 → 90,570 characters, roughly 3,900 tokens off every session.**

### Why an instrument first

The cheap reachability probe `measured-change` asks for, run before deciding anything: **how much of the always-on corpus would any test notice losing?** Measured — **0.15%**. The only guarded content is the 163 characters `dev-rules-contract.test.ts` pins about the web-jsdom hazard. The other 99.85% can be deleted and the suite stays green, so "the tests will catch a bad cut" is not available here and the safety has to be built.

`dev-rules-budget.test.ts` derives the always-on set from `paths:` frontmatter rather than a hardcoded list, so a rule file that gains or loses path scoping moves between budgets and has to say so in the diff. `package-canvas-render.md` gets its own line: at 77,606 characters it is thirteen times the median package rule, and a total would hide it.

Sizes are pinned in **1000-character buckets**, not exactly. An exact pin is right for a list of modules and wrong for prose — a typo fix would fail the test, and a test that fails on typos gets weakened until it means nothing. A bucket keeps the property that matters (a cut is exactly as loud as growth) while an ordinary clarification passes.

### What was cut, and how it was checked

Every cut was preceded by comparing the always-on text against its destination **claim by claim**, and anything the destination lacked was moved there rather than dropped.

| section | before | after | destination |
|---|---:|---:|---|
| `AGENTS.md` Test Layer Selection | 4,465 | 1,062 | `test-layer-selection` skill |
| `AGENTS.md` MCP Development Mode | 5,407 | 1,188 | `mcp-debugging.md`, `development.md` |
| `AGENTS.md` PR Visual Evidence | 2,797 | 840 | `visual-evidence` skill |
| `AGENTS.md` Doc Screenshots | 1,739 | 386 | `docs-sync` skill |
| `dev-flow.md` Disciplines | 9,042 | 4,340 | the three skills it already pointed at |

Moved rather than lost: the vacuous-generator rule and the independent-oracle rule (into `test-layer-selection`); "a guard that never reaches its subject" with all three worked instances (same skill — it is the third sibling of those two, and they are worth reading together); and why `docs:snapshots` is excluded from `pnpm test` and generates twice (into `docs-sync`).

Two paragraphs were dropped rather than moved, both announcements of `coverage-ledger.md` — which is path-scoped to the files where the decision arises and loads itself, so the always-on copy was a second announcement of a rule that already arrives on its own.

### Two corrections found on the way

- **`measured-change` was stale.** It listed an aligned re-score post-pass among the changes the instrument REJECTED, using the first shape's verdict as if it were final — while the third shape is in `spatial-edges.ts` today, running the search a second time with `align: true`. Verified three ways: the code, `package-canvas-render.md`'s record of the two discarded shapes at 13x and 4.5x, and `dev-flow.md`'s record of the 2x that shipped. The skill now says an instrument prices rather than vetoes, which is both more useful and true.
- **The instrument's first finding was its own commit.** It was written and verified green on one base, then rebased onto a newer `origin/main` where #1149 had added 273 characters to `vocabulary.md` — so its total pin said 105 where the tree said 106. The pre-push gate would have caught it; it was skipped with `LEFTHOOK=0` while working around an unrelated push failure. Corrected here.

## Test plan

- [x] New or updated tests added (`mcp-node`)

The budget instrument was mutation-checked five ways before any cut relied on it:

| mutation | result |
|---|---|
| +1000 chars on an always-on file | per-file and total fail |
| a new always-on rule file | file set fails |
| removing `paths:` from a scoped file | file set fails |
| **−1000 chars** on `package-canvas-render.md` | only its own line fails |
| a 20-character edit | nothing moves |

The fourth is the one that matters: a cut is as loud as growth. The fifth is what makes the bucket unit honest rather than a dodge.

Verified on **Node 24** (`.node-version`; this container defaults to 22, which `local-node-version.test.ts` catches and names): **365 files, 4,202 passed, 9 skipped, 0 failed**, matching what `dev-flow.md` records CI reporting. `pnpm lint`, `pnpm knip`, `pnpm secretlint` and `pnpm typecheck` are clean.

### A method failure worth recording

Checking "does the skill already say this?" with `grep <phrase>` produced **four false ABSENTs** in this work — `diagnosis-evidence`'s own "read the stack line" section, `docs-sync`'s "pin the clock", and `visual-evidence`'s compose-figure invocation and its `gh image` line. Each claim was there in different words. The direction is what makes it dangerous: a false absence licenses writing the content again, or deleting the copy that had it. Recorded in `diagnosis-evidence` — a negative from a text search is a hypothesis, only a positive is safe.

## Visual evidence

Visual evidence: none — the diff is rule prose, skill prose and one guard test. It moves no pixels and changes no rendered surface.

## Checklist

- [x] Docs updated if this changes user-visible behavior or a published contract — no behaviour changes; `docs/contributing/` is unchanged and is now the single home for the MCP dev-mode detail

---

## Summary by CodeRabbit

* **Documentation**
  * Streamlined development guidance and moved detailed workflows into focused skill documents.
  * Added clearer guidance for validating diagnoses, selecting test layers, and avoiding vacuous test results.
  * Documented measured optimization trade-offs and improved evidence-based review practices.
  * Clarified documentation snapshot checks, cross-platform differences, and local regeneration steps.
  * Added coverage-ledger guidance and expanded examples of common testing pitfalls.

* **Tests**
  * Added safeguards to track the size of always-on development guidance and prevent unexpected growth.

### DIFF
--- a/.claude/rules/dev-flow.md
+++ b/.claude/rules/dev-flow.md
@@ -78,58 +78,25 @@ TDD red-first; Zod single source of truth (`z.infer`, never a parallel hand-writ
 
 **Simplicity already has two executable rungs — reach for them before writing a prose rule.** `tools/arch-lint`'s allowed-third-party-dependency check fails the build on a dependency added outside a package's allowlist, and `pnpm knip` fails on the unused export a deleted feature left behind. Those are the "don't add a dependency" and "delete it" rungs made mechanical. What the `simplifier` agent and its `ponytail` ladder add on top is only the judgement-shaped rungs (does this need to exist, is this one line) — inherently prose, and the weakest rung by design.
 
-**Measure before you change what you cannot see in the diff.** A heuristic, a cost model, a search or an optimisation is correct-looking code whose worth is entirely in numbers nobody has taken — so the INSTRUMENT lands first, in its own commit, and the change is judged by it. Two exist: `edge-routing-quality.test.ts` (quality: a corpus plus aggregate counts pinned EXACTLY, so an improvement is as loud as a regression, with debt metrics separated from price metrics) and `pnpm bench` (performance: `vitest bench`, compared across INTERLEAVED paired runs because machine drift between separate runs exceeds the effect). For a behaviour-preserving optimisation the scoreboard NOT moving is the proof. This is not ceremony: the instruments rejected three changes that were obviously right on argument — per-blocker detours, a placement cache, and an excess-length penalty tier — and one cheap reachability probe (67 of 68 remaining defects had a clean path available) retired a planned task and redirected the work. They also **rescued** one: an aligned re-score was rejected in its first two shapes on cost (13x, then 4.5x layout time) and shipped in a third that reused machinery already in the file, at 2x — the measurement did not veto the idea, it priced each shape until one was worth paying for. Load the `measured-change` skill. When a change buys one metric with another, `PENALTY_RULES`' declared tier order decides it; when the currencies are genuinely different, that is a human decision worth interrupting for.
+**Three things you cannot see by reading a diff each have a skill, and the skill is where the
+worked measurements live.** Load it before the work, not after:
 
-**A guard that never reaches its subject passes, and reads exactly like a
-guard that checked.** Distinct from a vacuous PROPERTY (a generator too sparse
-to reach the interesting inputs) and from an empty `--project` filter: here the
-assertion is right, the subject is simply absent from the fixture, so nothing
-is wrong for it to find. Three instances in one session, each costing a real
-defect:
-
-- `route-scope-registry.test.ts` walked apps built WITHOUT `ServerDeps`, and
-  `/api/v1` mounts only when they are supplied — so nine routes were exempt
-  from the registry-wide guard by accident. Every `/api/v1/*` path resolved to
-  `null`, which server mode answers 500 to.
-- Nothing migrated the data dir before `http-server.ts` handed its ports a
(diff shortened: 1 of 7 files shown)


kamiazya:
@coderabbitai full review

The automatic pass hit the included-review limit when this PR opened. CI is green on `f2fd52a` (all 21 checks, `verify` and CodeQL included), so this is a good point for a full pass rather than an incremental one.

Worth knowing while reading it: the diff is almost entirely rule and skill prose being **moved**, not rewritten. `AGENTS.md` and `.claude/rules/dev-flow.md` shed sections whose detail already lived in a skill or a contributor doc; anything the destination did not already hold was moved there rather than dropped. The one piece of executable code is `packages/mcp-server/src/server/dev-rules-budget.test.ts`, which pins the always-on context budget so a future change to it is a decision in a diff rather than drift.

---

PR #1164, opened 2026-08-31
refactor(web): give the editor's navigation gestures one state, not twelve refs

kamiazya:
**Layer 3 of 3**, on top of #1160 (the reducer) and #1162 (its property and ledgers). `SpatialEditor.tsx` now asks `navigation.ts` what a pointer event means and performs what it answers.

> Not a `gh stack` — no `gh` CLI in the authoring environment, so the layers landed as sequential PRs. See #1160.

## What goes

Seven refs, all of them things a gesture owned and nothing could say were over:

`downPointersRef` · `isPanningRef` · `touchPointsRef` · `pinchActiveRef` · `gatherAnchorRef` · `gatherPointersRef` · `lastPanPointRef`

With them, three constants and the `computePinchUpdate` import — the arithmetic moved into the machine with the state it belongs to.

## What kept it small

`fallThrough`. Each of the four handlers now asks the machine, performs its effects, and falls through to **exactly the code it had before** when the answer is "not mine". The manipulation half — marquee, node gestures, selection, click semantics, the long-press timer — is untouched. This is a wrapping, not a rewrite of 570 lines, and that was the point of designing the seam in layer 1 rather than discovering the need for one here.

## What stays, and why it was never navigation's

- **`activePointerIdRef`** is what the unmount teardown releases capture from, and overlay gestures that never reach `handlePointerDown` set it too. The machine asks for it through `capture` / `release-capture` effects — which is also what keeps the old hazard closed: an editor-wide "is something active" flag let a lifted finger's ordinary handback answer for a finger that was still down.
- **`lastPressRef`** narrows to what it always was underneath: the **Select** tool's double press, keyed by what was under it. Hand mode's has no target to key on, which is exactly why it needed a distance bound (#1159), and it now lives with that bound in the machine. The two shared one slot and never shared a key.

"What is down" has **one** answer, not two: an overlay press joins the machine's set through the `external-press` event rather than a second ref beside it.

## One deliberate divergence

A gather now ends on its **anchor's** release rather than on a member's, carried from #1162 where the property found it. The two sets are disjoint in every ordinary gather; they collide only when a released anchor's id is reused, and the old order then left a gather holding a pointer that was not down.

## Test plan

Behaviour-preserving, and the oracle is the point of the stack's order — the instruments were built first precisely so this could be judged rather than argued.

| suite | result |
|---|---|
| `spatial-editor` browser files — **the subject** | **82 files / 415 tests** ✅ |
| `pnpm test:browser` (whole) | **159 files / 875 tests** ✅ |
| `web-jsdom` | **294 files / 3028 tests** ✅ |
| `web-node` | **13 files / 86 tests** ✅ |
| `tsc --noEmit`, `pnpm knip` | clean |
| `arch-lint-node` | 10 files / 111 tests ✅ |

Every pinch, hand-tool, gather, long-press, marquee and node-drag case passes unchanged, as do the two browser properties that found the defects this work began with.

The compiler did the completeness check for the deletions: after wiring, **every remaining error was an unused declaration** — the seven refs and three constants, and nothing else. That is a stronger statement than "the tests still pass": nothing else was still reading them.

## Visual evidence

`Visual evidence: none — a behaviour-preserving refactor whose whole claim is that the picture does not change. A before/after figure of an unchanged surface would show two identical panels, which compose-figure.mjs refuses on principle. The evidence is the suite table above, and specifically the 415 spatial-editor browser tests that exercise the surface this touches.`

## Checklist

- [x] Commit message follows Conventional Commits
- [ ] Docs — no user-visible behaviour or published contract changes
- [x] No `console.*` added to `src/server/**`

## What this stack did and did not buy

Stated plainly, since it was the reason for doing it. The lifetime family is now structurally closed: navigation is one value, and `NAVIGATION_MEMORY_KEYS` names the only field allowed to survive a return to idle. The **predicate** family (#1159's double press missing a dimension) and the **DOM-wiring** family (#1158's overlay eating a press) are untouched by this design — a reducer cannot see either. Of the four defects this session found, one becomes unrepresentable. The properties found all four.

---

## Summary by CodeRabbit

* **Improvements**
  * Improved spatial editor navigation for mouse, touch, and pen input.
  * Added smoother panning, pinch-to-zoom, two-finger gestures, touch multi-selection, and hand-mode activation.
  * Improved gesture handling when interacting with overlay controls, including more reliable pointer capture and release.
  * Enhanced long-press, double-press, and gesture cancellation behavior for more consistent editing interactions.
  * Improved handling of interrupted or overlapping gestures to help prevent unexpected navigation behavior.

### DIFF
--- a/apps/web/src/components/spatial-editor/SpatialEditor.tsx
+++ b/apps/web/src/components/spatial-editor/SpatialEditor.tsx
@@ -150,6 +150,7 @@ import { MemberOutlinesOverlay } from './MemberOutlinesOverlay.js'
 import { MinimapOverlay } from './MinimapOverlay.js'
 import {
   createIdleNavigation,
+  DOUBLE_PRESS_WINDOW_MS,
   type NavigationEvent,
   type NavigationResult,
   type NavigationState,
@@ -432,11 +433,6 @@ function navigationPointerKind(pointerType: string): PointerKind {
 
 const LONG_PRESS_SLOP_PX = 10
 const DEFAULT_TEST_ID = 'spatial-editor'
-/**
- * Window for OUR double-press detection (see handlePointerDown). Matches the
- * common OS double-click interval; not user-configurable today.
- */
-const DOUBLE_PRESS_WINDOW_MS = 400
 
 /** Breathing room kept around framed content (zoom to fit / selection). */
 const FRAME_MARGIN_PX = 24
--- a/apps/web/src/components/spatial-editor/SpatialEditor.tsx
+++ b/apps/web/src/components/spatial-editor/SpatialEditor.tsx
@@ -149,6 +149,14 @@ import { MarkdownNodeEditor } from './MarkdownNodeEditor.js'
 import { MemberOutlinesOverlay } from './MemberOutlinesOverlay.js'
 import { MinimapOverlay } from './MinimapOverlay.js'
 import {
+  createIdleNavigation,
+  type NavigationEvent,
+  type NavigationResult,
+  type NavigationState,
+  type PointerKind,
+  reduceNavigation,
+} from './navigation.js'
+import {
   DOCUMENT_NODE_HEIGHT,
   DOCUMENT_NODE_WIDTH,
   fileNodeDefaults,
@@ -184,7 +192,6 @@ import {
   type EditorTool,
   ToolPalette,
 } from './ToolPalette.js'
-import { computePinchUpdate } from './touch-pinch.js'
 import { useGestureCaptured } from './use-gesture-captured.js'
 import { useWorkerScene } from './use-worker-scene.js'
 import {
@@ -226,16 +233,8 @@ const SNAP_THRESHOLD_SCREEN_PX = 6
 const SNAP_GRID_CANVAS_PX = 20
 
 /** Overview size. Big enough to aim at, small enough not to cover content. */
-/**
- * One step of hand mode's double-press zoom. Bigger than a wheel notch:
- * a tap that barely changed the view reads as a missed tap, and the way
- * back out is one press on zoom-to-fit rather than N reverse taps.
- */
-const DOUBLE_PRESS_ZOOM_FACTOR = 2
 /** One keyboard step of zoom — finer than the double press, which jumps. */
 const STEP_ZOOM_FACTOR = 1.25
-/** Press-pairing key for hand mode, where no node identity is involved. */
-const HAND_PRESS_KEY = 'hand'
 const MINIMAP_WIDTH_PX = 160
 const MINIMAP_HEIGHT_PX = 110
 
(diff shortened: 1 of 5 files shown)


kamiazya:
@coderabbitai review

The hourly limit has reset. Asking explicitly rather than merging past it: this is the riskiest change of the three-PR stack — a wiring change in the editor's pointer handling, deleting seven refs — and nothing is queued behind it, so a review is worth the wait here in a way it was not for the two foundation layers.

Worth a reviewer's attention specifically:

- **`runNavigation`'s effect switch** — each arm is meant to be exactly what the deleted code did at the site it replaces. An arm that does something *nearly* the same is what a passing suite would hide.
- **The capture handling** — `capture` / `release-capture` replace `activePointerIdRef` bookkeeping that was spread across four handlers. `release-capture` is deliberately not emitted when a finger leaves a gather or a pinch, because the others are still down and still captured.
- **Ordering inside `handlePointerMove`** — capture-on-first-move is now taken *before* the machine reduces the move, so it reads the pan the press started rather than the one the move is about to advance. That ordering is load-bearing and the tests would pass either way for a mouse.

---

kamiazya:
## Two things happened while this sat, both recorded rather than worked around

**1. CodeRabbit did not review this PR.** I asked explicitly rather than merging past the limit, because this is the riskiest layer of the three. Its reply was `Review rate limited. Action not completed.` — nine hours later there is still no review. That is a definitive skip, not a pending one, so per `dev-flow.md` I am proceeding and noting it here. Three PRs in this stack in a row went unreviewed for the same reason. The verification this change rests on is therefore entirely local and CI: 415 spatial-editor browser tests, 875 browser, 3028 jsdom, and the compiler confirming that after wiring **every** remaining error was an unused declaration.

**2. `origin/main` is red on a test this PR does not touch**, and `fd720d41` carries the one-line fix.

Established before claiming it, from main's own blobs rather than from a property of my diff:

```
16506  AGENTS.md
17884  .claude/rules/architecture-map.md
28246  .claude/rules/dev-flow.md
13388  .claude/rules/integrator-flow.md
15699  .claude/rules/vocabulary.md
-----
91723  → bucket 91, against ALWAYS_ON_TOTAL_BUDGET = 90
```

91723 is digit-for-digit what CI reports. `git diff origin/main...HEAD -- .claude AGENTS.md` is empty, so this branch cannot be the cause.

A merge race: #1163 pinned the total from its own branch, #1165 grew `architecture-map.md` with ADR-0018 afterwards, each was green alone and neither CI saw the sum. Every **per-file** pin still holds — this is exactly the case that constant's own comment names, that the bucket of the sum is not the sum of the buckets.

Re-pinned rather than trimmed because the test states which it is: *"the instrument, not a limit … growth is legitimate — a rule that earns always-on status should be always-on."* ADR-0018's acceptance earned it, and the pin exists so that is a decision in a diff.

It is carried in this PR only because this environment can push to a single branch; it is unrelated to everything else here, reviewable on its own, and no-ops the moment the base carries an equivalent fix.

---

PR #1174, opened 2026-09-01
refactor(web): retire the orphaned user-libraries surface from StorageReportCard

kamiazya:
# Summary

- Deletes the user-libraries (`.excalidrawlib` pack) management surface from `StorageReportCard` — the schema block, the `libraries` category row with its soft-cap rendering, the list/remove dialog, and the fetch/delete wiring (705 → 517 lines).
- The feature is excalidraw-era and retired (user decision, 2026-08-31). Its server half is already gone: the daemon has no `/api/v1/user-libraries` route (live daemon answers 404 on both spellings), the storage categorizer (`runtime-storage.ts`) never emits a `libraries` bucket, and the `api-contracts` entry was removed earlier — so the client card kept a whole management surface that could never receive data.
- No contract change needed: `storageReportPayloadSchema.byCategory` is an open `z.record`, so the schema is untouched.
- Deliberately untouched: `validators.test.ts`'s seven `.excalidrawlib` URLs (SSRF-guard samples about a foreign format) and `vocabulary.md`'s foreign-format bullet (history).

# Test plan

- Red-first: new test `offers no user-libraries surface — the daemon retired the feature it called` pins the absence of the `libraries` row, the manage dialog, and any "User librar…" copy, while asserting the `blobs` row IS present (subject presence, so the absence cannot pass vacuously on an empty render). Confirmed red against the pre-deletion component for the right reason, green after.
- Removed the five stale libraries tests and the fixture's `libraries` bucket; row-order pin updated (`other` last, 7 categories).
- `StorageReportCard.test.tsx` + `SettingsPage.test.tsx`: 48/48 green; `pnpm -r typecheck`, biome (incl. noconsole), knip clean; repo-wide grep for the retired surface clean outside the two intentional keepers above.

# Visual evidence

Visual evidence: none attachable — no image upload path exists in this environment. Additionally the card only mounts when a daemon connection is present (Settings → Connections), which the scripted capture environment could not establish; verification is via the component tests above plus the live-daemon 404 evidence for the deleted route.

# Checklist

- [x] Nearest-layer automated test for the root cause (absence test with subject-presence guard)
- [x] Manual verification: live daemon 404 on `/api/v1/user-libraries` confirming the surface is orphaned
- [x] `pnpm -r typecheck` / biome / knip green
- [x] No model identifier in repository artifacts

---

### DIFF
--- a/apps/web/src/components/StorageReportCard.test.tsx
+++ b/apps/web/src/components/StorageReportCard.test.tsx
@@ -11,7 +11,6 @@ const PAYLOAD = {
     blobs: { bytes: 1024, files: 2 },
     versions: { bytes: 2048, files: 1 },
     files: { bytes: 0, files: 0 },
-    libraries: { bytes: 0, files: 0 },
     db: { bytes: 1024, files: 1 },
     other: { bytes: 0, files: 2 },
   },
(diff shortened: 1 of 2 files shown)


kamiazya:
Closing as superseded: #1161 (`63e416ea`, already on main) retired the same user-libraries surface, with a stronger guard — the rendered rows are compared against `storageCategorySchema.options` itself, so a category the contract loses (or gains) fails the test, which subsumes this PR's absence assertions. Main's only remaining `librar` mention in the card is the enduring-rationale comment. Nothing from this branch is worth carrying over.

---

PR #1202, opened 2026-09-01
refactor: move countAliveNodes out of store/, where it was never a mechanic

kamiazya:
## Summary

First slice of [ADR-0018](docs/contributing/adr/0018-operation-vs-mechanic.md)'s burn-down, which names `routes/document/restore.ts` as the cluster to pay off first (5 of the 39 allowlisted adapter → mechanic edges). **Two of those five turn out not to be debt at all.**

- `count-alive-nodes.ts` is a **pure function over a `LoroDoc`** — depends on loro-adapter, loro-crdt and zod; touches no store, cache, lock or scheduler; and answers *"how many elements does this doc have"*. By ADR-0018's own deciding question — *could a second surface need this?* — that is plainly an operation-side reader. It was classified as a mechanic **only because `adapter-mechanic-check.ts` classifies by directory and the file sat under `server/store/`**, so two routes that legitimately ask a document its size were counted as ADR-0018 debt.
- Moved to `server-core` as `document-counts.ts` — where a shared reader belongs, and where the second surface the ADR is about can actually reach it. `version-store.ts` keeps using it; a mechanic importing shared code is the allowed direction.
- Removes both `-> count-alive-nodes` edges (`routes/debug.ts`, `routes/document/restore.ts`) and lowers `ADAPTERS_REACHING_MECHANICS_CEILING` **39 → 37**.

**What this deliberately does not do.** The remaining four edges on `restore.ts` (`doc-cache`, `document-store`, `version-store`, `workspace-lock`) are genuine mechanics. Removing them means moving the restore operation itself into server-core, and that needs new **required** seams on `ServerDeps`: server-core reaches documents through the snapshot port, whereas `reconcileCommitSaveBroadcast` must mutate *the same live doc instance a connected client already holds* (a delta broadcast only means something against the doc it was diffed from). The ADR says to land a required-field change on its own small PR — it measured three CI failures from merge races alone during #1035 — so that is PR 2 (seams) and PR 3 (the operation move, 37 → 33), not this one.

## Test plan

- [x] New or updated `mcp-node` / `mcp-jsdom` / `mcp-browser` / `web-browser` tests added — the module's existing test moved with it and is unchanged in substance.
- [x] `pnpm test` passes locally — `server-core-node` + `mcp-node`: **415 files / 4637 passed, 9 skipped**, exit 0. `arch-lint-node` 112/112. `pnpm -r typecheck` green across all packages; `pnpm lint` clean.
- [x] Manual verification completed — red-first, and **the ratchet was mutation-checked in both directions** rather than assumed, since a guard that only fires one way reads exactly like one that checked:

  | state | result |
  |---|---|
  | ceiling lowered to 37, edges still present | **red** — `reports no adapter -> mechanic edge outside the allowlist` |
  | edges paid off, ceiling put back to 39 | **red** — `an edge was paid off — lower ADAPTERS_REACHING_MECHANICS_CEILING to match`, `expected 37 to be 39` |
  | edges paid off, ceiling 37 | green, 112/112 |

- [ ] E2E coverage added or extended where applicable — n/a; behaviour-preserving move, no surface changed.

### Note on the moved test

It inlines its two-line `makeSpatialDoc` over the real `writeSpatialCanvas` bridge rather than importing mcp-server's `shared/test-utils/spatial-doc.ts`. Nine mcp-server tests share that file; hauling it across a package boundary to avoid two lines would be the larger change. The fixture is still built through the real bridge, which is that helper's own stated reason for existing.

## Visual evidence

Visual evidence: none — a module move plus an allowlist ratchet. Nothing renders. The observable output is the arch-lint verdict, quoted in the table above, including both failing directions.

## Checklist

- [x] Docs updated if this changes user-visible behavior or a published contract — n/a; the reclassification's rationale is recorded on the moved file, replacing a doc comment whose "the daemon's" framing came from the old location

---

### DIFF
--- a/packages/mcp-server/src/server/routes/debug.ts
+++ b/packages/mcp-server/src/server/routes/debug.ts
@@ -1,6 +1,6 @@
+import { countAliveNodes, countLegacyTombstones } from '@kamiazya/whiteboard-server-core'
 import { Hono } from 'hono'
 import { isAuthorized } from '../security/bearer-token.js'
-import { countAliveNodes, countLegacyTombstones } from '../store/count-alive-nodes.js'
 import { getCacheKeys, peekDoc } from '../store/doc-cache.js'
 import { listDocuments, listWorkspaces, loadDocument } from '../store/document-store.js'
 
--- a/packages/mcp-server/src/server/routes/document/restore.test.ts
+++ b/packages/mcp-server/src/server/routes/document/restore.test.ts
@@ -20,7 +20,7 @@ const { clearCache, peekDoc } = await import('../../store/doc-cache.js')
 const { getDoc, saveDocument, getDocumentKind, loadDocument, onWorkspaceDocUpdated } = await import(
   '../../store/document-store.js'
 )
-const { countAliveNodes } = await import('../../store/count-alive-nodes.js')
+const { countAliveNodes } = await import('@kamiazya/whiteboard-server-core')
 const { createDocumentRouter } = await import('../document.js')
 // Pre-load ws.js before any restore call, mirroring restore-race.test.ts's
 // documented cycle workaround for document.ts's dynamic import.
--- a/packages/mcp-server/src/server/routes/document/restore.ts
+++ b/packages/mcp-server/src/server/routes/document/restore.ts
@@ -4,10 +4,10 @@ import {
   reconcileDocContent,
 } from '@kamiazya/whiteboard-loro-adapter'
 import type { DocumentKind } from '@kamiazya/whiteboard-model'
+import { countAliveNodes } from '@kamiazya/whiteboard-server-core'
 import { Hono } from 'hono'
 import type { LoroDoc } from 'loro-crdt'
 import { restoreVersionRequestSchema } from '../../../shared/api-contracts/document.js'
-import { countAliveNodes } from '../../store/count-alive-nodes.js'
 import { evictDoc } from '../../store/doc-cache.js'
 import {
   ConflictError,
--- a/packages/mcp-server/src/server/store/version-store.ts
+++ b/packages/mcp-server/src/server/store/version-store.ts
@@ -4,6 +4,7 @@ import {
   projectWorkspaceDocument,
   resolveWorkspaceDocument,
 } from '@kamiazya/whiteboard-loro-adapter'
+import { countAliveNodes } from '@kamiazya/whiteboard-server-core'
 import { DocumentStoreWorkspaceDocs } from '@kamiazya/whiteboard-workspace-index'
 import type { Frontiers } from 'loro-crdt'
 import { decodeFrontiers, encodeFrontiers, LoroDoc } from 'loro-crdt'
@@ -17,7 +18,6 @@ import {
   validateWorkspaceId,
 } from '../validators.js'
 import { corruptStoredData, isMissingFileError } from './corrupt-stored-data.js'
-import { countAliveNodes } from './count-alive-nodes.js'
 import { getDb } from './db/index.js'
 import { prepareDataDir } from './db/prepare.js'
 import { DocumentNotFoundError } from './document-not-found-error.js'
--- a/packages/server-core/src/document-counts.test.ts
+++ b/packages/server-core/src/document-counts.test.ts
@@ -1,7 +1,20 @@
+import { writeSpatialCanvas } from '@kamiazya/whiteboard-loro-adapter'
+import type { SpatialCanvas } from '@kamiazya/whiteboard-model'
 import { LoroDoc, LoroMap } from 'loro-crdt'
 import { describe, expect, it } from 'vitest'
-import { makeSpatialDoc } from '../../shared/test-utils/spatial-doc.js'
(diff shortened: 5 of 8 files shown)


kamiazya:
## `test-jsdom` is red, and it is not this PR's

```
FAIL web-jsdom src/components/spatial-editor/editor-state.property.test.ts
  > editor composite state (command-based)
AssertionError: no cut surface was ever reconnected: expected 2 to be greater than 2
Test Files  1 failed | 302 passed (303)
Tests  3142 passed (3142)
```

**This diff cannot reach that test.** `git diff origin/main -- apps/web packages/model packages/codec packages/canvas-render packages/loro-adapter` is empty, and `apps/web` has no `@kamiazya/whiteboard-server-core` dependency and no import of it — the only packages this PR touches are `server-core`, `mcp-server` and `tools/arch-lint`.

**What failed is a reach guard, not the property.** All 3142 tests passed; the red is the `afterAll` coverage block (hence "Failed **Suites** 1" while the summary reads all-passed — the shape `.claude/rules/coverage-ledger.md` warns about). `stats.reconnections` counts how often the generated command sequences actually reached a cut-with-boundary-edges arrangement.

### Measured rather than argued

I forced the assertion to report its value and ran the file five times on this branch (which is byte-identical to `origin/main` for everything this test imports):

| run | 1 | 2 | 3 | 4 | 5 |
|---|---|---|---|---|---|
| `reconnections` | 8 | 14 | 13 | 13 | 10 |

That sits inside the range the test file itself records — `// ... paste-inserts 38-59, reconnections 8-13, marquee selections 23-30` — from which the `> 2` floor was derived as "roughly a third of the minimum measured". So the arrangement **is** being reached normally; CI's 2 is a 4× outlier against the recorded minimum of 8, not a diluted generator.

I want to be honest about the limit of that: I could not reproduce CI's value locally and fast-check prints no seed for a passing property, so I have **no confirmed mechanism** for the local-vs-CI gap. What the numbers establish is that the counter runs far closer to its floor on CI than the measurement the floor was calibrated from — and that measurement was taken on a dev machine, not a runner.

### It is not an isolated incident

`test-jsdom` also went red **on `main` itself** three hours earlier, at `51ca4639` (run [33554108510](https://github.com/kamiazya/whiteboard/actions/runs/33554108510), the only failing job), on a different guard of the same kind:

```
AssertionError: the page never switched, so the DUPLICATE case would pass
without exercising anything
```

Two different vacuity guards in the same suite, one of them on `main`, inside three hours.

### Why I am not fixing it here

`.claude/skills/steward/reference/failure-modes.md` already carries the verdict for this class:

> | a reachability/density guard fails intermittently | The generator is too sparse. Denser domain plus a measured floor — never a pinned seed, never fewer assertions |

That fix belongs in `apps/web`'s property test. This PR is an ADR-0018 reclassification in `server-core`/`mcp-server`/`tools/arch-lint`; pulling an unrelated package's generator into it would widen the diff across a boundary for a reason that has nothing to do with the change under review. Per the same rules I am reporting it rather than widening, and **no fix PR exists yet to port**.

I also **cannot re-run the job** — `rerun-failed-jobs` returns `403 Resource not accessible by integration` for this integration — so the one permitted re-run is not available to me. Noting that here rather than leaving the PR to look unattended.

### Proposed patch, for whoever picks it up

Per the catalogue's verdict, and per this test file's own note that *"more runs is the honest lever for variance"* (it went 300 → 500 for exactly this reason when two counters landed on their floor):

1. **Re-measure the floors on a CI runner, not a dev machine.** Every number in that comment block — 30 counters — was taken locally. If `reconnections` really runs at ~2 on a runner while reading 8–14 here, then every floor in the block is calibrated against the wrong machine and this will keep firing on whichever counter is unluckiest.
2. **Then make the domain denser for the sparse counters** rather than lowering the floors, so the guard keeps meaning "this was exercised".
3. **Not** a pinned seed, **not** a widened budget, **not** fewer assertions.

I will keep this PR watched and re-check its state on a schedule until it is green, merged or closed.

---

kamiazya:
Closing this **parked, not abandoned**, at the author's direction: the `test-jsdom` vacuity flake documented above was reprioritised over the ADR-0018 burn-down, and this session can push to only one branch, so the branch is being reused for the flake fix rather than carrying two unrelated subjects in one diff.

Nothing here is lost or withdrawn. The work is one commit, `fa14dd614ab21ab3c84f5ee6a0b907f10bbc0766`, reachable from this closed PR, and it will come back as its own PR once the flake fix lands. Its conclusion stands unchanged and is worth restating so it does not have to be re-derived:

> `count-alive-nodes.ts` is a pure function over a `LoroDoc` and was never a mechanic. It counted as ADR-0018 debt only because `adapter-mechanic-check.ts` classifies by directory and the file sat under `server/store/`. Moving it to `server-core` removes both `-> count-alive-nodes` edges and lowers the ceiling 39 → 37.

Every check on it was green except the unrelated `test-jsdom` failure; `server-core-node` + `mcp-node` ran 415 files / 4637 passed locally, and the ratchet was mutation-checked in both directions.

---

PR #1229, opened 2026-09-02
refactor(mcp-server): move version restore into a server-core operation

kamiazya:
## Summary

[ADR-0018](docs/contributing/adr/0018-operation-vs-mechanic.md)'s first scheduled cluster. `routes/document/restore.ts` — five adapter → mechanic edges when the ADR was accepted, four after #1211 — is now a **translator**: it parses the request, calls `restoreVersion`, and maps the result and its refusals onto a response. Every decision about what a restore *means* lives in `server-core`, where a second surface can reach it.

- **`operations/restore-version.ts`** carries all four modes: in-place reconcile, restore into a new target, overwrite an existing target, subtree rollback. Addressed by **path on the way in and documentId everywhere after** — which is what removes the hazard the old route guarded against at length (a concurrent delete/rename between read and save inserting a phantom row). That hazard is a property of path addressing; an id is stable across a rename and absent after a delete, so resolving once dissolves it rather than defending against it.
- **Four allowlist entries retired; `ADAPTERS_REACHING_MECHANICS_CEILING` 37 → 33.**
- Two mechanic-side fixes the move needed, each its own commit and each mutation-checked:
  - `FileVersionStore.load/loadWorkspaceAt/list` now take the (reentrant) workspace lock. They open the stored record on the shared libsql connection, and the old route's handler-wide lock had been serialising them against writers by accident. Without it, five route tests went 500 with `SQLITE_BUSY: cannot commit transaction - SQL statements in progress`; with it they went green.
  - `WorkspaceRoutedDocumentStore.saveSnapshot` moves the tree entry's kind when the saved content declares a different one — the channel every tool write already uses (`document-set`, `canvas-edit`, `document-crud` all stamp kind). The route used to do this by hand.

### Three premises refuted before writing this

The first estimate was ~1000 lines and four new seams. Reading the code refuted three of the four, which is why it is 341 insertions against 358 deletions:

| premise | finding |
|---|---|
| a `liveDocument` transaction seam is needed | **no** — the daemon's `documentStore` binding already holds the lock and merges into the live projection |
| a lock seam is needed for the phantom-insert hazard | **no** — the port resolves by documentId inside the lock |
| `saveSnapshot` lacks cache eviction on failure | **no** — `saveWorkspaceDoc` evicts both caches in its catch (#1223 pinned it) |

The one seam that was genuinely missing, `versions`, landed in #1216.

## Test plan

- [x] New or updated tests added — 12 operation-layer tests over the four modes; one mechanic test each for the lock and the kind sync; the route suite keeps 18.
- [x] `pnpm test` passes locally — `mcp-node` **366 files / 4205 passed, 9 skipped** (run alone, since its loop-availability tests are load-sensitive); `server-core-node` 52 / 455; `arch-lint-node` 119 / 119 at ceiling 33; `pnpm -r typecheck` 0 errors; `pnpm lint` and `pnpm knip` clean.
- [x] Manual verification completed — **mutation-checked rather than assumed**, and the checks did real work. Of the ten mutations run against the operation, **five survived the first version of its tests**, each because a behaviour the comments called important was never actually asserted:

  | mutation | first result | closed by |
  |---|---|---|
  | overwrite an existing target without being asked | caught | — |
  | **swap content instead of reconciling** an existing target | **survived** | asserting **lineage** (`oplogVersion().compare`), not node ids — both end showing the past state |
  | drop the `targetPath === path` collapse | caught | — |
  | **don't carry the source's kind** onto a new target | **survived** | asserting the created entry's kind |
  | delete *after* restoring in a subtree rollback | caught | — |
  | resurrect a document deleted since the version | caught | — |
  | **skip the rename** in a subtree rollback | **survived** — deleting the move left 10/10 green | a case for a document that moved while staying alive |
  | never announce `restore_started` / never send `restore_complete` | caught (once the test existed) | — |

  Also verified: dropping either version-store read lock reintroduces the 500s; dropping the kind sync fails its test.
- [ ] E2E coverage — **not run against a live daemon.** This container cannot drive the real History panel. The 18 HTTP-level route tests and the operation tests are the evidence; a manual restore from the History panel before merge would close that gap.

### One route test rewritten, not weakened

`evicts the cached doc when reconcile/commit itself fails` spied `commit` on the cached live doc and asserted eviction — a mechanism the operation no longer has (it reconciles a fresh copy; only the keeper's save touches the live projection). It now asserts the **stronger** guarantee the design bought: a restore refused before persistence leaves the live document untouched, the *same instance* still cached, and the reader still served the persisted state. The failure-*during*-persistence case, where the live doc is already mutated, keeps its own unchanged test.

### Surfaced, not resolved: two version histories

`wb_version_restore` (server-core) reads a version record the document carries in its own `versions` map. This operation, the HTTP route and the History panel read the daemon's `FileVersionStore`. **A version saved through one is invisible to the other** — confirmed: `wb_version_save` writes `doc.getMap('versions')` and nothing under `store/` reads that map. That is a standing second answer to what restoring means, of the kind ADR-0018 exists to retire. It predates this PR, is recorded on `NoSuchVersionError`'s doc comment, and which history is canonical is a product decision I have not made here.

## Visual evidence

Visual evidence: none — an HTTP route's internals moved behind an operation with the same request and response shapes. Nothing renders differently. The observable outputs are the gate counts and mutation table above.

## Checklist

- [x] Docs updated if this changes user-visible behavior or a published contract — n/a; no contract changed, and the rationale is on the code
- [x] No hand-written TypeScript interfaces duplicate a Zod schema — input/output are `z.infer` of `restoreVersionInputSchema` / `restoreVersionOutputSchema`

https://claude.ai/code/session_016op9NR7gZeygmEoJ6UJzAu

---

### DIFF
--- a/packages/mcp-server/src/server/routes/document.ts
+++ b/packages/mcp-server/src/server/routes/document.ts
@@ -67,7 +67,7 @@ export function createDocumentRouter(options: DocumentRouterOptions = {}) {
   app.route('/', createMaintenanceRouter({ versionStore }))
   app.route('/', createDocumentSvgExportRouter())
   app.route('/', createThumbnailsRouter({ versionStore }))
-  app.route('/', createRestoreRouter({ versionStore }))
+  app.route('/', createRestoreRouter({ serverDeps: options.serverDeps }))
 
   return app
 }
--- a/packages/mcp-server/src/server/routes/document/restore.test.ts
+++ b/packages/mcp-server/src/server/routes/document/restore.test.ts
@@ -733,4 +733,68 @@ describe('overwrite restore reconciles instead of replacing', () => {
     const body = (await restoreRes.json()) as { error: string }
     expect(body.error).toBe('output_exists')
   })
+
+  // The overlay a client shows while a restore runs is driven by these two
+  // messages, and until now nothing asserted the server SENDS them. The
+  // consumer side is well covered — `sse-backend`, `daemon-backend` and
+  // apps/web's sync hooks all test what happens when one ARRIVES — so a
+  // change that stopped emitting them would leave every one of those tests
+  // green while the overlay silently stopped appearing and, worse, stopped
+  // being dismissed.
+  it('brackets a restore with restore_started and restore_complete for a connected client', async () => {
+    const app = createDocumentRouter({ autoVersionIntervalMs: 60_000 })
+    const initial = new LoroDoc()
+    const vv0 = initial.version()
+    writeSpatialCanvas(initial, {
+      nodes: [{ id: 'n1', type: 'text', x: 0, y: 0, width: 80, height: 40, text: 'before' }],
+      edges: [],
+    })
+    initial.commit()
+    await app.request('/api/w/session1/document/canvas-a/update', {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/octet-stream' },
+      body: initial.export({ mode: 'update', from: vv0 }),
+    })
+    const saveRes = await app.request('/api/workspaces/session1/documents/canvas-a/versions', {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({ label: 'v1' }),
+    })
+    const { version } = (await saveRes.json()) as { version: { id: string } }
+
+    const sent: string[] = []
+    const socket = {
+      send: (data: unknown) => {
+        if (typeof data === 'string') sent.push(data)
+      },
+      on: () => {},
+      close: () => {},
+    }
+    const { handleWsUpgrade } = await import('../ws.js')
+    await handleWsUpgrade(
+      { url: '/ws/session1/canvas-a', headers: { host: 'localhost:3099' } } as never,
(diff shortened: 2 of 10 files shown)


kamiazya:
Closing as superseded. #1224 (with #1226 / #1227 stacked on it) landed the same move — the restore route's mechanic reach extracted into a `server-core` operation — behind a `LiveDocuments` seam while this branch was still in flight, and the two designs cannot coexist: this PR conflicts on six files and merging it would mean unwinding those three.

What this branch adds on top of main's design is two tests, and only those are worth carrying over. They will follow as a test-only PR:

- a producer-side test that the restore route brackets the operation with `restore_started` / `restore_complete` (main's `restore.test.ts` has no assertion on either);
- a subtree rollback test for deletion-before-resurrection ordering — measured against main's 14 operation tests, the mutation that runs deletions after restores survives.

The lock around `FileVersionStore` reads and the mechanic-side kind sync are not ported: main's `withWriteLock` and `save({ kind })` solve the same two problems in its own shape.

---

PR #1245, opened 2026-09-02
feat(web): one version history for both keepers — bookmarks, checkpoints at a pause, preview-then-restore and restore lineage

kamiazya:
## Summary

The browser keeper gets a version history: manual save (⌘/Ctrl+S, the save dot, or the **Save version** button in the History panel), the History panel, and restore — the same panel, rows and restore semantics the daemon has, minus what the daemon adds on top (automatic checkpoints, thumbnails, branches).

- **Same design as the daemon's history, on purpose.** A version is a FRONTIER into the workspace record plus the row the panel lists — never a copy of content. The browser already keeps the whole workspace as one Loro record in IndexedDB (`DocumentStoreWorkspaceDocs`, shared with the daemon), and its fold re-snapshots rather than shallow-exports, so a frontier saved before any number of folds still checks out after them. A per-document copy would be the second place to keep a version, which is exactly the defect #1235 removed on the daemon side.
- **`BrowserVersionStore`** (`versions` object store, `DB_VERSION` 16, index `byDocument` on `[workspaceId, documentId]`): `save` records the STORED record's frontier and the row; `list` hydrates the path from the index so a rename keeps its history; `loadPast` clones the record, checks the frontier out and projects the document — and answers `null` for a version that belongs to another document, for the reason the daemon's operation refuses it.
- **`BrowserBackend.applyRestore`**: the same reconcile the daemon runs (`writeWorkspaceDocumentContent`, a diff and never a rewrite) on the record the backend holds, persisted, then delivered to the sync session as a remote update — the way a peer's edit would arrive — bracketed by `onRestoreStarted` / `onRestoreComplete`. Queued behind pending pushes so a keystroke in flight is reconciled over, not lost under, the restore.
- **`VersionsBackend` seam** (`lib/versions-backend.ts` + `VersionsBackendContext`): `list` / `save` / `restore` / optional `putThumbnail`. The daemon implementation wraps the routes; `useVersionsBackend()` falls back to it over `DaemonApiContext`'s fetch, so every daemon page and every existing test that mocks the routes is unchanged. `VersionTimeline` and `useSaveVersion` now speak the seam; `useBranches` gains `{ enabled }` so a keeper with no branches answers the resting state instead of logging a 404.
- **`VersionTimeline` capabilities prop** (`branches`, `autoVersions`, default both true): threaded through `HistoryCluster` → `VersionPanel`; the browser passes both false and gets the one-lane graph and an empty-state line that does not promise an auto-save.
- **`versions` leaves `WhiteboardCapabilities`** (second commit). The repo's own rule — a flag both keepers set the same way is not a capability, the one that already retired `workspaces` — said so once the browser had a history. The daemon page's "Version history" teaser and the shell's connect-a-daemon copy go with it (the copy now names what the daemon actually adds). Hiding the save dot for a document with no history backend (the browser's markdown documents, for now) is a page fact, so `WorkspaceTopBar` takes `versionsEnabled` for that.
- **A touch-reachable save** (fourth commit). ⌘/Ctrl+S is the only save path a phone does not have, and the save dot is a status affordance rather than a labelled control, so the History panel gets an explicit **Save version** button beside the empty state and the list. Same `useSaveVersion` call the shortcut makes.
- Not in this PR: automatic checkpoints (the IndexedDB growth has to be measured first), thumbnails, branches/merge.

## Test plan

  - `web-browser` (real IndexedDB): `browser-version-store.browser.test.tsx` — save → list newest-first → `loadPast` round trip through a FRESH store after a simulated reload, plus the other-document refusal; `browser-backend.restore.browser.test.tsx` — session doc reaches the past state through `onRemoteUpdate`, record persisted, bracket order, and the bracket completing when persistence fails; `BrowserDocumentPage.versions.browser.test.tsx` — the whole flow through the real page: panel open with the browser's empty-state copy, ⌘/Ctrl+S, the panel's own Save version button, the open panel picking the row up through the save-driven refresh (mutation-checked: dropping the refresh listener fails it), reopen on a later state, restore from the panel, record back at the saved state.
  - `web-jsdom`: `useSaveVersion.test.ts` re-injects the fetch through `DaemonApiContext`; `provider.test.ts` pins that `versions` is gone; `WorkspaceTopBar.test.tsx` gates on `versionsEnabled`; the two `DaemonDocumentPage` tests for a `versions: false` daemon are deleted with the capability; `VersionTimeline`, `useBranches`, `history-cluster` suites unchanged.
- [x] `pnpm test` passes locally — full apps/web run on the first commit: jsdom+node 3182/3186, web-browser 927/930; the 7 failures were the repo's guards reacting to the change (capability-difference rule, screen-state ledger, test-name budget, migration store list) and are fixed in the second commit, after which those suites plus everything the retirement touched pass (jsdom 231/231 across 10 files, browser 48/48 across 5); typecheck; `pnpm knip` clean
- [x] Two pre-existing flakes CI surfaced under load are fixed here rather than re-run, since both were tests sampling a moving thing once:
  - `navigation.property.test.ts`'s double-press generator drew `dt` uniformly over a range where the interesting window was 1 case in 6, so the property was vacuous most runs — densified (`fc.oneof` weighting the short gaps), counts measured 4-5 → 9-13 per run.
  - `PromoteWorkspaceSection.browser.test.tsx` read the progress live region once after `findBy` resolved; under load the 150ms document phase was already over and it read the images phase, reporting a correct sequence as a missing step. It now records phases with a `MutationObserver` from before the transfer starts and asserts their ORDER.
- [ ] Manual verification completed — on the preview
- [x] E2E coverage added or extended where applicable — the page-level browser flow above is the E2E for this keeper (no daemon, no routes, real IndexedDB)

## Visual evidence

Visual evidence: none — no pixels move for the history itself; the History panel and save dot are the existing components rendered where they were hidden before, which the page-level browser test asserts through the real DOM. The one new pixel, the Save version button, is a labelled button inside that panel and is asserted by the same test.

## Checklist

- [x] Docs updated if this changes user-visible behavior or a published contract — README pairing copy and the shell's connect-a-daemon copy (what the daemon adds on top of the browser's own history)
- [x] No hand-written TypeScript interfaces duplicate a Zod schema — `versionRowSchema` is Zod; `VersionsBackend` is an in-process seam, not a wire contract

https://claude.ai/code/session_016op9NR7gZeygmEoJ6UJzAu

## Summary by CodeRabbit

- **New Features**
  - Added persistent browser document history with named bookmarks, browsing, read-only previews, and restore after reopening.
  - Added Cmd/Ctrl+S support for opening the named-bookmark field.
  - Added automatic daemon checkpoints after editing pauses and during extended editing sessions.
  - Added Undo and Redo buttons to the Markdown editor.

- **Updates**
  - Updated history panels with responsive layouts, clearer details, restore lineage, and stale-data messaging.
  - Updated connection messaging to describe “automatic checkpoints.”
  - Improved canvas viewer sizing and framing.


kamiazya:
## Review triage

All nine CodeRabbit threads verified against the code. Six fixed in `0b380f5`, one already addressed, two refused with reasons.

### Fixed — two could lose something a person cannot get back

**Preview change during an in-flight restore** (`VersionTimeline.tsx`). Confirmed. `restorePreviewed` decides whether to run its success effects by comparing the version it asked for against what the panel shows when the response lands, so choosing a second row in that window skipped `onRestored()` and the refresh **for a restore that succeeded** — the undo history kept the pre-restore document while the document itself had already moved. `openPreview` now refuses while restoring, as `closePreview` already did.

**The automatic sweep took restore lineage with it** (`version-store.ts`). Confirmed, and worse than the report: a merge point is saved as an *automatic* version, so past the 50-checkpoint cap **both ends** of a restore were ordinary sweep candidates — the merge, and the point it named. A swept checkpoint is recoverable by editing again; a swept lineage is not. The cap now bends around a merge and around anything a merge references.

Its guard asserts the surviving **count** rather than a survivor list, because that one number says both halves: 53 saved, 52 survive ⇒ the sweep ran *and* spared exactly the two rows lineage needs. Had the protection not applied, all three over the cap would have gone and the count would be 50 — so it cannot pass by the pair merely happening to sit inside the cap. Mutation-checked.

### Fixed — four smaller

- **`putThumbnail` ignored `res.ok`** — a fetch resolves for 4xx/5xx, so a refused upload was indistinguishable from an accepted one and the caller's failure path was unreachable by anything but a network error.
- **Save results not bound to their document**, both pages. Neither remounts on a switch and the scope reset has already cleared the outcome by the time a slow save settles, so the message appeared under the *arrived* document as if that one had been saved. Both now fix the identity before the first await, the way `handleDuplicate` already did.
- **Historical markdown preview ignored dark mode** — neither call site passed the resolved theme.
- **Shutdown could still drop a checkpoint.** The registry's stop flushes, but `server.close()` keeps serving requests already in progress, and an update handler completing in that window arms a fresh debounce against an `unref`ed timer that never fires. A second flush runs once the listener is closed and nothing can arm another. The first stays — it is what runs on a bind-failure teardown, where there is no server to close.

### Already addressed

**Flush the auto-version trigger during shutdown** — landed in `b8d3af8`..`8841566`, and the thread is marked resolved.

### Refused

**`useBranches`: invalidate and reset on disable.** The reasoning is right — a request in flight when `enabled` flips `true → false` keeps a valid sequence and would commit into the disabled resting state. But `enabled` is `capabilities.branches`, which is fixed per page (daemon `true`, browser `false`) and never toggles at runtime, so the transition the finding depends on cannot occur. Adding a guard for a state that cannot be reached is code with no reader; if branches ever become switchable, this becomes real and should be fixed then.

---

PR #1270, opened 2026-09-03
feat(mcp): comment on any document through wb_thread_edit

kamiazya:
ADR-0026 decision 6, and **step 4** of its order of work (data → readers → UI → MCP). Independent of #1269: it stands on the merged data layer (#1267) and touches no `apps/web` file.

## Why a new tool rather than more `wb_canvas_edit` ops

`wb_canvas_edit`'s `comment.*` ops are canvas-scoped, so **an agent cannot comment on a markdown document at all**. That is not only a missing feature — it breaks ADR-0025's "no removal on either side" symmetry for a format with no canvas: a person can leave feedback on a note that an agent has no way to answer.

## The surface

`wb_thread_edit`, document-scoped, three ops:

- `thread.add` — anchor (spatial or text-quote) + body, id minted when absent
- `message.add` — reply into an existing thread
- `thread.resolve` — close, and `resolved: false` reopens

**No removal of either kind**, deliberately, for the reason ADR-0025 gave and this decision carries into a second format: a verb one side has and the other does not lets an agent erase feedback a person can only close. A test asserts the op set is exactly those three, so adding a fourth is a decision rather than a slip.

Unlike `wb_canvas_edit` this validates no document **kind**: the layer is the same plane on every format, and what varies is the anchor, which its own union already carries.

## Two things the storage shape forced

- **`message.add` refuses a thread the document does not hold**, rather than accepting it. `writeThreadMessage` is a no-op there *by design*, because opening a container is the one write that cannot merge — two replicas creating one under the same key with no common ancestor keep only one side. A quiet accept would report success over a lost reply. That refusal is its own test.
- **Message ids are minted against the ids a thread actually holds**, not against its message count. A peer's reply that merged in leaves the two disagreeing, and reusing an id is an overwrite of someone else's message rather than a reply.

Errors follow `wb_canvas_edit`'s wording, and "Nothing was written" is literal here rather than copied: every op mutates the in-memory `LoroDoc` and the snapshot is saved once, after the last op.

## Reach

Registered in `create-server`, on the MCP surface under the **same per-document write lock** `wb_canvas_edit` uses (this reads the threads a document holds, mints ids against what it read, and saves the whole snapshot back), in `TOOL_PROFILES` as `MUTATING` (the layer has no removal, so nothing here can lose a conversation), and in both tool lists.

`thread` is registered as an ADR-0009 entity with its reason, in the same style as the other additions: `vocabulary.md` already makes the thread the annotation layer's anchored unit, with comments as the messages inside one. `comment` would name the message, not the thing this tool opens, replies to and resolves.

## Test plan

- **Red first**: four `server-core-node` cases against an absent module.
- `pnpm smoke:e2e` — extended to drive all three ops through a **real MCP client against a markdown note**, which is the format the tool exists for; reaching it on a canvas would prove nothing the existing comment ops do not. Exit 0, and it printed `[e2e] wb_thread_edit → thread.add / message.add / thread.resolve on a markdown note`.
- The smoke's own `tools/list` guard caught the two hand-synced expected-tool lists before I did; both updated.
- `pnpm test` — 1164 files / 11254 tests, exit 0.
- `pnpm build`, `pnpm -r typecheck`, `pnpm check:local` — all exit 0. `knip` flagged two exports nothing imports; dropped rather than registered, since the schemas are used in the file that declares them.

**One unrelated flake observed, recorded rather than waved through**: the first full `pnpm test` run failed `BrowserDocumentPage.delete-confirm.browser.test.tsx > focus moves into the dialog…` with the page rendering its error `role="alert"` instead of the editor (`Unable to find role="button" and name "More actions"`). It passed 8/8 in isolation and the repeat full run was 1164/1164. This branch touches six files, all in `server-core`/`mcp-server`, and cannot reach `apps/web` — so it is not this change's. Flagging it because it is a shape `integrator-flow.md` does not yet list: the page reaching its *error state* under full-suite load, which reads as a missing control rather than as a failed load.

Visual evidence: none — this is an MCP tool surface with no rendered output. Its verification is the e2e smoke line quoted above, taken through a real client.

https://claude.ai/code/session_016W8zfiQN2orUM1hJt5ygs2

---

## Summary by CodeRabbit

- **New Features**
  - Added `wb_thread_edit` for managing document-scoped comment threads.
  - Create threads anchored to quoted text, add replies, and resolve or reopen discussions.
  - Responses include thread IDs, message counts, and current resolution status.
  - Added support for using document threads alongside existing canvas editing capabilities.

- **Tests**
  - Added end-to-end coverage for creating, updating, resolving, and reopening document threads.


kamiazya:
@coderabbitai review

---

PR #1277, opened 2026-09-04
fix(canvas-render): cut a node label between characters a reader sees as one

kamiazya:
## Summary

`fitToWidth` kept the longest prefix that fits by walking `[...text]` — **code points** — so a cut landed inside a grapheme cluster and the canvas drew a fragment. The never-empty rule made it worse by keeping one code *point*: a box too narrow for a family emoji drew `👨` rather than the family.

Measured over the swept widths of a label carrying six of each class:

| class | widths that fragmented | fragment |
|---|---|---|
| ZWJ family | **36** | a lone `👨` |
| Devanagari | 18 | a bare `क` |
| keycap | 12 | a bare `1` |
| flag | 6 | a lone regional indicator |
| skin-tone modifier | 6 | `👍` without its tone |

Same defect class as the search-snippet one fixed in #1273, on a more visible surface — the label drawn on the canvas rather than the edge of a search-result row.

## Visual repro

Rendered in a real browser (`canvas-render-browser`), because the subject is emoji and resvg carries only the vendored Roboto — a figure through the export path would have shown tofu, not the defect. Each row is cut at a width that lands **inside** the cluster; that width is derived per case (everything before the cluster, plus its first code point) rather than eyeballed, since a width chosen by hand cuts before the cluster and shows two identical columns.

| class | label | before — cut by code point | after — cut by grapheme |
|---|---|---|---|
| ZWJ family | `Team 👨‍👩‍👧‍👦 ships` | `Team 👨` | `Team 👨‍👩‍👧‍👦` |
| flag | `Ship 🇯🇵 v2` | `Ship 🇯` | `Ship` |
| keycap | `Step 1️⃣ go` | `Step 1` | `Step` |
| Devanagari | `Plan क्षि x` | `Plan क्ष` | `Plan` |

The PNG is at `tmp/screenshots/label-cut.png` in the branch's working tree and is **not embedded here**: this environment has no `gh` (so no `--attach`) and no ImageMagick (so `compose-figure.mjs` cannot run). For this change the table is the same evidence — what the fix moves is *which characters survive*, and characters render in markdown. What the table cannot show: the skin-tone and combining-accent rows differ only subtly at these widths, so they are omitted rather than presented as if they demonstrated something.

## Why one expression here and thirty lines in the snippet case

Structural, and worth stating because it is the reason this one is cheap: **this cut is a prefix from the string's start**, so the segmenter always sees the true beginning. A snippet cuts at an arbitrary interior offset, where a bounded window loses the context regional-indicator pairing needs — and no safe window size exists (Unicode's PRI 215 background says a correct implementation must scan back to the run's start; Rust's `GraphemeCursor` encodes that as an unbounded `PreContext` protocol).

## The instrument came first, and it refused the obvious shape

Landed as its own commit, because **the existing bench could not see this change at all**: its labels are `n0`..`n479`, about 29px against a 200px box, so `fitToWidth`'s walk never ran. A before/after on it would have read as "free".

With the instrument in place, for 480 overflowing labels:

| shape | cost |
|---|---|
| baseline (code points) | 1.1 ms |
| grapheme walk, materialised | **12.1 ms** |
| grapheme walk, lazy | 5.5 ms |

5× on a path that runs per label is about a third of a 60fps frame — the wrong trade for a fragment at a label's edge. So the walk is **gated**: text with no code point at or above U+0300 cannot hold a cluster, so it walks code points. Final price, measured on the landed rows with orders alternated: **ASCII 1.2×, an emoji-bearing label 5.8×**, and only when the label already overflowed.

The gate's premise is checked exhaustively against the segmenter in the test — every code point in U+0000..U+02FF against nine predecessors — rather than reasoned about, because a joiner that slips through is a label that fragments with nothing red.

### A tighter gate is a dead end, and the rule file now says so

Finding that out cost real time. "Can this character join something" answers **yes for every precomposed Hangul syllable**, since one may follow a jamo L — so it flags ordinary Korean and buys nothing. Only the segmenter can say whether a string *has* a cluster, which is the work being avoided.

An earlier hand-rolled class list was worse than coarse: it missed **U+FF9E and U+FF9F**, the halfwidth katakana voiced marks, so it would have fragmented Japanese while claiming to be safe.

## Test plan

- [x] New `canvas-render-node` tests — a per-class sweep with a reachability floor, the exhaustive gate check, and the never-empty rule at grapheme granularity. Written **red first**: seven failed with real fragments before the change.
- [x] `pnpm test` passes locally — `canvas-render-node` 1096/1096, including **both exactly-pinned scoreboards** (`text-wrapping-quality`, `frame-containment-quality`); `pnpm -r typecheck`, `pnpm lint`, `pnpm knip` clean.
- [x] Manual verification completed — the figure above, rendered through a real browser.
- [x] E2E coverage — n/a; the contract lives at this layer.

### Mutation checks — each kills a different assertion

| mutation | what fails |
|---|---|
| revert the walk to `[...text]` | the fragment tests **and** the reachability floor |
| force the gate to `false` | the same — confirming the gate is the path to correctness, not decoration |
| revert the never-empty unit to a code point | only the whole-cluster case: `expected '👨' to be '👨‍👩‍👧‍👦'` |

The reachability tally counts the **opportunity** — widths whose cut fell next to a multi-code-point cluster — not the fragments. Counting fragments is zero once the fix is in, by construction, so it could never tell a sweep that reached the case from one that reached none. The first draft counted fragments and read `0`, which is how that was found.

## Checklist

- [x] Docs updated — `package-canvas-render.md` decision #13, whose never-empty contract was stated in code-point language. Its pinned context budget moved 77 → 78; raised rather than trimmed, with the reason recorded at the pin, since that guard exists to make crossing a bucket a decision rather than to forbid it.

https://claude.ai/code/session_01AAhbSNwp1kBKr74gCR2voo

---
_Generated by [Claude Code](https://claude.ai/code/session_01AAhbSNwp1kBKr74gCR2voo)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved canvas label truncation so text is cut at grapheme boundaries, preventing multi-code-point characters such as emoji, flags, and accented characters from being split.
  * Preserved the first complete grapheme when no text fully fits within the available space.

* **Tests**
  * Added coverage and performance benchmarks for mixed-script and grapheme-rich labels.


kamiazya:
Triaged both post-push findings; both were real and both are fixed in `a45d5b38`.

**CodeRabbit's CRLF finding — valid.** `mayCluster` gated on `charCodeAt(i) >= 0x0300`, and LF (U+000A) joins CR (U+000D), so a CRLF label took the code-point path and could be cut between the two halves.

The interesting part is not the miss but why the guard didn't catch it. The gate's docblock claimed *"nothing below U+0300 joins the character before it"* and cited a test that swept every candidate joiner `0x00..0x2FF` against **nine hand-picked predecessors**. That reads as exhaustive and is exhaustive in one direction only — `\r` was not among the nine, so the one join in the range was structurally unreachable by the check that existed to find it.

Sweeping both axes settles it: over all 768×768 pairs below U+0300 there is **exactly one** join, and it is `U+000D + U+000A`. The test now covers both axes in ~600ms (against 2.2s for the naive pairwise form) by probing each candidate against every sub-U+0300 predecessor in **one** segmentation — `a0 j a1 j …` holds one segment per character unless something joined, and a join can only *lower* the count. That admits a false positive when the candidate joins what *follows* it (which is exactly how `\r` shows up), so each suspect is then re-tested pairwise for the direction that matters.

The gate keys on **CR**, not LF: a lone LF is ordinary in a body and a lone CR is not, so keying on the leader leaves LF-only text on the cheap path.

Stated plainly, because the merge-risk note reads more severe than the effect: **the picture does not change.** Both halves are non-printing and measure zero, so a cut before or after the CR draws the same label. What changes is that the gate's claim is now true and the guard behind it can fail. Mutation-checked — dropping the CR clause turns the new `does not cut CRLF in half` case red.

**Mutation report — one of ten survivors was real.** `truncate.ts:24` `>= 0x0300` → `> 0x0300` is a genuine gap: a combining grave (U+0300 itself) is the gate's own boundary and nothing pinned it. `treats U+0300 itself as a joiner` now does, mutation-checked.

The other nine are equivalent, and the invariant behind that is now written next to the code rather than left as a list to re-triage every weekly run: **the gate has only one direction of error.** Answering `true` for text with no cluster costs a slower walk over the same units; answering `false` for text with one cuts it in half. So every mutant that makes the gate *more eager* — `< 0x0300`, `true`, `return true`, emptying the fast-path block, `i <= text.length` (`charCodeAt` past the end is `NaN`, which fails both comparisons) — produces identical output by construction. `ObjectLiteral → {}` on line 4 is equivalent for a different reason: `grapheme` is `Intl.Segmenter`'s default granularity. Line 96's `firstUnit === ''` is only ever read when the loop broke on its first iteration, at which point both branches hold the same unit.

**Docstring coverage (50%) — not acted on.** The two functions it counts as undocumented are the ones this PR added, and both carry docblocks longer than their bodies; the check is scoped to functions the diff touched and appears to be counting the arrow function in the bench fixture and a test-local helper.

Local gates: `canvas-render-node` 1098/1098, `pnpm -r typecheck` 0 errors, `pnpm lint` clean, `pnpm knip` clean, `dev-rules-budget` back under its pin at 78,970 chars (trimmed the rule prose rather than raising the pin a second time for one decision).

Visual evidence: none — CR and LF are non-printing, so there is no before/after to show. The figure in the PR body covers the change that does move pixels.

---

kamiazya:
Standing down on the `check` failure — it is not this PR's, and there is no fix to port.

**What is failing.** `check` → `pnpm audit --prod --audit-level=high`. All three attempts against `https://registry.npmjs.org/-/npm/v1/security/advisories/bulk` time out (`error (23)` → 10s → 60s → abort), burning ~4 minutes before the job exits 1. Nothing else in the job fails: `pnpm install --frozen-lockfile` succeeds and the supply-chain policy check passes 1595 entries in 5.2s.

**Why it is not this PR's.** Three independent observations, the last of which is the one that could have refuted the claim:

1. The diff is three files in `packages/canvas-render` plus a rule document. Nothing in it can reach an npm advisory lookup.
2. The same job **passed** on this branch at 01:00 (`437b7e58`), before the outage began.
3. **`main` is red for the same reason.** Run `33822259235` on `b52e20ca` — this PR's own base — failed at 00:33 in `check` with the identical abort-controller timeout stack. So the failure is on the base branch, not introduced here.

**No fix to port.** This is npm registry availability, not repository code. The only in-repo change that would go green is weakening or bypassing the supply-chain audit gate, which is well outside a truncation fix's scope and not something to slip in as a CI workaround.

**Not re-run.** I lack `rerun-failed-jobs` permission on this repository (403). Per the flake rules that is one re-run I would otherwise have spent here, and it has already reproduced identically on two separate commits.

**Everything else on this head is green**, on the merge of `1cde5efb` into `b52e20ca`: `test-unit (1)`/`(2)`, `test-jsdom`, `test-browser (1)`/`(2)`, `packaged-smoke`, `widget-smoke`, `sbom-npm`, `build`, `pr`, `pr-title`, CodeQL (all three analyses), WIP. `verify`, `dry-run-npm` and `dry-run-docker` show `skipped` only because they depend on `check`, so the authoritative `verify` gate has not run on this head — which is why this is a stand-down rather than a merge.

Also on this push: the branch now carries `origin/main` (4 commits), so CI is running against the real post-merge state, and both of this PR's commits are re-signed — the earlier `a45d5b38` was unsigned and would have shown Unverified.

I will keep checking back and merge once the registry recovers and `check`/`verify` go green.

---

PR #1293, opened 2026-09-04
feat(web): put the last two surfaces behind the render broker, and make a new kind fail the build

kamiazya:
## Summary

- The tree row's icon and the tab favicon were the two surfaces ADR-0027 declared `broker: 'not yet:'`. Both now ask the broker, and `RENDER_SURFACES` answers `through` for every surface it declares — the two reasons are **gone**, not reworded.
- The tree row was the last place still decoding a stored snapshot on the **main thread** — the same call `load-row-render.ts` stopped making, at the same measured cost, once per visible row. It hands the bytes over now.
- The favicon's outline was a `useMemo` in the page's render path, so every edit computed a shape the next edit threw away 150ms later. It is subscribed now, computed in the worker at a new `idle` priority.
- Manual verification then caught a defect no test could, and answering "is that fixed at the root?" turned up the same defect shape in two more places. Both sections below.

## What made this safe rather than fast

A cached picture's worst failure is being **wrong** rather than missing, so most of this diff is about the key.

**The pipeline is part of the key's identity.** The broker holds one map, and its path named only the document, kind, version and theme — so a tree row's outline and a list row's SVG named the same entry, and whichever arrived first answered the other in a type the caller has no reason to check. `pipeline` joins the key; `render-broker.ts` is generic over what a family answers with, and the cross-family isolation the cast rests on is asserted rather than assumed.

**A session can now name the state a picture of it would be a picture of.** `updatedAt` cannot serve — stamped per push, it names one value for a document that changed many times. `getFrontier()` answers loro's own id for the committed state. Two choices in it were measured, not argued:

- the **state** frontier, not the oplog's — a document checked out to an older version shows that version; and the state frontier does not move on a commit that changed nothing;
- it names the **committed** state, which is deliberately not the instant the canvas is published. `onChange` publishes immediately and commits on a debounce, so a key read at render time files the new picture under the old version. Measured: reading the frontier straight after `onChange` returns the pre-edit value while `getCanvas()` already shows `hello` → `edited`.

That last measurement is why the favicon is driven by the document's change notification rather than by React state.

## Where the cost went

Exporting a snapshot to hand the favicon's document to the worker, against reading the canvas out on the main thread instead. Real browser, medians of 21 interleaved rounds:

| nodes | `doc.export({snapshot})` | `readSpatialCanvas(doc)` | bytes |
|---|---|---|---|
| 12 | **0.10 ms** | 0.50 ms | 1.5 KB |
| 40 | **0.10 ms** | 0.80 ms | 3.6 KB |
| 120 | **0.20 ms** | 1.90 ms | 9.6 KB |

The handover is 5–9× cheaper than the alternative and barely grows with the document — a release rather than a relocation. The tree row's removed decode is the figure from #1275: 1.20 / 2.60 / 4.60 ms at 12 / 40 / 120 nodes, once per visible row.

`idle` is a third priority band below `background`. Its slot budget is over **all** the deferrable bands together rather than one cap per band, and that is correctness rather than tidiness: counted per band, `idle` is free to take the very slot `background`'s cap reserves for an interactive request. Measured on a two-worker fleet before the fix — an idle request arriving between two background ones ran immediately, and both background requests then waited.

## What manual verification caught, and what your question caught after it

Every gate was green and the favicon still never changed however much was typed into a markdown document. Browser-mode markdown documents live in `useMarkdownDocument`'s own LoroDoc and never reach `DocumentSyncSession`, so asking the session alone answered `null` for all of them. **No test could have found it** — every test of the hook injects its source and so never makes that choice. The discriminator was a reload: the typed text persisted, which ruled out "the document never changed". After the fix, 1690 → 1294 bytes of favicon.

That fix was a branch, not a root cause. Adding a third `DocumentKind` and typechecking says so exactly:

| | before | after |
|---|---|---|
| `render-surfaces.ts` | ✅ errors | ✅ |
| `DocumentEditorSurface.tsx` | ✅ | ✅ |
| **`outline-source.ts`** | ❌ **silent** | ✅ |
| **`load-row-render.ts`** | ❌ **silent** | ✅ |
| **`load-row-outline.ts`** | ❌ **silent** | ✅ |
| `local-files-source` + 2 others | ⚠️ incidental mismatch | error gone |

`kind === 'markdown' ? … : <the other one>` reads as a complete decision and is a default: a new kind would have been drawn by the spatial pipeline in two places and asked an owner that does not hold it in a third. Each is now an exhaustive switch whose default takes the value as `never`. Two hand-written parallel unions (`WorkspaceDocumentEntry.kind`, `RenderKeySubject.kind`) were why the row loaders could not see a new kind at all — that is the drift this repo's Zod discipline exists to stop, and its effect was to move the error somewhere unhelpful. The error now lands at the decision, and the incidental ones are gone.

## Test plan

- [ ] `pnpm test` passes locally — **the browser project does not complete in this container any more; see below**
- [x] Manual verification completed (described above)
- [ ] E2E coverage added or extended where applicable — nothing here depends on real routing, persistence or the daemon

- `pnpm test --project web-jsdom` — **3503 passed**
- `pnpm --filter @kamiazya/whiteboard-web typecheck`, `pnpm lint`, `pnpm knip` — clean
- New: `layout-worker-outline.browser.test.tsx` (all three request shapes against the real worker and its lazily imported WASM), `outline-source.test.ts` (the branch manual verification found), plus the `idle` band, the pipeline axis, the cross-family isolation, and the frontier's two directions
- Mutation-checked: the `idle` band three ways, `getFrontier` three ways, the pipeline axis, the outline key — each fails a different assertion, baseline and restore green

**The browser suite is not green locally and I am not claiming it is.** It completed twice today at **194/194 files, 1037 tests**, then stopped completing. Five runs since, with a different set of files aborting each time and no overlap between them:

| run | files that aborted |
|---|---|
| 1 | `touch-pan-zoom` |
| 2 | `comment-compose-look`, `delete-confirm` |
| 3 | `backend-identity`, `canvas-settings`, `select-none` |
| 4 | `canvas-embed-lod`, `comment-placement`, `node-symbol-menu` |
| 5 | `touch-pan-zoom` |

The errors are `Browser connection was closed while running tests` and `[birpc] rpc is closed` — the Chromium process dying, not assertions — and the run aborts at 116–140 of 194 files. A real regression fails the same test every time. Per this repo's own rule I stopped re-running rather than fishing for a green. CI on a fresh runner is the authority here.

## Visual evidence

Visual evidence: none — `compose-figure.mjs` needs ImageMagick, which is absent in this container (it is also what stops `pnpm check:local` completing here). The user-visible change is a 32px tab icon and a 24px row icon; the measurement in its place is the favicon's own bytes before and after a real edit in the running app, 1690 → 1294, with the pre-fix pair 1690 → 1690 as the control.

## Checklist

- [x] Docs updated — ADR-0027's path grammar, its status, the `idle` band as decision 6a, and the version-key section rewritten now that the open document has one
- [x] No hand-written TypeScript interfaces duplicate a Zod schema — this PR **removes** two that did

## Follow-ups, named rather than assumed

- `whiteboard:doc_changed` never fires in browser markdown mode (`dispatchIdentityEvent` returns early without a workspace identity), so `useDirtyState`'s save dot is inert there too. Pre-existing, not fixed here; the hook no longer depends on that event alone.
- OPFS persistence still needs a version key on the **list** surfaces — `documentSummarySchema` carries no frontier. The open document now has one.
- The worker pool's cap of 4 rests on a premise measured false in #1275.

---

### DIFF
--- a/apps/web/src/components/markdown-editor/rail-write-mode.browser.test.tsx
+++ b/apps/web/src/components/markdown-editor/rail-write-mode.browser.test.tsx
@@ -1,7 +1,9 @@
 import { cleanup, fireEvent, render, waitFor } from '@testing-library/react'
+import { useCallback, useMemo } from 'react'
 import { afterEach, describe, expect, it } from 'vitest'
 import { useDocumentOutline } from '../../hooks/useDocumentOutline.js'
 import { useMarkdownOutline } from '../../hooks/useMarkdownOutline.js'
+import { createInTabRenderBroker } from '../../lib/render-broker.js'
 import { MarkdownEditor } from './MarkdownEditor.js'
 
 const SHORT = '# One\n\nJust a line.\n'
@@ -98,10 +100,16 @@ describe('the rail in write mode', () => {
 // starts another fleet of workers — measurably enough to destabilise the
 // whole project.
 function OutlineProbe({ body }: { body: string }) {
+  const broker = useMemo(() => createInTabRenderBroker(), [])
+  // A version that moves with the text, which is what the real source's
+  // frontier does — without one the memo would answer the second body from
+  // the first body's entry, which is the case these tests are about.
+  const readSource = useCallback(() => ({ frontier: `len-${body.length}`, body }), [body])
   const rects = useDocumentOutline({
+    documentId: 'outline-probe',
     kind: 'markdown',
-    canvas: { nodes: [], edges: [] },
-    markdownBody: body,
+    readSource,
+    broker,
   })
   return <div data-testid="outline-count">{rects.length}</div>
 }
--- a/apps/web/src/components/markdown-editor/rail-write-mode.browser.test.tsx
+++ b/apps/web/src/components/markdown-editor/rail-write-mode.browser.test.tsx
@@ -108,6 +108,7 @@ function OutlineProbe({ body }: { body: string }) {
   const rects = useDocumentOutline({
     documentId: 'outline-probe',
     kind: 'markdown',
+    revision: body,
     readSource,
     broker,
   })
--- a/apps/web/src/components/workspace-files/WorkspaceFilesPanel.tsx
+++ b/apps/web/src/components/workspace-files/WorkspaceFilesPanel.tsx
@@ -212,7 +212,7 @@ export function WorkspaceFilesPanel({
   // The row-size rendition. Separate from `loadRender` on purpose: it asks
   // the worker for block geometry rather than a serialized SVG, which is
   // both cheaper and the only thing legible at 24px (see DocumentMinimap).
-  const loadOutline = useMemo(() => createRowOutlineLoader({ source }), [source])
+  const loadOutline = useMemo(() => createRowOutlineLoader({ source, broker }), [source, broker])
 
   const readList = useCallback(() => source.listDocuments(), [source])
 
--- a/apps/web/src/components/workspace-files/document-entry.ts
+++ b/apps/web/src/components/workspace-files/document-entry.ts
@@ -3,6 +3,9 @@
  * needs it. It lives here rather than with any one pane because three of
  * them read it and none of them owns it.
  */
+
+import type { DocumentKind } from '@kamiazya/whiteboard-model'
+
 export interface WorkspaceDocumentEntry {
   readonly documentId: string
   readonly path: string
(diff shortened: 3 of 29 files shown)


kamiazya:
Two claims in this PR were wrong. Both were consequences I asserted rather than read, so recording them here rather than quietly editing the body.

**1. "so `useDirtyState`'s save dot is inert there too"** (follow-ups, first bullet — and the same sentence went into `useDocumentOutline.ts`'s module comment).

`useDirtyState` has exactly one call site, `DaemonDocumentPage.tsx:529`. It is not mounted on the browser page at all, whose save chip is `mergePersistence(renderState.persistence, markdownDoc.saveState)` — a signal `whiteboard:doc_changed` never touches. The first half of the bullet stands: `dispatchIdentityEvent` returns before dispatching unless both identity fields are set, and `BrowserDocumentPage` passes no `identity` to `useDocumentSync`. The event really never fires there; it just reaches no consumer, which makes it a latent gap rather than the visible symptom I claimed.

The source comment is corrected in #1307.

**2. "`RENDER_SURFACES` answers `through` for every surface it declares — the two reasons are gone rather than reworded."**

`editor-preview-pane` and `version-thumbnail` still carry `not yet:` reasons, and both did before this PR. What is true is the narrower sentence before it: the two surfaces this PR set out to move are `through` now. The commit message is merged and cannot be edited, so this comment is the correction.

Neither changes what the diff does. The measurements in the body — the handover table, the `idle` band, the favicon's 1690 → 1294 bytes — were taken and stand.

---

PR #1327, opened 2026-09-04
fix(web): docs snapshots resolve the workspace id before rendering, and the images catch up

kamiazya:
## Summary

Fixes the filed ordering bug `issues/docs-snapshots-broken-on-main`: `pnpm --filter @kamiazya/whiteboard-web docs:snapshots` failed 5/8 on a clean main checkout, so the docs images could not be regenerated since the ADR-0023 boot changes — and the document-browser images predated the picker redesign.

Three independent causes, all in the snapshot harness (no production code touched):

1. **Boot order**: the docs-snapshots project was the ONE vitest project whose setup did not seed the workspace-identity test seam, so every `BrowserIndexPage` capture threw `browser workspace id read before resolveBrowserWorkspaceId() completed`. `_setup.ts` now seeds a FIXED id (byte-stable regeneration) with the `default` segment, mirroring `vitest.setup.ts` / `browser-setup.ts` / `isolated-whiteboard-db.ts`.
2. **Retired chrome**: three canvas captures waited for a workspace name the TopBar stopped rendering when the title moved into the caller's `titleSlot`. The frame now passes a plain-span slot (production passes `DocumentProperties`) and the wait targets the document title — which still arrives via `/names`, so the wait still proves that fetch chain settled before capture.
3. **Determinism, found by regenerating twice**: the list capture races the UI webfont (measured: the card titles were the only differing pixels between two runs) — it now awaits `document.fonts.ready`; and both `BrowserIndexPage` fixtures give their workspace row the `default` segment a real fresh database mints, because the heading otherwise showed a raw ULID no user ever sees.

`docs/assets/*.png` regenerated: the document-browser images now show the shipped picker redesign (kind icons, pin glyph, aspect-video thumbnails, `default` heading). Byte-stability, measured honestly: 2/2 identical in this session's back-to-back runs; the review gate's QA lane ran 6 more independent trials and found 4/6 identical, with the misses being sub-pixel font-rasterization jitter on card-title glyphs (862 of 829k pixels, 0.1%) — the residue `document.fonts.ready` narrows but does not fully close under load. No CI job gates on regeneration determinism today, so this is a known residue, not a flake source; noted in the tracking issue's close-out.

## Visual repro

Visual evidence: the regenerated `docs/assets/*.png` ARE this diff — GitHub renders each image's before/after inline in the Files view, which is a richer figure than any composition of it (and the owner asked to skip separate figure attachments).

## Test plan

- [x] `pnpm --filter @kamiazya/whiteboard-web docs:snapshots`: 8/8 (was 3/8), run twice back-to-back with a clean tree after the second run.
- [x] Each failure mode was reproduced on clean main first (identical 5 failures), so the fixes target pre-existing breakage, not something this branch introduced.
- [x] typecheck + biome clean. No production source touched — `docs-snapshots/` fixtures, `_setup.ts`, `_helpers.ts`, `_top-bar-frame.tsx`, and the PNGs.

https://claude.ai/code/session_01PLiRSAKQA8rRW7j6XBJhAE


kamiazya:
Superseded by #1326, a parallel session's fix for the same filed issue (issues/docs-snapshots-broken-on-main) that landed on main mid-flight — same seam seed in _setup.ts, same titleSlot approach in _top-bar-frame.tsx, images regenerated. The only unique delta here was a `document.fonts.ready` wait before the list capture; the review gate's QA lane measured it narrows but does not close sub-pixel font-rasterization drift (4/6 identical regenerations under load), #1326 explicitly treats such drift as ordinary, and no CI job gates regeneration determinism — so the line is not worth a PR of its own. Branch deleted.

PR #1372, opened 2026-09-05
test(canvas-render): rule-reachability guard becomes per-rule witnesses, ending its second flake

kamiazya:
## What

`flake-watch` reported `edge-rules.properties.test.ts` red on **three main runs in 14 days** (33321928991 / 33313408933 / 33304624859), each the same way: `border-tracing` contributed fewer than `SELF_CONTRIBUTION_FLOOR=5` times in an unseeded 600-draw sample.

The guard has now flaked in **both** of its statistical forms — "at least once" before (fixed by densifying the generator + a measured floor), the floor after. So the sampling goes entirely: each rule gets one hand-built **witness input** that provably makes it contribute through the composed `selfPenalty` call (pair-only rules assert through `pairTerm`), and the witness table is keyed by rule name so a **new rule still fails until its witness exists** — the forcing function the sampled floor was for, kept without the sampling.

The trade is stated in the test comment: a generator that stops reaching a rule now weakens the properties silently instead of failing this guard loudly — accepted, because the loud failure was wrong three times and true zero times.

## Verification

- Mutation-checked: sliding the border-tracing witness rect off the path fails red naming the rule (`border-tracing: its witness input scores zero through the composed call`).
- `canvas-render-node` full project 1101/1101; typecheck clean.

Visual evidence: none — test-only change; the mutation check above is the evidence.

https://claude.ai/code/session_018sV1euXVHjCdB6MuPgWc3D

## Summary by CodeRabbit

* **Tests**
  * Replaced statistical sampling with deterministic test cases covering each penalty rule.
  * Added validation to ensure every penalty rule has a corresponding witness case.
  * Added explicit handling for rules that operate only on paired inputs.

### DIFF
--- a/packages/canvas-render/src/layout/edges/edge-rules.properties.test.ts
+++ b/packages/canvas-render/src/layout/edges/edge-rules.properties.test.ts
@@ -331,36 +331,105 @@ describe('PENALTY_RULES: each rule writes only its declared tier slot', () => {
    * The tier-placement properties compare a composed slot against the rule's
    * own term, so a domain where every term is zero compares zero to zero and
    * a composition that never wrote a slot passes. This is the reachability
-   * half: each rule has to be seen contributing something, at least once,
-   * through the same composed call.
+   * half: each rule has to be seen contributing something through the same
+   * composed call.
+   *
+   * A hand-built WITNESS per rule, not a sampled floor. This guard has been
+   * statistical twice and flaked both times: "at least once over 600
+   * samples" failed on a domain nobody changed, and the floor of 5 that
+   * replaced it (set from measured counts of 16-31) still fell short on
+   * three main runs inside two weeks — `border-tracing` landed under 5 in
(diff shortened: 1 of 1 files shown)


kamiazya:
Correction for the record: the three red runs this PR cites (33304624859 / 33313408933 / 33321928991) are all 2026-08-30, **before** #1146 (the denser-generator + floor-5 fix) merged at 17:37 that day — the flake census had already classified them as stale window entries, and the floor-5 guard itself never fell short on CI. The PR's claim that the floor "still fell short on three main runs" is wrong; flake-watch's 14-day window outlived the fix, and this lane was launched without applying the census's own rule (compare each run's timestamp against the fix's mergedAt).

The change stands on its remaining merit: the witness table closes the statistical class structurally (nothing samples any more), so the guard cannot re-enter any future flake window, and a new rule still fails until its witness exists. But the urgency was misread, and this comment is the correction.

PR #1392, opened 2026-09-05
feat(web): the picker keeps a lane of what this device opened recently

kamiazya:
The last deferred item from the 2026-09-05 picker redesign, unblocked by the owner's decision on the question that was holding it: **recency is kept per device and not synced.**

"Recently opened" is a record of what happened on *this* device, so a phone and a desktop holding different answers is correct rather than a gap — and keeping it out of the document model means no schema change and identical behaviour whether the workspace is kept by the browser or by the daemon.

## Visual repro

before — reopening means scanning a path-sorted grid; after — the last two opened, one tap, at a fixed position

Look at the **grid below the lane**: `design` and `notes` are in the same positions on both panels. That is the design claim — an addition beside the grid, never a reordering of it (NN/g: reordering destroys the spatial memory the path sort exists to give).

> The image is **not uploaded**: `gh` on this machine is 2.97.0 and `--attach` landed in 2.99.0, so the reference above resolves only in a local checkout. The panel digests below are what tells you the two panels are genuinely different renders, and `compose-figure.mjs` refuses identical ones.

What the figure cannot show: the tiles carry the kind icon rather than a document thumbnail, because the fixture's source produces no render. In the app each tile draws through the same `DocumentThumbnail` the cards below it use. Panels `90dff68a` / `06738804`.

## Design

- **Ids, not paths.** A rename must not drop a document out of the lane, and a path freed by a rename can be taken by a different document.
- **Resolved against the listing already in hand**, so a deleted document leaves the lane on its own — no reconciliation pass, no extra read.
- **One word of chrome.** The strip first shipped with only an `aria-label`; looking at the rendered figure is what caught it, since two unexplained tiles above the grid read as pinned documents or as a selection. Every surveyed product labels this lane (Finder and iOS Files "Recents", Figma "Recently viewed") — what this redesign cuts is instructional prose, not the noun that says what a thing is.
- **Absent, not empty**, when nothing is recorded; absent during a search, which has its own answer to "which one did you mean"; absent on a host that cannot open a document, since every tile would be inert.

## Two things the tests were built to catch, both of which fired

- Splitting "clear" and "load" across two effects made the **effect order load-bearing, and it was wrong** — the handle-keyed load ran first and the scope reset blanked it, so the lane never survived a remount. The browser test failed exactly there. Both halves now live in the scope-reset effect and each ends on the same value, so whichever runs last is right.
- The scoped-state ledger **refused a reset that writes a computed value**; it wants a real empty. That is the stricter and better rule, and following it is what made the fix above order-proof rather than order-dependent.

A test bug worth recording: `renderPanel(workspace = 'space')` used a **default parameter, which fires on an explicit `undefined` too** — so the "host passes no handle" case was silently exercising the `'space'` workspace. It surfaced only because the lane wrongly appeared.

## Verification

Mutation-checked, each mutation **asserted to have applied** before its run — the discipline exists because two mutations earlier in this session silently matched nothing and came back green:

| mutation | result |
|---|---|
| cap removed | 1 property red |
| dedupe removed | 2 properties red |
| append instead of prepend | 2 properties red |
| recording call removed | 2 browser tests red |
| empty-lane early return removed | 2 jsdom tests red |
| recency order → listing order | 1 jsdom test red |
| search gate removed | 1 browser test red |

`web-jsdom` and `web-browser` over `components/workspace-files` and `pages` (735 + 106 tests), plus typecheck, lint and knip. One unrelated failure in `BrowserDocumentPage.markdown-orphan.test.tsx` — the known full-suite flake already filed as `create-delete-node-full-suite-flake`; it passes in isolation and this diff does not touch `pages/` at all.

### DIFF
--- a/apps/web/src/components/workspace-files/RecentLane.tsx
+++ b/apps/web/src/components/workspace-files/RecentLane.tsx
@@ -1,0 +1,78 @@
+import { FileText, LayoutGrid } from 'lucide-react'
+import type { ReactNode } from 'react'
+import type { WorkspaceDocumentEntry } from '../../lib/document-entry.js'
+
+export interface RecentLaneProps {
+  /** The listing as loaded. Recorded ids are resolved against it. */
+  documents: readonly WorkspaceDocumentEntry[]
+  /** Most recent first, from `recent-documents.ts`. */
+  recentIds: readonly string[]
+  onOpen: (entry: WorkspaceDocumentEntry) => void
+  renderThumbnail?: (entry: WorkspaceDocumentEntry) => ReactNode
+}
+
+/**
+ * The documents this device opened most recently, as a strip above the grid.
+ *
+ * An ADDITION beside the grid rather than a reordering of it: the grid is
+ * sorted by path, and where a card sits is what a person remembers about it.
+ * Putting the newest first would move every card every time anything was
+ * opened, which is the spatial memory the sort exists to give (NN/g).
+ *
+ * Resolved against the listing already in hand, so a deleted or renamed
+ * document falls out on its own — the lane needs no reconciliation pass, and
+ * a stale id costs nothing but its own absence.
+ */
+export function RecentLane({ documents, recentIds, onOpen, renderThumbnail }: RecentLaneProps) {
+  const entries = recentIds
+    .map((id) => documents.find((each) => each.documentId === id))
+    .filter((each): each is WorkspaceDocumentEntry => each !== undefined)
+
+  // Nothing recorded, or nothing recorded that still exists: the lane is
+  // absent rather than empty. An empty strip is a row of chrome explaining
+  // that there is nothing to show, which is the wordiness this redesign is
+  // removing everywhere else.
+  if (entries.length === 0) return null
+
+  return (
+    <section aria-label="Recently opened" data-testid="recent-lane" className="px-2 pb-2">
+      {/* One word, and it earns its place: without it the strip reads as two
+          unexplained tiles above the grid — pinned, maybe, or a selection.
+          Every surveyed product labels this lane (Finder and iOS Files
+          "Recents", Figma "Recently viewed"); what this redesign is cutting is
+          instructional prose, not the noun that says what a thing is. */}
+      <h2 className="text-muted-foreground pb-1 text-[11px] font-medium">Recent</h2>
+      {/* Its own scroller: a phone fits two tiles, and the page body must
+          never scroll sideways to reach the third. */}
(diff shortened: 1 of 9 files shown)


kamiazya:
## Review lane findings, addressed

The `review.workflow.mjs` lane (12 agents) returned **1 confirmed finding, and it was real** — pushed as `bf1fe182`.

**A workspace handle named after a prototype member crashed the picker.** `readAll()` answers a plain object, so `all[workspace]` returned an *inherited* member for some handles. That value is truthy, so `?? []` never fired: the lane was handed `Object.prototype.constructor` and threw on `.map`, taking the whole panel down — with storage **empty**, so the person could not clear it, and on every mount.

The verify pass corrected the finding in both directions, and both corrections matter:

- **The named key is not reachable.** `__proto__` fails the segment charset (`[a-zA-Z0-9-]`, no underscore). What *is* reachable is `constructor` — it passes the charset, and `deriveWorkspaceSegment` lowercases a display name, so a workspace named "Constructor" mints it exactly. Same defect, same lines.
- **It is not prototype pollution.** `{ ...all, [workspace]: … }` uses a computed key, which *defines* an own property rather than invoking the `__proto__` setter, so nothing is ever written onto `Object.prototype`. A read-side confusion — a correctness bug, not a security one.

Guarded with `Object.hasOwn` in one accessor both entry points route through. Mutation-checked: removing it turns **14 of 23** cases red.

The regression tests cover the whole prototype class rather than the one reachable member, since what the charset allows is not that module's to know. `__proto__` is deliberately held to the weaker guarantee — never throws, always an array — and **not** to a round trip: zod's `z.record` rebuilds the parsed object by assignment, and assigning `__proto__` sets a prototype instead of defining an own property, so the entry cannot survive storage. Said in the test so the asymmetry does not read as an oversight.

## A wiring gap I closed in the same push

The panel's own suite supplies the workspace handle itself, so **nothing proved `BrowserIndexPage` passes one at all** — a gap every component-level test passes straight through, and exactly the `reachability` class. `BrowserIndexPage.recent-lane.browser.test.tsx` now drives the real page over real IndexedDB. Mutation-checked by making the page pass `undefined` (red); five fresh-process runs green.

## Verification after the fix

Full `web-browser` project **1124/1125** — the one failure is `BrowserDocumentPage.markdown.browser.test.tsx`, the known full-suite flake already filed as `create-delete-node-full-suite-flake`; it does not touch this diff. `web-jsdom` over `lib` + `workspace-files`: **1849/1849**.

CodeRabbit reported its rate limit rather than a review on this PR, so the AI-review surface here is the local lane above.

PR #1400, opened 2026-09-05
feat(web): the picker gets a selection mode, whose one verb is bulk delete

kamiazya:
The last deferred item from the 2026-09-05 picker redesign. Its blocker was that bulk actions did not exist; the owner scoped them to **bulk delete only** — the narrowest useful surface, since the trash already provides the undo, so the mode can be judged on one verb rather than a family.

Entered from the card menu, the object surface both pointers already reach: right-click on a desktop, long-press on a touch screen. Finder and iOS Files put it there too, and it costs no new affordance on the card.

## Visual repro

before — deleting five means five menu-Delete-confirm cycles; after — five toggles and one confirmation

> Not uploaded: `gh` here is 2.97.0 and `--attach` landed in 2.99.0, so the reference resolves only in a local checkout. Panels `27beb260` / `80e7e8de`; `compose-figure.mjs` refuses identical ones.

In the after panel: the bar with the count and its two verbs, an **empty circle** on the unselected "Reading list" (so the affordance says "this is a toggle" before anything is chosen), and the filled check plus derived border on the two selected.

## Two decisions worth reviewing

**A single selection borrows the singular confirmation**, which can name the document. Not a shortcut — it removes number agreement from the copy entirely, so the two new plural sentences in `destructive-copy.ts` are plural-only and branch-free, which is the property that module exists to have.

**The selection is not cleared when Delete is pressed.** The page confirms first, and a cancelled confirmation must return the person to what they chose. What clears it is the re-read after a delete that landed: the pruning effect drops every path that left the listing, so a **partial failure leaves exactly the survivors still selected**, with no outcome plumbed back through a prop. Both pages loop sequentially and report `N of M could not be deleted` rather than closing on a half-done job.

## Three things a test caught, all of them mine

- **The first double-click test asserted nothing.** Its own first click deselected the only selected card, ending the mode, so the double-click that followed opened correctly. It now uses a second card and says why in the test.
- **The pruning property was declared in the design and not implemented.** The test written for it failed — which is what a property is for.
- **`toggle-state-surface` rejected the card outright.** It announced `aria-pressed` while drawing the check from a separate `selection.has()` expression — the doubled shape that guard forbids. The look is now derived from the attribute through `group-aria-pressed:`, so the picture cannot disagree with what is announced.

And one **user-facing regression** the existing suite caught: reworking the delete handler into a loop changed the single-failure sentence from "Failed to delete the document from this browser." to a shorter one. Restored.

## Scope ceiling, stated rather than left silent

Selection is offered in the folder-contents **grid only** — not in search results, not in the one-column tree. The grid is the browse surface where tidying happens; the other two multiply the interaction surface without changing whether the feature works.

## Verification

Mutation-checked, each mutation **asserted to have applied** before its run:

| mutation | result |
|---|---|
| click no longer routes to toggle | 6 red |
| pruning removed | 1 red |
| single selection no longer borrows the singular delete | 1 red |
| `aria-pressed` always absent | 2 red |
| dialog ignores the count | 1 red |

`web-jsdom` **3813/3814** and `web-browser` **1131/1132** before the merge of current main; after it, the touched area re-ran **759/759**. Each single failure was the known `BrowserDocumentPage` full-suite flake already filed as `create-delete-node-full-suite-flake` — both pass in isolation, and this diff touches no document page. Typecheck, lint and knip clean.

### DIFF
--- a/apps/web/src/components/document-list/DeleteDocumentDialog.tsx
+++ b/apps/web/src/components/document-list/DeleteDocumentDialog.tsx
@@ -13,8 +13,16 @@ import { DESTRUCTIVE_COPY, type DestructiveActionId } from '../../lib/destructiv
 import { kindNoun } from '../../lib/kind-noun.js'
 
 export interface DeleteDocumentDialogProps {
-  // The document pending deletion, or null when the dialog is closed.
-  pending: { displayName: string; kind?: DocumentKind } | null
+  /**
+   * The document pending deletion, or null when the dialog is closed.
+   *
+   * `count` makes the subject a NUMBER rather than a name, for a bulk
+   * delete. Its absence is the singular case, so no existing caller changes
+   * and there is one rule rather than two competing subjects. A count of one
+   * never arrives: the panel routes a single selection to the singular
+   * confirmation, which can say which document it means.
+   */
+  pending: { displayName: string; kind?: DocumentKind; count?: number } | null
   busy: boolean
   error: string | null
   /**
@@ -51,10 +59,18 @@ export function DeleteDocumentDialog({
       <AlertDialogContent>
         <AlertDialogHeader>
           <AlertDialogTitle>
-            {pending ? `Delete "${pending.displayName}"?` : 'Delete canvas?'}
+            {pending === null
+              ? 'Delete canvas?'
+              : pending.count === undefined
+                ? `Delete "${pending.displayName}"?`
+                : `Delete ${pending.count} documents?`}
           </AlertDialogTitle>
           <AlertDialogDescription>
-            {DESTRUCTIVE_COPY[action](kindNoun(pending?.kind))}
+            {/* A bulk delete spans kinds, so the noun cannot come from one
+                document's kind the way the singular subject's does. */}
+            {DESTRUCTIVE_COPY[action](
+              pending?.count === undefined ? kindNoun(pending?.kind) : 'documents',
+            )}
             {error && <span className="mt-2 block text-destructive">{error}</span>}
           </AlertDialogDescription>
         </AlertDialogHeader>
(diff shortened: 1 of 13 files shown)


kamiazya:
## Review lane findings, addressed

The `review.workflow.mjs` lane (17 agents) returned **3 confirmed findings — all on the same branch of the same feature, and all real.** Pushed as `56327c6c`.

### 1. The retry re-sent DELETE for paths that had already succeeded (HIGH)

Neither page narrowed `pendingDelete.paths` after a partial failure, so pressing Delete again re-ran the whole list. On the daemon that genuinely diverges: its DELETE answers **404 for an absent path** (unlike the browser index, which no-ops), so every already-deleted path failed on the second attempt and the "N of M" progress went *backwards*. Both pages now narrow to the failures.

A lone survivor also gets its **name** back rather than staying `"2 documents"` — the count of the *attempt* is not the count of what the dialog is now offering to do, and `Delete "2 documents"?` is what it read before.

### 2. The daemon page never refreshed on that branch (HIGH)

`closeDeleteDialog` is its only caller of `loadWorkspace`, and the partial-failure path returns before it — so deleted rows stayed on screen behind a dialog saying they had gone.

The comment claiming otherwise **was copied from the browser page**, where it is true because that page refreshes before its own return. A comment that is true in one file and false in the file it was pasted into is worse than no comment; both now describe their own branch.

### 3. The daemon page's whole loop had no test (HIGH)

The browser page got an end-to-end bulk test and the other keeper got nothing — exactly the parity gap this repo keeps a ledger about. Three tests now drive it through its own entry point: the multi-path loop, the partial failure with its refresh and narrowed retry, and a 404 inside a bulk.

## One change tried and reverted

Counting a 404 as "already gone". It failed the existing single-delete test, and rightly — that page deliberately shows the daemon's message for a failed delete, and the same operation meaning two different things depending on selection size is worse than a pessimistic count. **Narrowing alone fixes the divergence the finding was actually about**, and is the smaller change.

## Verification after the fix

Mutation-checked, each asserted to have applied: browser narrowing (1 red), daemon narrowing (1 red), daemon refresh (1 red).

`web-jsdom` **3821/3821** and `web-browser` **1132/1132** across all 210 files — no flake this run, unlike the two earlier runs on this branch. Typecheck, lint and knip clean.

CodeRabbit reported its rate limit rather than a review on this PR (the third in a row), so the AI-review surface here is the local lane above.

PR #1406, opened 2026-09-05
feat(web): a card says when its content moved since this device opened it

kamiazya:
The last deferred item from the 2026-09-05 picker redesign — and **the shape it ships in is not the one that was deferred.**

## Why the deferred idea was rejected rather than postponed

The original was a freshness *ring*: age as a gradient. Three reasons from the initiative's own record:

- The non-verbal wins in slices 1–3 all rode **existing conventions** (tap to open, long-press, right-click, the file/grid kind icons). "How old" has no such mapping — a ring thickness or an opacity has to be learned, so it **adds a legend rather than removing words**, which is the opposite of what this initiative was for.
- It would have been built on `updatedAt`, which `document-entry.ts` records a measurement against: *a register one replica wrote that a merge does not consult*, observed leaving the stamp put while the content became a state nobody had written. A freshness signal on that field can call a document unchanged in the exact case worth showing.
- The recently-opened lane (#1392) already answers the question recency is *for* — "where was I" — by position, which is genuinely non-verbal.

## What ships instead

**Binary: did this change while I was not looking.** An unread dot needs no legend, and in this product the question is actionable rather than decorative — MCP agents write to documents behind the person's back, and until now nothing on the card said which one had moved.

Keyed on `contentDigest`, which both keepers already supply and which the same comment names as the reliable identity.

## Visual repro

before — an agent rewrote one of these and nothing says which; after — the one that moved since you opened it

> Not uploaded: `gh` here is 2.97.0 and `--attach` landed in 2.99.0. Panels `c540f156` / `091e860c`; `compose-figure.mjs` refuses identical ones.

**The figure is the argument.** Both panels read `2d ago` on every card — the age text cannot tell you anything — and only the dot distinguishes the document an agent rewrote.

## Design notes worth a reviewer's eye

- **Stored separately from the recency lane**, though both are written at the same moment. The lane is an ordered list capped at a handful; a digest riding on it would make the dot vanish for the 9th-oldest document for no reason a person could see.
- **Absence is a first-class answer in two places**, both tested: a document this device never opened has no baseline and stays *silent* rather than reading as changed, and a row the keeper gave no digest carries no dot rather than one that means nothing.
- **The age text stays** (owner decision): the dot says "changed while you were away", the text says "how old", and neither answers the other.
- **The dot carries a text alternative**, so nothing is conveyed by colour alone.

## One claim deliberately weakened

`Object.hasOwn` is here from the start after the lane shipped without it — but the failure it prevents is **milder than the lane's, and the commit says so rather than repeating that story**. `recordSeen` uses a spread, not `.filter`, so nothing crashes. Measured, exactly one case breaks without it: `readSeenDigest(ws, '__proto__')` after a record returns `Object.prototype` instead of a string, because zod's `z.record` rebuilds by assignment and cannot persist that key.

## Verification

Mutation-checked, each **asserted to have applied** before its run:

| mutation | result |
|---|---|
| cap eviction removed | 1 red |
| own-key guard removed | 1 red |
| newest write no longer wins | 4 red |
| the digest comparison always says changed | 4 red |
| no baseline recorded on open | 2 red |
| the dot never renders | 2 red |

`web-jsdom` **3870/3870** and `web-browser` **1136/1136** across 211 files. Typecheck, lint and knip clean.

### DIFF
--- a/apps/web/src/components/workspace-files/FolderContentsList.tsx
+++ b/apps/web/src/components/workspace-files/FolderContentsList.tsx
@@ -40,6 +40,13 @@ export interface FolderContentsListProps {
    */
   selection?: ReadonlySet<string>
   /**
+   * The documentIds whose content has changed since this device last opened
+   * them. Absent means the caller has no baseline to compare against, which
+   * is not the same as "nothing changed" — a fresh device shows no dots
+   * rather than dots on everything.
+   */
+  changed?: ReadonlySet<string>
+  /**
    * A card's picture. This component neither fetches nor renders, so a
    * caller with no daemon to read from still gets a working list — with the
    * kind label the list already carries.
@@ -75,6 +82,7 @@ export function FolderContentsList({
   onDocumentContextMenu,
   selectedPath,
   selection,
+  changed,
   renderThumbnail,
   className,
 }: FolderContentsListProps) {
@@ -180,6 +188,19 @@ export function FolderContentsList({
                   cannot disagree with what is announced. Presence is gated on
                   the MODE, which is a different question from the state.
                   `aria-hidden` because the button already says it. */}
+              {changed?.has(entry.documentId) && (
+                /* Binary, and in the corner opposite the selection check so
+                   the two never collide. `role="img"` with a name rather
+                   than a bare colour: a dot nobody can see or hear is not a
+                   signal, and this one contributes to the card's own
+                   accessible name the way the kind badge already does. */
+                <span
+                  role="img"
+                  aria-label="Changed since you last opened it"
+                  data-testid="card-changed-dot"
+                  className="bg-primary absolute top-1 right-1 size-2.5 rounded-full"
+                />
+              )}
               {selection !== undefined && (
                 <>
                   <Circle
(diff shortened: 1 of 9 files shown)


kamiazya:
## CI caught a real budget, and the review lane found nothing

**`file-size-budget` refused the panel** at 1237 lines against a recorded ceiling of 1196 — a shrink-only grandfather, so the guard offers two answers and asks for one deliberately: shrink it, or raise the ceiling.

**Shrunk** (`bc1368fc`). The panel had grown across three increments this session (the recency lane, selection mode, this dot), and the newest two are one subject: what *this device* remembers about a workspace. `useDeviceMemory` holds both — the lane's ordered ids and the changed-dot baselines — because they are written at the same instant, an open, and read by the same render.

They stay **stored** apart, which is the difference that matters and is not the hook's to flatten: the lane is capped at a handful, so a digest riding on it would lose its dot the moment the document fell off the end.

Panel: **1237 → 1166** lines, under the ceiling without moving it.

The hook is registered in `scoped-screen-state.test.ts`'s scan globs, or the state that moved into it would silently leave the ledger — the failure that file exists to prevent, one level up. It carries its own `// SCOPE RESET` block, which is how the ledger checks a screen whose state lives in a hook.

## Review lane

`review.workflow.mjs` (10 agents) returned **0 findings** — no confirmed issues across correctness, security, reachability, contract or test-coverage; QA's three scenarios (smoke, error-recovery, startup) all passed, including a runtime probe that forced each localStorage failure mode against the real module and confirmed it degrades silently rather than throwing.

Worth stating plainly: **that lane ran against `91970c4e`, before the extraction above.** The refactor is behaviour-preserving and covered by the same tests (jsdom 3871/3871, browser 1135/1136), but it has not been through an independent review pass.

## Verification after the extraction

`web-jsdom` **3871/3871**; `web-browser` **1135/1136** — the one failure is the known `BrowserDocumentPage.markdown.browser.test.tsx` full-suite flake, its fourth observation this session and already filed as `create-delete-node-full-suite-flake`; this diff touches no document page. `file-size-budget` green, lint and knip clean.

PR #1421, opened 2026-09-05
fix(server): canvas_view names what a canvas's own text nodes embed

kamiazya:
**Two subjects, and the reason they share a branch is a constraint on me, not a judgement:** this session may only push to `claude/wikilink-embed-support-wjpohz`, so the dependency work below could not be split off. Say the word and I will land them separately. The title is the shipped fix, because that is the half release-please must record as a patch bump; the dependency half is dev-only and needs none.

## 1. `canvas_view` dropped a canvas's own text-node references

The MCP Apps widget lays out one canvas and its tool allowlist is `canvas_view` plus `wb_canvas_edit`, so **nothing else ever reads that canvas's text-node bodies.** `loadReferenceGraph` already loaded what they name — `referenceTargets` scans text-node bodies — and this tool then threw it away, because it built its `references` payload by filtering `canvas.nodes` to file nodes.

So a text node's `[[note]]` drew as a bracket literal in the chat-inline widget, while the same canvas through `wb_scene_render` drew the note. Two tools disagreeing over one shared loader.

The payload is now keyed by `referenceTargets({ canvases: [canvas] })` — the one definition of what a reference is, rather than a node-kind filter. Seeded without `loaded`, so it still stops at what the canvas itself says and does not ship the transitive closure the graph also holds: the existing policy about a *referenced* body's own embeds is unchanged, it just applies to the right set now. Image file nodes stay out, as `referenceTargets` already excludes them.

Found by a wiring-gaps audit of the reference layer, and confirmed by reading both sides of the seam: `mount.ts` builds the widget's bundle with `referenceSeams(new Map(Object.entries(references)))`, so whatever this tool omits cannot be resolved downstream.

## 2. The three dev-only advisories the audit still reported

From a `dependabot-review` pass. **There are no open Dependabot PRs**, and **`pnpm audit --prod --audit-level=high`, the CI gate, was already clean and stays so.** Nothing here reaches the published package.

| advisory | severity | path | how it was cleared |
|---|---|---|---|
| GHSA-r292-9mhp-454m (`tar`) | high | cyclonedx → node-gyp | lockfile refresh, 7.5.19 → 7.5.22; no override needed |
| GHSA-2v37-7h3g-55p8 (`nanoid`) | high | vitest → vite → postcss | override raised to `<3.3.18: ^3.3.18` |
| GHSA-g7r4-m6w7-qqqr (`esbuild`) | low | stryker → vitest → vite | pinned by its exact vulnerable range to `^0.28.1` |

Two things that were wrong on first reading:

- **`nanoid` is a load-bearing published runtime dep, and was never in range.** The published package asks for `^6`, installed 6.0.1; the advisory is `<3.3.18`. The vulnerable copy was postcss's 3.3.16 alone, which is why the prod audit never saw it.
- **The 7-day `minimumReleaseAge` was not what held it back**, so no `minimumReleaseAgeExclude` entry is needed: 3.3.18 published 2026-08-07, a month outside the window. The real cause is that the existing `nanoid@<3.3.8` override no longer *reached* the advisory — 3.3.16 satisfies `^3.3.8` and does not match `<3.3.8`, so it sat under the override untouched.

`pnpm audit --audit-level=low` now reports no known vulnerabilities.

## Test plan

- [x] New or updated `mcp-node` / `mcp-jsdom` / `mcp-browser` / `web-browser` tests added — one `canvas_view` case seeding a text node with `[[notes]]` and asserting the payload carries its id, name and body. Red before the fix; **mutation-checked**, with the node-kind filter restored it fails on the reference being absent. The five existing reference tests all seeded through a *file* node, which is why nothing was red.
- [x] `pnpm test` passes locally — 184 files across `server-core-node` / `canvas-render-node` / `arch-lint-node`; 481 node-project files; `pnpm test:browser` 220 files / 1164 tests; `pnpm smoke:e2e` ALL OK; `pnpm -r typecheck`, `pnpm lint`, `pnpm knip`, `pnpm audit --prod --audit-level=high` clean.
- [x] Manual verification completed — installed versions read back out of the lockfile after each step, not inferred from the audit's silence.
- [ ] E2E coverage added or extended where applicable — `smoke:e2e` run as a regression check, not extended; `canvas_view` is already covered there.

### The two browser failures I saw, and why they are not this change

The first `test:browser` run with the dependency change failed two tests (`BrowserDocumentPage.delete-confirm` focus return, `BrowserDocumentPage.markdown` heading past the save debounce). Since esbuild and vite moved, that had to be ruled out by measurement:

| run | dependency change | tree | result |
|---|---|---|---|
| treatment 1 | applied | two audit subagents finishing alongside | 2 failed / 1164 |
| control | reverted, reinstalled | quiet | **220 files, 1164 passed** |
| treatment 2 | applied, reinstalled | quiet | **220 files, 1164 passed** |

It does not reproduce with the change on a quiet tree, and both files pass in isolation (28/28). That is the load-dependent `BrowserDocumentPage` + real-IndexedDB shape `.claude/rules/integrator-flow.md` describes and #1397 root-causes. Not introduced here.

The one other red locally is `local-node-version`, which reports this container's Node 22 against the pinned 24 and fails identically on a clean checkout; CI installs 24.

## Visual evidence

Visual evidence: none — the change is an MCP tool payload and a lockfile. What a reader would want to see is the widget resolving the embed, and that is asserted on the payload the widget builds its bundle from, in `canvas-view.test.ts`.

## Checklist

- [x] Docs updated if this changes user-visible behavior or a published contract — the tool's own comment now states which set it names and why; each override entry carries its GHSA and reason inline

https://claude.ai/code/session_01A8HzhgLpsT6XLvJNcz2fxD

## Summary by CodeRabbit

- **Bug Fixes**
  - Canvas views now include documents referenced through embedded or linked text, not only directly represented file nodes.
  - Referenced document details now correctly include the document ID, title, and Markdown content.
  - Canvas reference results are limited to items named by the canvas, avoiding unrelated transitive references.

- **Security**
  - Updated bundled components to address known security vulnerabilities.


kamiazya:
`test-browser (2)` failed on `7e5c7b6`, and **no test assertion failed in it.** Every test the log shows reported ✓, including this PR's own `draft-link-preview.browser.test.tsx`. The job died at teardown:

```
Unhandled Rejection
Error: [vitest] There was an error when mocking a module…
Caused by: Error: [vitest] Browser connection was closed while running tests. Was the page closed unexpectedly?
Serialized Error: { code: 'VITEST_BROWSER_CONNECTION_CLOSED' }
Caused by: Error: [birpc] rpc is closed, cannot call "resolveManualMock"
```

The browser page went away and an in-flight module-mock request could not be answered, so the run exited 1 with every test green — the shape `.claude/rules/integrator-flow.md` records as "vitest reports every test as PASSED while the file exits 1; the exit code is the only tell". That is runner loss, not this diff: the change is an MCP tool payload plus a lockfile, and `test-browser (1)` passed on the same commit.

The same suite was green on this branch's three previous heads and green locally twice at 220 files / 1164 tests.

**I could not spend the one permitted re-run**: `rerun-failed-jobs` answers `403 Resource not accessible by integration` for this session. Nothing in this PR's scope would make the failure more robust — the fix would be in the browser-runner setup, which is outside the diff and would widen it. So I am standing down on this one and keeping the PR watched until it is green, merged or closed.

---

PR #1425, opened 2026-09-05
feat(web): the browser keeper has variations, and `branches` stops being a capability

kamiazya:
## Summary

- **The browser keeper answers the branches seam.** `createBrowserBranchesBackend` runs `@kamiazya/whiteboard-history`'s pure steps over the workspace record's branch plane, so the rules are the daemon's rules rather than a second implementation. `BrowserBackend` gains `mutateRecord`/`readRecord` for it.
- **`branches` stops being a capability**, because six of this repo's own guards say it must. That was scheduled for a later increment and turns out not to be schedulable — see below.
- **What replaces the flag is a better answer, not a missing one**: whether to show the chip is now the backend's `hasBranches`, which is per DOCUMENT. A markdown body has no record-holding backend and so no variations; no provider-level flag could ever have said that.
- **`merge` stays a capability.** The keepers really do still differ there until the commit half lands, and the chip already separates the two (`mergeEnabled`) — so a browser user gets variations without a control that would throw.

## The regression this nearly shipped

The page has no `BrowserBackend` for a markdown document, or before one loads. My first wiring answered the context `null` in those states — and `null` in `BranchesBackendContext` means *fall back to the daemon*. That would have issued requests to a daemon that is not there, on every markdown document in browser mode, which is the single thing mounting a browser provider exists to stop.

So `createBrowserBranchesBackend` takes a **nullable** backend and always answers. `branchlessBackendContract` was therefore not dead as deleting it assumed — it changed subject, and now describes exactly that null case.

## Why the capability had to go, and why it is not a scope grab

Setting `BROWSER_CAPABILITIES.branches` true turned six guards red, all saying one thing:

```
× declares nothing both keepers agree on
× src/lib/branches-backend.ts names a capability the keepers really differ on
× agrees with the provider about branches
```

`provider.capability-reach.test.ts` and `keeper-parity.test.ts` refuse a flag both keepers set the same way, by name. `provider.ts` already records `workspaces` and `versions` leaving for exactly this reason. Retiring `branches` is not a decision this PR took on its own — it is what those guards mean, and the alternative was a state the suite refuses.

Consequences a reviewer should look at:

- The daemon page's `capabilities.branches` guards were **all always-true** and are removed; its "Variations" teaser could never render and is deleted.
- Two daemon-page cases asserted behaviour under `branches: false`. That state no longer exists: one is retargeted at `merge: false` and now asserts the chip **is** there; the other is deleted with a note, its subject covered by the positive case beside it.
- `keeper-parity`'s two branch entries move from `capability` to `both-keepers`, naming the browser module that answers.
- **`VersionTimeline`'s own `branches` flag stays**, and that is deliberate. It asks a different question — do the version ROWS carry a branch to lane by — and the browser's do not until automatic checkpoints land. Pinned as a decision rather than left as a leftover.

## Test plan

- [x] New or updated `mcp-node` / `mcp-jsdom` / `mcp-browser` / `web-browser` tests added — `branches-backend.contract.browser.test.tsx` runs the nine-case contract against the browser keeper over real IndexedDB (the daemon's run stays in `branches-backend.contract.test.ts`, so both keepers answer the same suite); `BrowserDocumentPage.branches.browser.test.tsx` drives the flow through the real page — chip in the header, create a variation, and it lands on the record, read back the way a replica would rather than through the backend that wrote it.
- [x] **Mutation-checked, four ways.** Forcing every read to the resting state fails 5 of 9 contract cases; dropping the writes fails 6 of 9; restored, 9 of 9. On the page flow, forcing `hasBranches` false fails with `Unable to find an element by: [data-testid="header-branch-chip"]` — precisely the before picture.
- [x] Suites for the area: **web-jsdom 374 files / 3918 tests** and **web-browser 218 files / 1162 tests**, both green. `pnpm -r typecheck` clean, `pnpm lint` clean (the 3 warnings are main's `docker-contract.test.ts`, outside this diff), `pnpm knip` clean.
- [x] Manual verification — the flow test above drives the real page in a real browser, which is the same path a person takes. The Cloudflare Pages preview on this PR serves browser mode, so the chip can be exercised there directly.
- [ ] E2E coverage added or extended where applicable — not applicable; the browser-mode flow has no routing, websocket or daemon dependency, and is covered end to end above.

## Visual evidence

Two panels, both the browser-mode header rendered through the real page in `web-browser`; the before panel taken with `hasBranches` forced false, which is the state this PR replaces. Distinct renders — `before.png` md5 `4fbb81491d1a691fe07b633c2f2bc28d`, `after.png` md5 `d6e133e2533430e79d35fbf6357017e4`.

- **before** — the header ends at the comments and overflow icons; there is no variation control at all.
- **after** — a `● Main ⌄` chip and its `⋯` kebab sit between them and the history icon. No merge entry point inside it, because `merge` is still false for this keeper.

**The image itself could not be attached from this environment**: there is no `gh` CLI here (so no `gh pr create --attach`) and the GitHub MCP surface has no upload endpoint. The figure was produced and read before writing this, and handed to the author directly. What a reviewer can check without it: open the preview link this PR's bot comment posts, in browser mode, and the chip is in the header — or read the two assertions in `BrowserDocumentPage.branches.browser.test.tsx`, which are what the panels show.

## Checklist

- [x] Docs updated if this changes user-visible behavior or a published contract — the capability's departure is recorded in `provider.ts` beside the two that went before it, which is where the next reader of that type will look. ADR-0022's addendum and the "daemon only" lines in DESIGN.md / the how-to belong with the merge half, and stay with it.
- [x] No hand-written TypeScript interfaces duplicate a Zod schema — the state is `packages/history`'s `documentBranchesStateSchema` throughout

---

## Summary by CodeRabbit

* **New Features**
  * Added browser support for branch management, including creating, removing, renaming, switching, merging previews, and loading branch documents.
  * Added browser workspace record read and write capabilities for branch operations.
  * Added fallback handling for workspaces without existing branch data.

* **Tests**
  * Added browser contract coverage for branch backend behavior.

### DIFF
--- a/.claude/rules/architecture-map.md
+++ b/.claude/rules/architecture-map.md
@@ -19,7 +19,7 @@ Package boundaries are cut by **runtime requirements**, not by feature. The shar
 | `packages/daemon-client` | the daemon's browser-safe client half: the `/api` Zod contracts the web app parses, the fetch/WS/SSE document backends, `api-client`, and the shared backend contract test suites. Extracted from mcp-server's `src/shared` so browser-safety is structural (this table scans it), not positional; consumed directly by both composition roots | model, server-core, history, zod, @opentelemetry (browser SDK set) |
 | `packages/canvas-viewer` | Read-only spatial-canvas scene viewer UI (renders canvas-render SVG), shared between `apps/web` and the MCP Apps widget | model, codec, render, `@modelcontextprotocol/ext-apps`, react, zod |
 | `packages/mcp-server` | Node composition root: CLI, stdio, local store impls, resvg, Inversify container | server-core + port impls |
-| `apps/web` | Browser composition root: Canvas API backend, IndexedDB store impls, read-write spatial canvas editor, markdown editor | loro-adapter, model, codec, render, canvas-viewer, ports, facet-engine, facet-ui, plugin-visual, search, workspace-index, daemon-client + port impls |
+| `apps/web` | Browser composition root: Canvas API backend, IndexedDB store impls, read-write spatial canvas editor, markdown editor | loro-adapter, model, codec, render, canvas-viewer, ports, facet-engine, facet-ui, plugin-visual, search, workspace-index, daemon-client, history + port impls |
 
 **A third-party dependency is judged by that criterion, not by a quota.** The
 `allowedThirdParty` lists in `architecture-map.ts` are a RECORD of what has
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -36,6 +36,7 @@
     "@kamiazya/whiteboard-facet-ui": "workspace:*",
     "@kamiazya/whiteboard-loro-adapter": "workspace:*",
     "@kamiazya/whiteboard-daemon-client": "workspace:*",
+    "@kamiazya/whiteboard-history": "workspace:*",
     "@kamiazya/whiteboard-model": "workspace:*",
     "@kamiazya/whiteboard-plugin-visual": "workspace:*",
     "@kamiazya/whiteboard-ports": "workspace:*",
--- a/apps/web/src/App.route-sync-stale-replay.test.tsx
+++ b/apps/web/src/App.route-sync-stale-replay.test.tsx
@@ -93,7 +93,6 @@ const { App } = await import('./App.js')
 const BROWSER_STATE: ProviderState = {
   kind: 'browser',
   capabilities: {
-    branches: false,
     merge: false,
   },
 }
--- a/apps/web/src/App.test.tsx
+++ b/apps/web/src/App.test.tsx
@@ -143,7 +143,6 @@ vi.mock('./pages/DaemonIndexPage.js', () => ({
 const BROWSER_STATE: ProviderState = {
   kind: 'browser',
   capabilities: {
-    branches: false,
     merge: false,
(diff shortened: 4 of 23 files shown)


kamiazya:
Note for anyone reading the Merge Risk banner above: it is pinned to `99f31a2`, which was this branch's head for a while by mistake — I built the wiring on a side branch and then pushed the *named* branch, which was still at the mechanic commit. The wiring is here now (`02e7a81`), and it is what addresses both of the concerns the banner raises:

- **"Browser branch variations may be shown for documents that should not support them."** Correct at `99f31a2`, where `hasBranches` was a hardcoded `true`. It is `backend !== null` at the current head, which is the per-document answer: a markdown body has no record-holding backend and so no variations, and `WorkspaceTopBar` gates the chip on that rather than on a keeper flag. `branchlessBackendContract` runs against exactly that state.
- **"The persistence contract test can fail intermittently before its workspace is ready."** Fixed in `bf088f7` — the thread is resolved. It waits on `readRecord` answering non-null, which is the same question the backend asks itself, rather than on a duration. `test-fixed-sleep-ledger` caught the same thing independently and is what turned CI red.

No action needed on either; the banner is just older than the diff.

---

PR #1448, opened 2026-09-06
feat(annotations): focus handoff, message counts, icon-first verbs and resolve motion

kamiazya:
## Summary

Four reported symptoms in the annotation layer, plus the two structural changes they turned out to need. Each commit is one increment with its own nearest-layer test.

- **Opening a conversation moves the reader into it, and Escape brings them back.** Pressing a gutter marker in the markdown source used to leave the caret in the body — the rail opened behind the reader. `use-comments-rail` now captures where focus came from on reveal/compose and returns it on Escape.
- **Every surface says how much is in a conversation, and the rail says when it last moved.** The rail already showed `N messages`; the source pane's gutter dot, the preview marker and the canvas pin did not, so one conversation was described three different ways depending on where you met it. `threadLastActivityAt` takes the max of `createdAt`/`editedAt` **by instant** (`Date.parse`), because `okfTimestampSchema` accepts both `Z` and `±HH:MM` and a lexical compare gets those wrong.
- **The rail's verbs go icon-first, and the status dot becomes the Resolve toggle.** DESIGN.md's "Object-action surfaces are icon-first" was written and the rail was never held to it — the same four verbs rendered icon-only on the canvas card and icon + label here. Both halves ship together: dropping the labels while keeping the old padding would take the width the label was giving the target and give nothing back, so every verb is 44px. Merging the dot into the verb means the press lands on the object that then changes, and makes Resolve reachable without opening the conversation.
- **Resolving moves a conversation instead of making it disappear.** The row crosses to the resolved look, holds long enough to be read, then leaves while the rows below glide into the gap. The write is issued immediately — only the presentation waits.
- **The preview marker measures its origin from the document, not from another marker.** Root cause: four call sites did `querySelector('svg')` inside the preview column, which finds a comment marker's own icon (markers live in that column and render before the pane). Every origin measured that way is wrong the moment the document has a conversation on it — the placement was measuring its own previous output, a constant +62px. One `previewDocumentSvg` definition now, held mechanically because the defect is one character wide.
- **The comment pin says how many messages its conversation holds.** Composed into the scene by `canvas-render`, so the widget, the export and `wb_scene_render` get it for the same reason they get the pin.

`apps/web/DESIGN.md` carries the reasoning, including two corrections to claims made about the canvas pin that measurement refuted.

## Test plan

Seven new test files, all at the nearest layer that can see the claim:

| file | project | what it holds |
|---|---|---|
| `use-comments-rail.focus-return.browser.test.ts` | `web-browser` | focus is captured on reveal/compose and returned on Escape |
| `comments-panel-focus.browser.test.tsx` | `web-browser` | the Escape ladder: edit → compose → return |
| `comments-panel-icon-first.browser.test.tsx` | `web-browser` | icon-only names, 44px boxes, dot-as-toggle, toggle is a sibling of the row |
| `comments-panel-motion.browser.test.tsx` | `web-browser` | the row crosses, holds, leaves; survivors glide; the write does not wait |
| `preview-marker-placement.browser.test.tsx` | `web-browser` | marker top vs the block's own rendered `<text>`, plus a scan banning bare `querySelector('svg')` |
| `comment-pin-count.browser.test.tsx` | `web-browser` | the editor hands the layout its threads, so the pin can carry a count |
| `thread-activity.test.ts` | `web-jsdom` | last-activity by instant, across mixed offsets |

Local runs on this branch (area suites; CI runs everything):

```
web-browser  (annotations)                   12 files, 79 tests passed
web-jsdom    (activity/parity/meta/state)     3 files, 52 tests passed
canvas-render-node (comments)                 1 file,  31 tests passed
pnpm -r typecheck                            0 errors
pnpm lint / pnpm lint:noconsole              clean
```

Manual verification: both figures below were produced by driving the real components in a real browser (`web-browser` project, real fonts, `page.screenshot`), once with the change and once with the touched source checked out at the merge base.

Two guards moved in the same diff, on the record: `file-size-budget.test.ts` raises `MarkdownEditor.tsx` 1021 → 1066 and `spatial-canvas.ts` 1840 → 1889, each with the reason beside it.

Three things measured rather than argued, because they changed what shipped:

- Both "defensive" focus guards written first were **unfalsifiable** — mutation checks passed in jsdom and in Chromium alike, because `focus()` is a spec no-op on `document.body` and on a detached node. Deleted; the behavioural test stayed.
- The first motion design was **wrong and its own tests were green**. Frames from a real browser at 130ms showed the row already fully muted with its verb already flipped while a 12px dot animated in the corner. The subject moved from the marker to the row.
- The glide assertion was flaky twice for two different reasons (a `getAnimations()` check has to land inside a 220ms window; `getBoundingClientRect()` includes the running transform). It now reads both claims off the `Element.prototype.animate` call that creates the glide.

## Visual evidence

Two before/after figures were produced through the product's own pipeline and are **not inline below**: this session's environment has no `gh` CLI, so `gh pr create --attach` is unavailable and there is no API path to upload a binary into a PR body. The PNGs were handed back in the session and can be dropped into this description directly. Panel digests, so it is checkable that each pair is two different renders rather than one file twice:

| figure | before panel | after panel |
|---|---|---|
| `comments-rail.png` | `8035814124f2361dd8c3a850b6a21e1c` | `b802ed2a0e720a97f9107c20f541051a` |
| `canvas-pin-count.png` | `0322dd29dd676bd078036acab07b2077` | `abddce5b6142d2b8e3215c6d93152481` |

**`comments-rail.png`** — the same three conversations with one expanded. Before: `Edit` and `Resolve` as labelled buttons under the row, a `Reply` button under the composer, no state on a collapsed row, and no last-activity stamp. After: the ring dot on the left of *every* row is the Resolve toggle, the verbs are icon-only at 44px, and each row's meta line ends `N messages · <when>`.

**`canvas-pin-count.png`** — two comment pins on a canvas, one with a four-message conversation and one with a single remark. Before they are identical dots; after, the busy one carries `4`.

What the figures cannot show, and what stands in for each:

- **The motion.** Resolve/Reopen is behaviour over time; a still frame of it is the misleading kind of evidence. `comments-panel-motion.browser.test.tsx` asserts the crossing, the hold, the leave and the glide, and the design was iterated against real captured frames rather than against the code.
- **The focus handoff.** Where the caret is does not photograph. `use-comments-rail.focus-return.browser.test.ts` and `comments-panel-focus.browser.test.tsx` hold it.
- **The gutter and preview markers.** Both carry the same count as the rail; the placement fix is asserted against the block's own rendered `<text>` in `preview-marker-placement.browser.test.tsx`, which is a stronger claim than a picture — it fails on a drift a reader's eye would forgive.

Reproduce either figure: render the component in the `web-browser` project, `page.screenshot` the host, then `git checkout 579fbd8 -- <the touched source>` for the before panel and `git checkout HEAD -- <same>` to restore.

## Checklist

## Deferred, each its own increment

- The canvas pin's own resolve/reopen **motion** — blocked on `keyed-svg-patcher` replacing a group whose bytes change, so cross-fading one is shared machinery rather than an annotation change.
- **Read status.** There is no read/unread field in `commentMessageSchema`, and adding one is an ADR-0026 amendment plus an identity decision (per-device, or a shared `readBy`). Deliberately not smuggled in here.

https://claude.ai/code/session_01BPGrBEknX78whd38tHXnLH

---
_Generated by [Claude Code](https://claude.ai/code/session_01BPGrBEknX78whd38tHXnLH)_

## Summary by CodeRabbit

* **New Features**
  * Comment counts now appear on conversation rows, gutter markers, preview markers, and canvas pins.
  * Added clearer last-activity timestamps for multi-message conversations.
  * Added icon-based comment actions with accessible labels and larger touch targets.
  * Added focus management when opening conversations, composing, editing, and returning with Escape.
  * Resolving a conversation now provides an immediate visual transition before the row leaves the active list.

* **Bug Fixes**
  * Improved comment marker placement in document previews.
  * Clarified accessibility labels for comments and multi-thread markers.


kamiazya:
### Standing down on `test-unit (2)` — this failure is not this PR's

`test-unit (2)` failed with a per-test timeout, and `test-unit (1)` was cancelled by fail-fast behind it:

```
FAIL  mcp-node  src/server/store/db/migrations/0012-ulid-remaining-document-ids.test.ts
      > re-mints a third-site nanoid row, carrying versions/branches/docKeys/blob
Error: Test timed out in 10000ms.
```

Five measurements, rather than calling it a flake and moving on:

| observation | result |
|---|---|
| that file in isolation, this checkout | **1.27s** against a 10s ceiling |
| whole `mcp-node` project locally, 372 files / 4127 tests in parallel | **passes** — the only local failure is `local-node-version.test.ts`, this container being on Node 22 while `.node-version` pins 24 |
| this PR's diff against `packages/mcp-server/src/server/store/` | **empty** — the one mcp-server file this PR touches is `file-size-budget.test.ts`, a byte-count assertion |
| base branch `main` on this check | **not red** — main's recent CI failures are in `test-browser`, not `test-unit` |
| this PR's own previous commit (`b008508`) | `test-unit (1)` and `test-unit (2)` **both succeeded**, over identical code in that path |

So: a load-dependent per-test timeout on a runner, in code this PR does not touch and did not slow, which passed on this branch one commit ago. That is the documented shape in `.claude/rules/integrator-flow.md` — a test whose premise is a quiet machine, meeting a full parallel shard.

**No fix exists to port**, and I am deliberately not inventing one here. The right patch is a `testTimeout` on that `it`, and this repo's rule is that a timeout is *a ceiling sized on a recorded measurement, never a delay*. All this run establishes is `> 10s` under CI load; I cannot reproduce the runner's disk contention locally, so any number I picked would be the guess the rule exists to prevent. Proposed follow-up, outside this PR: record the file's cost across several loaded CI runs and set the ceiling from that — or, better, cut the cost, since the test walks two full migration chains against real SQLite in a tmpdir and only the second is what it asserts on.

Spending the single re-run this rule allows now. If it fails again on the same test, it is real and I will root-cause it rather than re-run again.

For the record, the failure this push *did* fix was mine: `thread-activity.test.ts`'s generator called `toISOString()` on the `Invalid Date` that `fc.date({ min, max })` still emits (measured at 2 in 500 runs), so the run failed from the generator rather than from the property. `noInvalidDate: true`, mutation-checked at 3-of-5 fresh runs failing without it and 5-of-5 passing with it. `test-jsdom` and `stress-changed-tests` are both green on this commit.

---

kamiazya:
Correction to the comment above: I could not spend that re-run. `rerun-failed-jobs` on run `34028339834` returns `403 Resource not accessible by integration` for this integration's token, so the single allowed re-run is still unspent and `ci-gate` stays red on a failure that is not this PR's.

Nothing further from me changes that — I will not push an empty commit to kick CI, and I am not widening this PR into an unrelated migration test to work around it. A maintainer re-running the failed jobs from the Actions UI is the one click that settles whether `0012-ulid-remaining-document-ids` is load-dependent (expected) or real (in which case it is mine to root-cause, and I will). I am keeping this PR watched until then.

---

PR #1459, opened 2026-09-06
feat(annotations): the markdown markers cross to the resolved look instead of cutting

kamiazya:
## Summary

The last two surfaces where resolving a conversation was a hard cut: the source pane's gutter dot and the preview marker. The rail crosses, the canvas ramps out — these still swapped between frames.

**This surface is the OPPOSITE of the canvas, and the measurement is what says so.** The markdown editor is handed every thread including resolved ones (`threads.annotations`, unfiltered), so nothing leaves — the gutter dot, the passage underline and the preview marker stay and change. A leave animation would be wrong here for the same reason a cross would be wrong there.

DOM identity did not agree across the three markers, measured over one resolve rather than reasoned about:

| marker | on resolve | what it needed |
|---|---|---|
| CodeMirror gutter dot | **replaced** (`same-element=false`) | the state moved to `.cm-gutterElement`, which was the same element throughout |
| preview marker | **reused** (React key), but carried no resolved state at all | `data-status` — a gap, not a motion problem |
| passage underline | replaced | left cutting, deliberately (below) |

**The gutter took a wrong turn first, and the reason is worth having on the record.** The obvious move — an `elementClass` getter on `ThreadGutterMarker` — never reached the wrapper. CodeMirror re-reads `elementClass` only when a line's marker set actually differs, and it decides that with the same `eq` that governs whether the dot's DOM is reused (`setMarkers` runs behind `!sameMarkers(...)`, and `compare` is `eq`). So **one marker cannot both keep its element and announce a new class.** `gutterLineClass` is the facet provided for exactly this, and `ResolvedLineMarker` satisfies its documented contract: an `elementClass`, no `toDOM`.

**The preview finding may matter more than the motion.** A resolved conversation's marker was drawn exactly like an open one, so a reader in Read mode could not tell them apart at all.

## Test plan

`annotation-resolve-motion.browser.test.tsx` (`web-browser`) holds three claims: the gutter dot is the SAME element after a resolve while its wrapper carries the resolved class; the dot declares a colour crossing; and the preview marker says which state it is in, is reused, and crosses.

```
web-browser        235 files, 1236 tests passed
web-jsdom + mcp-node (annotation, file-size-budget)   green
pnpm -r typecheck  0 errors
pnpm lint          clean
```

Five fresh-process runs of the new file, then the whole project. All four claims mutation-checked — restoring `status` to `eq` (so the dot is replaced again), removing the line-class facet, dropping the marker's `data-status`, and removing the transitions each turn a test red.

`ThreadGutterMarker` loses its `status` field entirely: it is the line's business now, and an unread field would have been the omission with a name on it. `file-size-budget.test.ts` raises `MarkdownEditor.tsx` 1066 → 1073, with the reason beside it.

Manual verification in a real browser, and the numbers the figure is labelled with come from `getComputedStyle` in that same run: the gutter dot goes `solid` → `dotted`, the preview marker `opacity: 1` → `0.45`.

## Visual evidence

`markdown-resolve.png` — the split view before and after one resolve. **Not inline**: this environment has no `gh` CLI, so `gh pr create --attach` is unavailable and there is no API path to upload a binary into a PR body. The PNG was handed back in the session and can be dropped into this description directly. Digest: `c8f6ad18464f06196204a90694c2c5cc`.

Left panel is the open conversation — solid gutter dot, solid passage underline, full-strength preview marker. Right is the same document resolved: dotted dot, dotted underline, faded marker, everything else untouched.

What it cannot show is the crossing itself, which is behaviour over time; the computed `transition-duration` is asserted instead, in a real browser, because a declared transition and a running one are different claims and only one is visible to jsdom.

## Checklist

`apps/web/DESIGN.md` records why this surface crosses where the canvas ramps out, and the CodeMirror mechanism above.

## Left cutting, deliberately

The passage underline. Its state is a `Decoration.mark` class, and a mark has no stable wrapper to move to — the same trick would need a second decoration layer over a span whose meaning the dot beside it already carries. Recorded in `DESIGN.md` rather than left for a reader to wonder about.

https://claude.ai/code/session_01BPGrBEknX78whd38tHXnLH

---
_Generated by [Claude Code](https://claude.ai/code/session_01BPGrBEknX78whd38tHXnLH)_

### DIFF
--- a/apps/web/DESIGN.md
+++ b/apps/web/DESIGN.md
@@ -810,6 +810,34 @@ Two things measurement corrected after the mechanism worked:
   patch — the viewBox is the union of everything in the scene, so a ghost is
   the only thing that can ever be outside it.
 
+**The markdown markers cross rather than leave, and that is the opposite of
+the canvas for a measured reason.** This editor is handed EVERY thread,
+resolved ones included, so nothing goes: the gutter dot, the passage
+underline and the preview marker all stay and change. A leave animation
+would be wrong here for the same reason a cross would be wrong there.
+
+Three markers, and DOM identity did not agree across them — measured over
+one resolve, not assumed:
+
+- **The CodeMirror gutter dot was REPLACED** (`same-element=false`), so
+  nothing could transition on it. Its `.cm-gutterElement` wrapper was the
+  same element throughout, and that is where the state moved. It cannot ride
+  the dot's own marker: CodeMirror re-reads `elementClass` only when a
+  line's marker set differs, and decides that with the same `eq` that
+  governs whether the dot's DOM is reused — so one marker cannot both keep
+  its element and announce a new class. An `elementClass` getter on it never
+  reached the wrapper at all. `gutterLineClass` is the facet provided for
+  exactly this, and `ResolvedLineMarker` satisfies its contract: a class, no
+  `toDOM`.
+- **The passage highlight is replaced too**, and is left cutting. Its state
+  is carried by a `Decoration.mark` class, which has no wrapper to move to;
+  the same trick would need a second decoration layer for a span that is
+  already legible from the dot beside it.
+- **The preview marker was reused** by React key and said nothing about
+  resolved at all — a closed conversation's marker was drawn exactly like an
+  open one, which is a gap rather than a motion problem. With `data-status`
+  present the crossing needs nothing but a class.
+
 **The rail's verbs are icon-first, and the status dot IS the Resolve
 toggle.** "Object-action surfaces are icon-first" below was written, the
 canvas card followed it (`CardAction`), and the rail never got held to it:
--- a/apps/web/src/components/markdown-editor/MarkdownEditor.tsx
+++ b/apps/web/src/components/markdown-editor/MarkdownEditor.tsx
@@ -7,6 +7,7 @@ import type { MeasureText, ReferenceSeams } from '@kamiazya/whiteboard-canvas-re
 import { createBrowserMeasureText } from '@kamiazya/whiteboard-canvas-viewer'
 import {
   type CommentThread,
+  type CommentThreadStatus,
   documentIdSchema,
   type StoredCoreFacets,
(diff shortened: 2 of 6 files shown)


kamiazya:
### `test-browser (2)` died before it ran a test

Not a failure of this change, and not a flake judgement — the job's own step list says it:

| step | outcome |
|---|---|
| Set up job | success |
| checkout | success |
| `./.github/actions/setup-pnpm` | **failure** |
| Install dependencies | skipped |
| Test browser projects (shard 2/2) | **skipped** |

`corepack prepare --activate` crashed while fetching pnpm, inside Node's own bundled HTTP client:

```
Preparing pnpm@11.12.0 for immediate activation...
AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:
  assert(!this.paused)
    at Parser.finish (node:internal/deps/undici/undici:7388:9)
    at TLSSocket.onHttpSocketEnd (node:internal/deps/undici/undici:7827:34)
```

No test body was reached, no dependency was installed. This is the one case `.claude/rules/integrator-flow.md` names as re-runnable on sight — a job that died in checkout/install rather than under test.

Everything else on this commit is green, including **`test-browser (1)`**, the sibling shard: 22 of 24 checks pass and only `ci-gate` follows shard 2 down. That sibling running at all is the `fail-fast: false` from #1448 doing its job — before that change a shard death cancelled its sibling, and the cancelled leg could never be revived by "Re-run failed jobs" at all. Here the leg is `failure`, so that button does work on it.

**I cannot press it.** `rerun-failed-jobs` returns `403 Resource not accessible by integration` for this integration's token, on the correct run id (`34040773256`), so the single re-run the rule allows is still unspent.

Nothing further from me changes that: I will not push an empty commit to kick CI, and there is nothing in this PR's scope to make robust — the failure is in corepack's download, not in any code this diff touches. A maintainer re-running the failed job is the one click that settles it. I am keeping this PR watched until then.

---

PR #1480, opened 2026-09-08
test(web): web-jsdom fails a swallowed failure too, and the claims assert

kamiazya:
## Summary

- Brings `web-jsdom` to the line `web-browser` has held since #1478: a test that logs a failure and carries on **fails**, because the assertion that eventually breaks will be somewhere else.
- The ledger was filed at ~63 claims. It came in at **24**, because two thirds of the noise was never a judgement call — the subtraction ran before the ledger.
- The 24 that remain **do not get an exemption**. The claim waits for the named record and fails if it never arrives.

### Sizing down came first, and that is the reusable part

| | records | tests | what it was |
|---|---|---|---|
| measured | 219 | 70 | |
| after the act fixes (#1478) | 82 | 63 | React's act-environment complaint, 138 of them |
| after mocking the fold (#1478) | 43 | 35 | jsdom has no IndexedDB, so `foldWorkspaceDocuments` throws; the guarded warn was 36 records over 29 tests |
| after fixing one fixture | 36 | 29 | **new here** — see below |
| claimed | **0 unclaimed** | 24 claiming | tests driving a failure path on purpose |

**Before writing a ledger, subtract the entries that are the environment rather than a decision.**

The fixture one was a real gap: five `DaemonDocumentPage` version tests mocked only the **POST** `/versions`, so the panel's own **GET** fell through to the catch-all `new Response('{}')` and failed `versionsResponseSchema`. Those tests are about saving a version and announcing a bookmark — they were quietly exercising a schema failure none of them is about, logging it and carrying on. They serve the listing now.

### A claim that asserts beats an allowlist that exempts

`await expectLoggedFailure('<fragment>')` claims the record **and waits for it**, so a test that stops producing its degraded report fails rather than passing quietly — the direction a bare allowlist cannot see.

Each of those 24 already asserted that a failure is **survived** — `celebrate` resolves, `reloadFresh` still reloads, the tool registry does not take the page down. The claim adds the half every one of them was missing: that the failure is also **reported**. So this is not 24 exemptions bought; it is 24 tests that got stronger.

Two mechanics it cost, both measured rather than reasoned:

- **The claim must WAIT.** These reports are fire-and-forget catches on rejected promises and land *after* the body that provoked them. Asserting synchronously saw `(none)` and let the record leak into the NEXT test — one cause, two failing tests. The fragment is claimed *before* the wait, so a record arriving during it is already spoken for.
- **State on `globalThis`**, for the reason `browser-setup.ts` already records: the setup file and a test importing the helper are two module instances. Module scope gave each its own copy and failed the test that had just claimed its record.

## Test plan

- [x] New or updated tests — 24 tests gain an assertion they were missing; one fixture fixed
- [x] `web-jsdom` passes: **3818 passed, 0 unclaimed records**
- [ ] Manual verification — not applicable; this is test infrastructure, pinned by the mutation check below
- [ ] E2E — not applicable

Mutation-checked in both directions on `celebrate`:

| mutation | result |
|---|---|
| drop the claim | test fails as an unclaimed record |
| make the failure stop happening | claim fails: `expected a logged failure containing "confetti burst failed"` |

The second is the one that matters — it is what makes this an assertion rather than an allowlist entry.

**One pre-existing failure stays red locally and is unrelated**: `VersionTimeline` / thumbnail-via-objectURL is the documented wrong-Node-major symptom (this container has Node 22, `.node-version` pins 24). Unchanged by this diff, and `test-jsdom` is green on main.

## Visual evidence

Visual evidence: none — test infrastructure only, no production code in the diff and no pixel moves.

## Checklist

- [x] Docs updated — `executable-rungs.md` carries the full sizing table and both mechanics; `test-authoring.md` gains the write-time rule
- [x] No `console.*` added to `src/server/**`
- [x] No hand-written interfaces duplicating a Zod schema

https://claude.ai/code/session_0173KdAbXzLi3Nypu4fyneDv

---
_Generated by [Claude Code](https://claude.ai/code/session_0173KdAbXzLi3Nypu4fyneDv)_

## Summary by CodeRabbit

* **Tests**
  * Improved test coverage for expected console errors and warnings across failure scenarios.
  * Tests now verify that expected failure messages are logged, rather than only checking fallback behavior.
  * Added automatic detection of unverified console failures to reduce silently swallowed test errors.
* **Documentation**
  * Updated testing guidance with instructions and examples for asserting expected logged failures.
  * Documented how failure tracking and asynchronous assertions work.

### DIFF
--- a/.claude/rules/test-authoring.md
+++ b/.claude/rules/test-authoring.md
@@ -37,8 +37,9 @@ The write-time rules, so the skill is a lookup rather than a prerequisite:
     growth loop floors that dimension — otherwise the slower machine gets the stricter test.
 14. `act` comes from `@testing-library/react`, never `react`, and never wraps a `waitFor` /
     `findBy*` (RTL turns the act environment off inside those, deliberately).
-15. A test that provokes a logged failure claims it with `await expectLoggedFailure('<fragment>')`,
-    which asserts the report arrived — never an allowlist entry.
+15. A test that provokes a logged failure claims it — `await expectLoggedFailure('<fragment>')`
+    in `web-jsdom`, which asserts the report arrived; `expectLoggedFailures()` in `web-browser`,
+    which returns the records to assert on. Never an allowlist entry.
 
 Executable rungs already hold most of these (`pnpm lint`'s GritQL plugin, `arch-lint`'s scans,
 the jsdom setup's teardown). A shape that costs a real defect twice moves up the ladder —
--- a/.claude/rules/test-authoring.md
+++ b/.claude/rules/test-authoring.md
@@ -37,6 +37,8 @@ The write-time rules, so the skill is a lookup rather than a prerequisite:
     growth loop floors that dimension — otherwise the slower machine gets the stricter test.
 14. `act` comes from `@testing-library/react`, never `react`, and never wraps a `waitFor` /
     `findBy*` (RTL turns the act environment off inside those, deliberately).
+15. A test that provokes a logged failure claims it with `await expectLoggedFailure('<fragment>')`,
+    which asserts the report arrived — never an allowlist entry.
 
 Executable rungs already hold most of these (`pnpm lint`'s GritQL plugin, `arch-lint`'s scans,
 the jsdom setup's teardown). A shape that costs a real defect twice moves up the ladder —
--- a/.claude/skills/steward/reference/failure-modes.md
+++ b/.claude/skills/steward/reference/failure-modes.md
@@ -20,6 +20,7 @@ order effects.
 | every test PASSED and the file exits 1, `NotFoundError: removeChild` | A teardown doing `document.body.innerHTML = ''` while React roots are mounted. Use `cleanup()` |
 | "the list does not contain this item" | No list was opened — the trigger was clicked while a menu was still dismissing. Wait for `[role="menu"]` to be gone |
 | an assertion reads an empty value that was definitely typed | The element was remounted and the held reference is detached. Query inside the assertion |
+| `stress-changed-tests` red while both `test-jsdom` shards are green | The two steps split the classes: five fresh processes catch order/module-state flakes, `--repeats=3` catches state carried BETWEEN repetitions of one test. A once-per-process record (a memoised startup attempt) passes the first and fails the second |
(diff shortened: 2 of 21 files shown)


kamiazya:
Triaged all five against the code. Three taken in `86d1f886`; two not, with reasons.

**Taken**

- **Versions matcher too broad.** `init?.method !== 'POST'` also answered PUT/PATCH/DELETE. Nothing in these fixtures issues one, so this is precision rather than a fix — `(init?.method ?? 'GET') === 'GET'` at all five sites.
- **`test-authoring.md` item 15 named only the jsdom helper.** Correct, and worth fixing: the two projects genuinely differ — `expectLoggedFailure(fragment)` asserts in `web-jsdom`, `expectLoggedFailures()` returns the records in `web-browser` — so a rule naming one teaches the wrong call in the other. Both named now.
- **The sizing table read as deltas.** Each row is an independent measurement of the whole project on the tree at that point, and the test counts move between rows too, so subtracting two rows does not give the size of one cause. Each row now states its own share (138 of 219, 36 of 82, 8 of 43). The two totals are reconciled as well: **3819 is what the project collects, 3818 is what passes** — the difference is one pre-existing wrong-Node-major failure, unrelated to this diff.

**Not taken**

- **`App.test.tsx`'s unmatched `fold before counting failed` claim** — already removed in `a7896fdf`, one commit after the tree this review read. Worth recording that `stress-changed-tests` found the same defect independently: the claim named an *environmental* record that `browser-document-counts` memoises, so it exists on the first execution in a process and never again. Five fresh-process runs passed and the in-process `--repeats=3` run failed, which is exactly the split those two steps exist to separate.

- **One-to-one claim-to-record matching.** The hazard is real — one claim currently covers repeated identical records — but the fix costs more than it buys here, and the data says so. On the run this ledger was built from, **five claimed tests emit their record twice**, two of them still after the fixture fix (`startBootSequence … never settles`, `a document this build cannot read … stops reporting`). One-to-one matching makes each of those assert a **count** that is incidental to what the test is about, and a count nobody chose is the kind of assertion that breaks on an unrelated retry. The claim is meant to carry the judgement "this failure class is expected here", not "it happens exactly N times".

  If a repeated record ever turns out to hide a real defect, the right answer is a count assertion in *that* test rather than one-to-one matching for all 24.

Verified after the changes: `web-jsdom` 3818 passed, and the stress job's own `--repeats=3` command over all changed files, 250 passed.

---

